PR #4, opened 2013-04-06
Expose a number of standard JVM metrics automatically

kim:
Please advise on package structure and deliberate use of labels

### DIFF
--- a/client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
+++ b/client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
@@ -209,11 +209,11 @@ public class JVMMetrics {
           (Long) mBeanServer.getAttribute(onDirect, "MemoryUsed"));
 
       bufferPoolsMappedCapacity.set(emptyLabels(),
-          (Long) mBeanServer.getAttribute(onDirect, "TotalCapacity"));
+          (Long) mBeanServer.getAttribute(onMapped, "TotalCapacity"));
       bufferPoolsMappedCount.set(emptyLabels(),
-          (Long) mBeanServer.getAttribute(onDirect, "Count"));
+          (Long) mBeanServer.getAttribute(onMapped, "Count"));
       bufferPoolsMappedMem.set(emptyLabels(),
-          (Long) mBeanServer.getAttribute(onDirect, "MemoryUsed"));
+          (Long) mBeanServer.getAttribute(onMapped, "MemoryUsed"));
     } catch (JMException ignored) {
       log.debug("Buffer Pool metrics unavailable (Java 6?)");
     }
--- a/client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
+++ b/client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
@@ -22,8 +22,8 @@ import java.util.regex.Pattern;
  *
  * @author kim.altintop@gmail.com (Kim Altintop)
  */
-public class JvmMetrics {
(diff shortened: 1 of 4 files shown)


client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
```diff
@@ -0,0 +1,233 @@
...
+ * </p>
+ *
+ * @author kim.altintop@gmail.com (Kim Altintop)
+ */
+public class JVMMetrics {
+  private static final Logger log = LoggerFactory.getLogger(JVMMetrics.class);
+  private static final Pattern WHITESPACE = Pattern.compile("[\\s]+");
+  private static final Pattern CAMEL_CASE = Pattern.compile("([a-z\\d])([A-Z])");
+
+  @Register(name = "memory_usage_heap_total",
```
mwitkow-io: I've been implementing something similar. Can we prefix all jvm-related stats with jvm_?
jvm_gc_runs
jvm_buffer_pools?

This will make it easy to discover all jvm-related metrics in dropdowns and lists of collected metrics.

client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
```diff
@@ -0,0 +1,233 @@
...
+public class JVMMetrics {
+  private static final Logger log = LoggerFactory.getLogger(JVMMetrics.class);
+  private static final Pattern WHITESPACE = Pattern.compile("[\\s]+");
+  private static final Pattern CAMEL_CASE = Pattern.compile("([a-z\\d])([A-Z])");
+
+  @Register(name = "memory_usage_heap_total", 
+      docstring = "Overall JVM heap memory usage", baseLabels = {})
+  private static final Gauge heapMemTotal = new Gauge();
+
+  @Register(name = "memory_usage_non_heap_total",
```
mwitkow-io: Actually, can we have heap and non-heap in a single variable that has a label to differentiate the two? It then makes it easy to stack.

client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
```diff
@@ -0,0 +1,233 @@
...
+          " open (constant value)",
+      baseLabels = {})
+  private final static Gauge maxFDs = new Gauge();
+
+  @Register(name = "buffer_pools_direct_count",
+      docstring = "Number of direct NIO buffer pools allocated",
+      baseLabels = {})
+  private final static Gauge bufferPoolsDirectCount = new Gauge();
+
+  @Register(name = "buffer_pools_mapped_count",
```
mwitkow-io: I believe that mapped/direct should be a label.

client/src/main/java/io/prometheus/client/metrics/builtin/JVMMetrics.java
brian-brazil: This is an old PR, I'm planning on writing a new one of these over in simpleclient_hotspot. If you want to port this over and resolve your comments, that'd be great.

mwitkow-io: I actually have an implementation of this in the works. We're using it internally, but it would need some cleanup before upstreaming.

I'm going on vacation, so won't be able to work on it. But expect a PR in your inbox in 2 weeks time. Sounds right?

brian-brazil: Sounds good, though there's a chance I'll get there before you depending on our own internal timelines.

brian-brazil:
This is covered by #58 and previous changes, closing out.

PR #54, opened 2015-01-22
License cleanup

beorn7:
@brian-brazil

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -2,4 +2,7 @@ Prometheus instrumentation library for JVM applications
 Copyright 2012-2015 The Prometheus Authors
 
 This product includes software developed at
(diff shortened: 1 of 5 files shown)


NOTICE
```diff
@@ -0,0 +1,5 @@
+Prometheus instrumentation library for JVM applications
+Copyright 2012-2015 The Prometheus Authors
+
+This product includes software developed at
+SoundCloud Ltd. (http://soundcloud.com/).
```
brian-brazil: We should probably mention Boxever here too.

brian-brazil:
:+1:

NOTICE
beorn7: Done.

PR #59, opened 2015-02-09
Add checks for invalid metric names and labels.

### LINKED ISSUE #28: Metric and label names should be checked for validity
Reporting this because users placed a dash into a metric name, which was then not queriable via Prometheus (the query language cannot tell a dash/minus apart).

The client library should check that metric and label names are valid ones that can be queried through the query language later (and we should add the same check upon ingestion in Prometheus):

Metric names have to adhere to this regex:

- `[a-zA-Z_:]([a-zA-Z0-9_:])*`

Similarly, label names need need to adhere to this one (basically, no colons allowed):

- `[a-zA-Z_]([a-zA-Z0-9_])*`

There are currently no restrictions on label values.

brian-brazil:
Document who's responsible for checking produced metrics are valid.

Closes #28 
@juliusv

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -3,12 +3,16 @@ package io.prometheus.client;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.regex.Pattern;
 
 /**
  * A collector for a set of metrics.
  * <p>
  * Normal users should use {@link Gauge}, {@link Counter} and {@link Summary}.
+ * <p>
  * Subclasssing Collector is for advanced uses, such as proxying metrics from another monitoring system.
+ * It is it the responsibility of subclasses to ensure they produce valid metrics.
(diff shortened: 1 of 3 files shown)


juliusv:
:+1: in principle, but the build is failing with OOM. @brian-brazil @beorn7 Is lack of memory for Java client library tests a known issue?

brian-brazil:
This happens a fair bit of the time on the pushgateway test. All tests are passing on my laptop.

checketts:
This fix doesn't catch hyphens in the metric name.

`Gauge.build().name("a-b").labelNames("__name__").help("h").create();`

brian-brazil:
Are you sure your're using a recent enough version of the code?

checketts:
Sorry for leaving that off. I'm using version `0.0.19`

checketts:
Oh my code snippet was misleading. My actual code (in an exporter) is doing:

`return new MetricFamilySamples(cleanName, Type.GAUGE, help, samples);`

So it is bypassing the check. Sorry for the noise.

brian-brazil:
Custom collectors don't have sanity checks, it's up to you to ensure they're sane.

PR #62, opened 2015-02-13
Add histogram to simpleclient.

brian-brazil:
See https://docs.google.com/document/d/1uSenXRDjDaJLV3qnSD09GqgPdEEDPjER0mVsnGaCYF0/edit

Render +Inf as the server expects it.

Main issue: https://github.com/prometheus/prometheus/issues/480

@beorn7

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -9,13 +9,14 @@ The main outcomes of the benchmarks:
 * The original client is much slower than the Simpleclient or Codahale, especially when used concurrently.
 * Codahale Meters are slower than Codahale/Simpleclient Counters
 * Codahale and original client Summaries are 10x slower than other metrics.
+* Simpleclient Histograms are 10-100X faster than Codahale and original client Summaries.
 * Simpleclient `Gauge.Child.set` is relatively slow, especially when done concurrently.
-* Label lookups in both Prometheus clients is relatively slow.
+* Label lookups in both Prometheus clients are relatively slow.
 
 Accordingly, in terms of client instrumentation performance I suggest the following:
 * It's cheap to extensively instrument your code with Simpleclient Counters/Gauges/Summaries without labels, or Codahale Counters.
 * Avoid Codahale Meters, in favour of Codahale/Simpleclient Counters and calculating the rate in your monitoring system (e.g. the `rate()` function in Prometheus).
-* Avoid original client Summaries and Codahale Histograms/Timers on high update rate code paths.
+* Use Simpleclient Histograms rather than original client Summaries and Codahale Histograms/Timers.
 * Avoid the original client.
 * For high update rate (&gt;1000 per second) prometheus metrics using labels, you should cache the Child. Java 8 may make this better due to an improved ConcurrentHashMap implementation.
 * If a use case appears for high update rate use of SimpleClient's `Gauge.Child.set`, we should alter `DoubleAdder` to more efficiently handle this use case.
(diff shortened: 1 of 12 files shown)


beorn7:
Will look during the weekend, hopefully.

For Java specifics, I'd appreciate an additional reviewer who actually knows Java well.

beorn7:
Weekend was too busy with "real life" (tm). I'll get to this ASAP.

brian-brazil:
Updated with benchmarks, and change from Will to let .buckets() not have to be directly after the .build()

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -0,0 +1,262 @@
...
+    private DoubleAdder sum = new DoubleAdder();
+
+    static TimeProvider timeProvider = new TimeProvider();
+    /**
+     * Observe the given amount.
+     */
+    public void observe(double amt) {
+      for (int i = 0; i < upperBounds.length; ++i) {
+        // The last bucket is +Inf, so we always increment.
+        if (amt <= upperBounds[i]) {
```
willfleury: The code looks right for the design doc but i do have a question on the design and the fact that bucket bounds must be defined up front. All 3 bucket definition options seem to required this (some both high and low and others just low)..

That's not an easy task and would require trial and error and ongoing tuning to ensure it was a decent histogram?

brian-brazil: Yeah, it requires trial and error to get this right. I don't see any other option if you want to keep things aggregateable and to a sane number of buckets.

The goal of this metric is to give a useful indication as to what the percentile is in real-time, for an exact answer you need to do logs processing.

beorn7: @willfleury We will have a extensive section in the docs about this. It boils down to aggregatability (is that a word?) vs. accuracy & ease of handling (the latter being accomplished by the pre-calculated client-side quantiles).

beorn7: @brian-brazil I'll do everything to review this code today.

benchmark/README.md
```diff
@@ -9,13 +9,14 @@ The main outcomes of the benchmarks:
 * The original client is much slower than the Simpleclient or Codahale, especially when used concurrently.
 * Codahale Meters are slower than Codahale/Simpleclient Counters
 * Codahale and original client Summaries are 10x slower than other metrics.
+* Simpleclient Histograms are 10-100X faster than Codahale and original client Summaries.
 * Simpleclient `Gauge.Child.set` is relatively slow, especially when done concurrently.
 * Label lookups in both Prometheus clients is relatively slow.
```
beorn7: Unrelated to change:

Either "s/lookups/lookup/" or "s/is/are".

brian-brazil: Done.

beorn7:
2am, and I still haven't finished the review... As my main distraction had to do with building a house, I guess you can sympathize...

willfleury:
I did quick pass review on it and all looks good. We can do the pass to make appropriate fields final in another PR so we don't confuse this one.

simpleclient/src/main/java/io/prometheus/client/Histogram.java
willfleury: @beorn7 yeah, i read the design doc on it. Agree it is a solid solution to start with. 

It would be nice if we could figure out a more dynamic way to range it in future. Or even if we had a configuration based approach to setting it built into prometheus client so it doesn't require code changes or some custom configuration solution to manage it) - again just thoughts for future work. 

Its great to get the percentiles in.

beorn7:
:+1: As far as I can see with my limited and aging Java skills, this looks good.

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -0,0 +1,262 @@
...
+  double[] buckets;
+
+  Histogram(Builder b) {
+    super(b);
+    buckets = b.buckets;
+    initializeNoLabelsChild();
+  }
+
+  public static class Builder extends SimpleCollector.Builder<Builder> {
+    double[] buckets = new double[]{.005, .01, .025, .05, .075, .1, .25, .5, .75, 1, 2.5, 5, 7.5, 10};
```
beorn7: Late question: What's the rationale of the 7.5eX bits in there?

brian-brazil: I was trying to get reasonable granularity in an understandable way that doesn't result in too many buckets.

This will get tuned after I start using it more seriously.

beorn7: OK. Just wondered if that was already based on something sophisticated. (If you just take out the 7.5ers, you get nice almost-even "currency-style" spacing of either factor 2 or 2.5.)

PR #68, opened 2015-04-14
Add apache felix bundle plugin to add OSGi metadata to manifest during build

utobi:
@brian-brazil:
We are working on using the simpleclient in OSGi (https://github.com/swookiee/). For that (and for being usable in OSGi in general), the jars need to have additional metadata in the manifest. This PR adds the Apache Felix bundle plugin to the parent pom and sets the packaging to "bundle" in all simpleclient poms. This adds the necessary metadata to the manifest during build. 

Impact assessment: There should be no impact on other things as this only adds additional metadata to the manifest.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,12 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <version>2.7</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>2.4.0</version>
+                <extensions>true</extensions>
+            </plugin>
         </plugins>
     </build>
 
--- a/simpleclient/pom.xml
+++ b/simpleclient/pom.xml
@@ -10,6 +10,7 @@
 
     <groupId>io.prometheus</groupId>
     <artifactId>simpleclient</artifactId>
+    <packaging>bundle</packaging>
 
     <name>Prometheus Java Simpleclient</name>
     <description>
--- a/simpleclient_common/pom.xml
+++ b/simpleclient_common/pom.xml
@@ -10,6 +10,7 @@
 
     <groupId>io.prometheus</groupId>
     <artifactId>simpleclient_common</artifactId>
+    <packaging>bundle</packaging>
 
     <name>Prometheus Java Simpleclient Common</name>
     <description>
--- a/simpleclient_hotspot/pom.xml
+++ b/simpleclient_hotspot/pom.xml
@@ -10,6 +10,7 @@
 
     <groupId>io.prometheus</groupId>
     <artifactId>simpleclient_hotspot</artifactId>
+    <packaging>bundle</packaging>
 
     <name>Prometheus Java Simpleclient Hotspot</name>
     <description>
--- a/simpleclient_pushgateway/pom.xml
+++ b/simpleclient_pushgateway/pom.xml
@@ -10,6 +10,7 @@
 
     <groupId>io.prometheus</groupId>
     <artifactId>simpleclient_pushgateway</artifactId>
+    <packaging>bundle</packaging>
 
     <name>Prometheus Java Simpleclient Pushgateway</name>
     <description>
--- a/simpleclient_servlet/pom.xml
+++ b/simpleclient_servlet/pom.xml
@@ -10,6 +10,7 @@
 
     <groupId>io.prometheus</groupId>
     <artifactId>simpleclient_servlet</artifactId>
+    <packaging>bundle</packaging>
 
     <name>Prometheus Java Simpleclient Servlet</name>
     <description>


brian-brazil:
I'm not too experienced with this, but it seems okay.

I'd be happy to help with figuring out how best to instrument your library, http://prometheus.io/docs/practices/instrumentation/ are the general guidelines.

I note that you currently export to graphite, if you want to support that via prometheus I'd be happy to accept a PR to add that as a module to the simpleclient (unless I happen to get around to it first, but don't hold your breath).

utobi:
Thanks for merging and the offer.
I already experimented on a PoC feature branch (https://github.com/swookiee/com.swookiee.runtime/tree/feature/prometheus-metrics/com.swookiee.runtime.prometheus) and the basic functionality seems to be working well so far. Still needs a lot of work though.

We will likely need to keep supporting both graphite and prometheus but supporting both over the same API is a nice idea. I'll discuss it with the other guys.

brian-brazil:
From a quick peek, a few things stand out:
https://github.com/swookiee/com.swookiee.runtime/blob/feature/prometheus-metrics/com.swookiee.runtime.prometheus/src/main/java/com/swookiee/runtime/prometheus/TimingResourceFilter.java#L57

5000 buckets will cause performance problems, especially when broken out by method and resource. Aim for 10-20 buckets. You can also remove the summary, as the histogram exports the _sum and _count. You may wish to prefix all the metric names with "swookie_" as there can be many types of requests in a system.

As a high level comment, if it's possible within the framework I'd suggest directly instrumenting the request dispatcher/router so that your users get instrumentation for free rather than having to manually add filters to each resource themselves.

https://github.com/swookiee/com.swookiee.runtime/blob/feature/prometheus-metrics/com.swookiee.runtime.prometheus/src/main/java/com/swookiee/runtime/prometheus/StatusCodeFilter.java
_total is the usual suffix for counters.

https://github.com/swookiee/com.swookiee.runtime/tree/feature/prometheus-metrics/com.swookiee.runtime.prometheus/src/main/java/com/swookiee/runtime/prometheus/jvm
Why are you duplicating simpleclient_hotspot, Is there something we can do to improve it?

utobi:
The filters are active for all resources so it is not necessary to add them manually. Right now we are not sure yet which route to go (histograms vs summaries, sane defaults, resource whitelist vs resource blacklist etc. etc.).

The rest is there because it is really in a very early stage to see if it is possible and to test a couple of things that would be relevant for our current production usage (e.g. the 5k buckets). Hotspot stuff requires us to do some work to expose com.sun.management before we can switch over to simpleclient_hotspot. 

The 10-20 buckets thing is interesting. Can you share a bit more about that? Especially interesting would be what the performance impact is for more buckets and if it is on the node or on the prometheus server. Use-case I am currently seeing for us is calculating latency percentiles over multiple hosts. 10-20 buckets seems to be very coarse for that. If it is only on the node. I would be willing to sacrifice a bit performance on the node for more granularity :)

Thanks for taking the time.

brian-brazil:
Each bucket is sent back as a timeseries to prometheus, so you'll have memory costs on the client side and cpu/disk costs on the prometheus side. With say 20 resources and 10 machines, that's 1M timeseries - which is going to require a very beefy machine to run prometheus on and that's just for one metric.

10-20 buckets tends to be enough in practice to get a good idea as to what's going on, and the default buckets should cover a reasonable range for web services.

I'd generally encourage summaries everywhere, and histograms very sparingly (e.g. only over the server as a whole) as they can get expensive.

utobi:
That makes sense. Sounds reasonable to have summaries as default and only use histograms on demand.

Do you know if there are some numbers on the memory impact? If not, we'll happily run some tests.

utobi:
I just had a thought. Now we are way off from the pull request but maybe it is interesting to someone else too :)

Having a high number of buckets to have a high granularity to be able to calculate very precise percentiles over multiple servers might not require a high resolution. For example for us, it might be sufficient to scrape this only every 20m, every 30m or even every 60m. Instrumenting the 5-6 important resources with ~500 buckets and exposing that on a different endpoint (e.g. /histogram_metrics) should ease the load on the prometheus server a lot. What do you think?

brian-brazil:
> Do you know if there are some numbers on the memory impact? If not, we'll happily run some tests.

It depends on how much concurrency it has, I don't have numbers off the top of my head. Wouldn't hurt to have those alongside the benchmarks of the simpleclient.

> Instrumenting the 5-6 important resources with ~500 buckets and exposing that on a different endpoint (e.g. /histogram_metrics) should ease the load on the prometheus server a lot. What do you think?

That should work in theory, but not in practice. Prometheus will consider a timeseries with no new data points in the last 5 minutes (configurable) to be stale, and aggregation across servers scraped with that much skew (scrapes are randomised across the collection interval) won't be very useful.

I'd generally advise if you want that level of accuracy that you do it based on logs rather than based on an online monitoring system such as Prometheus, as that's a simpler approach that doesn't risk your alerting falling over.

Issue #69, opened 2015-04-30
HashCode stored in ThreadLocal causes warnings during deploy


jsuchenia:
Hi

During redeploy of our application we got a lot of errors regarding memory leak. (Tomcat 8 with Leak Listener enabed).

Example log:
```
SEVERE: The web application [ROOT] created a ThreadLocal with key of type [io.prometheus.client.Striped64.ThreadHashCode] (value [io.prometheus.client.Striped64$ThreadHashCode@7b13ffe9]) and a value of type [io.prometheus.client.Striped64.HashCode] (value [io.prometheus.client.Striped64$HashCode@184dd3c2]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
```

brian-brazil:
The design of Striped64 is that once it creates additional Cells due to contention that it won't remove them afterwards, on the basis that you're likely to have that level of concurrency again in the future.

https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Striped64.java#L17 has more detail.

Would this explain what you're seeing?

jsuchenia:
I know a design a solution, question is: Do we need to store own class in a ThreadLocal storage?
I found only one place where it's used. It's in DoubleAdder:
```
int h = (hc = threadHashCode.get()).code;
```
Maybe you can store generated hashCode as an Integer, not HashCode inner class ?

brian-brazil:
This code is copied from a recent versions of Java (older versions are missing it), so I'm afraid I can't comment on the exact implementation details. If there's a newer upstream version I'll be happy to update our copy.

jsuchenia:
This is not a master repo for prometheus libraries? Should I prepare a Pull request for that?
Sorry I didn't got your answer..

brian-brazil:
This is the master repo for the Prometheus java library. The code in question is from JCP JSR-166 and is included in Java 8 for example. As not all versions of Java include this library, we've copied it into this repository to avoid the need for users to pull in dependencies.

If there's an issue with this code, it should be fixed in the core Java libraries so that everyone can benefit.

jsuchenia:
In a latest version of Stped64 in Java code:
http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/ce1c874903cb/src/share/classes/java/util/concurrent/atomic/Striped64.java
they are storing a Double (via ThreadLocalRandom class)

In a JSR proposal:
[...]

PR #70, opened 2015-05-04
Update Striped64 & DoubleAdded to avoid ClassLoader leak

jsuchenia:
Here is a change for issue #69. Both fies related to JSR166 were updated to latest versions


brian-brazil:
Thanks!

PR #74, opened 2015-05-26
add support for log4j and logback appenders

willfleury:
Also cleaned up the poms a bit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@
 target/
 simpleclient_pushgateway/mockserver.log
 simpleclient_pushgateway/mockserver_request.log
+nb-configuration.xml
+
--- a/simpleclient_log4j/pom.xml
+++ b/simpleclient_log4j/pom.xml
@@ -9,12 +9,12 @@
     </parent>
 
     <groupId>io.prometheus</groupId>
-    <artifactId>simpleclient_common</artifactId>
+    <artifactId>simpleclient_log4j</artifactId>
     <packaging>bundle</packaging>
 
-    <name>Prometheus Java Simpleclient Common</name>
+    <name>Prometheus Java Simpleclient log4j</name>
     <description>
-        Common code used by various modules of the Simpleclient. 
+        Metrics collector for log4j appender logging
     </description>
 
     <licenses>
@@ -27,11 +27,15 @@
 
     <developers>
         <developer>
-            <id>brian-brazil</id>
-            <name>Brian Brazil</name>
-            <email>brian.brazil@boxever.com</email>
+            <id>willfleury</id>
+            <name>Will Fleury</name>
+            <email>will.fleury@boxever.com</email>
         </developer>
     </developers>
+
+    <properties>
+        <log4j.version>1.2.17</log4j.version>
+    </properties>
 
     <dependencies>
         <dependency>
@@ -39,11 +43,24 @@
             <artifactId>simpleclient</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <!-- Test Dependencies Follow -->
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+            <version>${log4j.version}</version>
+        </dependency>
+        
+
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+        
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
+++ b/simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
@@ -1,0 +1,80 @@
+package io.prometheus.client.log4j;
+
+import io.prometheus.client.Counter;
+import org.apache.log4j.AppenderSkeleton;
+import org.apache.log4j.Level;
+import org.apache.log4j.spi.LoggingEvent;
+
+public class InstrumentedAppender extends AppenderSkeleton {
+
+  public static final String COUNTER_NAME = "log4j_appender_total";
+
+  private static final Counter COUNTER;
+  private static final Counter.Child TRACE_LABEL;
+  private static final Counter.Child DEBUG_LABEL;
+  private static final Counter.Child INFO_LABEL;
+  private static final Counter.Child WARN_LABEL;
+  private static final Counter.Child ERROR_LABEL;
+  private static final Counter.Child FATAL_LABEL;
+  
+  static {
+    COUNTER = Counter.build().name(COUNTER_NAME)
+            .help("Log4j log statements at various log levels")
+            .labelNames("level")
+            .register();
+
+    TRACE_LABEL = COUNTER.labels("trace");
+    DEBUG_LABEL = COUNTER.labels("debug");
+    INFO_LABEL = COUNTER.labels("info");
+    WARN_LABEL = COUNTER.labels("warn");
+    ERROR_LABEL = COUNTER.labels("error");
+    FATAL_LABEL = COUNTER.labels("fatal");
+  }
+
+  /**
+   * Create a new instrumented appender using the default registry.
+   */
+  public InstrumentedAppender() {
+  }
+
+
+  @Override
+  public void activateOptions() {
+  }
+
+  @Override
+  protected void append(LoggingEvent event) {
+    switch (event.getLevel().toInt()) {
+      case Level.TRACE_INT:
+        TRACE_LABEL.inc();
+        break;
+      case Level.DEBUG_INT:
+        DEBUG_LABEL.inc();
+        break;
+      case Level.INFO_INT:
+        INFO_LABEL.inc();
+        break;
+      case Level.WARN_INT:
+        WARN_LABEL.inc();
+        break;
+      case Level.ERROR_INT:
+        ERROR_LABEL.inc();
+        break;
+      case Level.FATAL_INT:
+        FATAL_LABEL.inc();
+        break;
+      default:
+        break;
+    }
+  }
+
+  @Override
+  public void close() {
+    // nothing doing
+  }
+
+  @Override
+  public boolean requiresLayout() {
+    return false;
+  }
+}
--- a/simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
+++ b/simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
@@ -1,0 +1,66 @@
+package io.prometheus.client.logback;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.UnsynchronizedAppenderBase;
+import io.prometheus.client.Counter;
+
+public class InstrumentedAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
+
+  public static final String COUNTER_NAME = "logback_appender_total";
+  
+  private static final Counter COUNTER;
+  private static final Counter.Child TRACE_LABEL;
+  private static final Counter.Child DEBUG_LABEL;
+  private static final Counter.Child INFO_LABEL;
+  private static final Counter.Child WARN_LABEL;
+  private static final Counter.Child ERROR_LABEL;
+  
+  static {
+    COUNTER = Counter.build().name(COUNTER_NAME)
+            .help("Log4j log statements at various log levels")
+            .labelNames("level")
+            .register();
+
+    TRACE_LABEL = COUNTER.labels("trace");
+    DEBUG_LABEL = COUNTER.labels("debug");
+    INFO_LABEL = COUNTER.labels("info");
+    WARN_LABEL = COUNTER.labels("warn");
+    ERROR_LABEL = COUNTER.labels("error");
+  }
+
+  /**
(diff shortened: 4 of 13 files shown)


simpleclient_log4j/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+        <version>0.0.9-SNAPSHOT</version>
+    </parent>
+
+    <groupId>io.prometheus</groupId>
+    <artifactId>simpleclient_log4j</artifactId>
+    <packaging>bundle</packaging>
+
+    <name>Prometheus Java Simpleclient log4j</name>
+    <description>
+        Log4j appender for Prometheus
```
brian-brazil: Can you clarify that this is a collector for metrics, not something that logs things about prometheus?

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
```diff
@@ -0,0 +1,93 @@
...
+        this.trace = createCounterForLevel("trace");
+        this.debug = createCounterForLevel("debug");
+        this.info = createCounterForLevel("info");
+        this.warn = createCounterForLevel("warn");
+        this.error = createCounterForLevel("error");
+        this.fatal = createCounterForLevel("fatal");
+    }
+	
+	private Counter createCounterForLevel(String level) {
+		return Counter.build().name("log4j_appender_" + level)
```
brian-brazil: Any time you find yourself dynamically generating metric names, you should probably be using a label. Counters should also end with `_total` by convention.

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
```diff
@@ -0,0 +1,93 @@
...
+import io.prometheus.client.Counter;
+import org.apache.log4j.Appender;
+import org.apache.log4j.AppenderSkeleton;
+import org.apache.log4j.Level;
+import org.apache.log4j.spi.LoggingEvent;
+
+
+public class InstrumentedAppender extends AppenderSkeleton {
+    
+	private final CollectorRegistry registry;
```
brian-brazil: Indentation is off here

simpleclient_log4j/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_common</artifactId>
```
brian-brazil: I don't think you're using this

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
```diff
@@ -0,0 +1,93 @@
...
+    private Counter all;
+    private Counter trace;
+    private Counter debug;
+    private Counter info;
+    private Counter warn;
+    private Counter error;
+    private Counter fatal;
+
+    /**
+     * Create a new instrumented appender using the default registry.
```
brian-brazil: This won't work if someone instantiates two of this class. I'd say for the initial version that you instrument in the usual way with static variables and without the registry being configurable. If other use cases come up, we can figure out how to do this properly then.

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.log4j.AppenderSkeleton;
+import org.apache.log4j.Level;
+import org.apache.log4j.spi.LoggingEvent;
+
+
+public class InstrumentedAppender extends AppenderSkeleton {
+    
+	private final CollectorRegistry registry;
+
+    private Counter all;
```
brian-brazil: The only time to do aggregation in the client is when you're running into performance issues due to high cardinality. That's not going to be a problem here, so there's no need to have an `all`.

.gitignore
```diff
@@ -11,3 +11,4 @@
 target/
 simpleclient_pushgateway/mockserver.log
 simpleclient_pushgateway/mockserver_request.log
+/nb-configuration.xml
```
brian-brazil: Missing newline at end of file

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
willfleury: Yeah, you're right. I toyed with both but went with less expensive option. However this is on an IO path so I suppose its fine!

name=log4j_appender_total label={level} make sense so?

simpleclient_log4j/pom.xml
willfleury: Nope. Copy paste error

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
willfleury: Yeah, you're right. I'm pretty sure its a singleton so only single instance but best to avoid headaches I suppose. It makes testing easier when its configurable

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
willfleury: Yeah, when using labels it makes sense

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
brian-brazil: yeah, I'd pre-cache the labels() return value at construction time to be safe. It might be an I/O path, but many people overuse logging enough that it may be relevant.

simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
```diff
@@ -0,0 +1,84 @@
...
+    trace = counter.labels("trace");
+    debug = counter.labels("debug");
+    info = counter.labels("info");
+    warn = counter.labels("warn");
+    error = counter.labels("error");
+
+    super.start();
+  }
+
+  private Counter createCounterForLevel(String level) {
```
brian-brazil: I think you meant to remove this

simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
```diff
@@ -0,0 +1,84 @@
...
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.Appender;
+import ch.qos.logback.core.UnsynchronizedAppenderBase;
+import io.prometheus.client.CollectorRegistry;
+import io.prometheus.client.Counter;
+
+public class InstrumentedAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
+
+  public static final String COUNTER_NAME = "logback_appender_total";
```
brian-brazil: This is only used in one place, put in the value rather than requiring someone reading this to jump around the code.

simpleclient_log4j/src/main/java/io/prometheus/client/log4j/InstrumentedAppender.java
willfleury: When I went back to change the counters to static I remembered that appenders must be singleton for file logging etc to work correctly so I left them as non static which made testing easier. 

Hope thats ok..

simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
willfleury: yep, its gone in subsequent commit

simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
willfleury: its used by the test class also to get the counter value from the registry

simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * @param registry the collector registry
+   */
+  public InstrumentedAppender(CollectorRegistry registry) {
+    this.registry = registry;
+    setName(Appender.class.toString());
+  }
+
+  @Override
+  public void start() {
```
brian-brazil: Can start be called multiple times?

willfleury: Nope
http://logback.qos.ch/manual/appenders.html

brian-brazil: Even across unittests that reinitilize the entire logging setup?

If there's any cases where this can be called twice or this class instantiated twice then it'll break scraping the entire target, so I think we should play this one safe.

willfleury: If you think about file Appenders etc, they have to be provided as singletons or they would break. Its the framework (log4j and logbakc) to ensure that regardless of usage. They use statics so we don't have to :) 

You'd wouldn't create a static FileOutputStream.. Counter is same thing really. Trust me, it handles it.

brian-brazil: We also have multiple of a given type of Appender configured in some of our services with different parameters, which I can only see working with multiple objects. I also don't see any restriction otherwise mentioned as part of the interface.
I see no code in the AppenderBase to prevent multiple calls to start(), and there's no statics to protect against multiple instances existing. The protections are on doAppend(), rather than on the object lifecycle.

As we need to be absolutely certain that this code is run only once, I think it's safest to do the statics ourself.

willfleury: Yeah, multiple instances of the same appender can be declared in the logging config but in this case that should thrown an exception as we would double count otherwise seeing as we only ever have a single registry. We can have a static boolean to check and throw this exception explicitly if we want. A static counter would have hidden that issue and led to so!e fun debugging! Typically each appender would have to have different key config such as filename. This doesn't apply here - different registries is only use case I can think of and statics again couldn't work for that.

AppenderBase/Skeleton just do the work of ensuring its started / initialised before doAppend is called and in the logback impl we made it so its not sync doAppend as we don't need it to be. 

I don't mind if you want to update the pr to use statics. I consider it bad coding style for this usage though so I'd rather not add them myself.

brian-brazil: > in this case that should thrown an exception as we would double count otherwise seeing as we only ever have a single registry.

No exception would be thrown currently, prometheus scrapes of the target will fail though due to duplicate metrics.
I think double counting is preferable to failing scrapes for the entire target.

willfleury: Neither is acceptable as both are debugging nightmares.
 I forgot how Prometheus currently handles that and I think its not good. It should thrown an exception saying already registered. There's no way with Prometheus to check if metric already registered is there? If not it'll require a sync flag check and will throw exception indicating metric appender cannot be registered twice.

brian-brazil: Not at the moment, catching this automatically is a todo item (there's subtleties with it and it's not possible in the general case).

willfleury: I added an effective singleton check to both of the appender implementations.. However, I'm not a big fan of how its done or the need for it. It doesn't feel right. I read the existing log4j and logback appender implementations for things like files and sockets etc and they don't try to be clever and ensure anything is configured incorrectly (as two file or socket appenders pointing to the same file/socket).  Its by convention.

It would be good in the long run to have prometheus throw an exception or at least provide the ability to check if a metric is already registered. Accidentally being able to break and entire endpoint from scraping by adding a metric with the same name isn't feasible in a large system. Is there a ticket where I can vote for this to be fixed? A solution to that would also remove the need for the separate static boolean singletonFlag in this code.

willfleury: I updated to use static counter now so you'll be happy :+1:  

The alternative didn't feel right at all. I changed my mind after seeing that with other similar implementations, if someone configures it wrong then tough, you will get double counting, same as if you configure another appender to the same target by mistake. So its sticking to what people should be used to with both implementations.

simpleclient_log4j/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+        <log4j.version>1.2.17</log4j.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
```
brian-brazil: Should this be provided to avoid replacing the user's own version?

brian-brazil:
Can you squash your commits please with `git rebase -i`?

simpleclient_log4j/pom.xml
willfleury: not if the person was including this dependency only (i.e. the only wanted to use the log metrics and nothing else).. 

The user shouldn't be including different versions of this module and the core anyway.

PR #76, opened 2015-07-06
Undo properties for versions.

### LINKED ISSUE #75: Distribution error
Hello! I have this error while bootstrapping of dependencies:

```
Retrieving io/prometheus/simpleclient/0.0.9/simpleclient-0.0.9.pom from central
Could not transfer artifact io.prometheus:simpleclient:pom:0.0.9 from/to central (https://repo1.maven.org/maven2/): Checksum validation failed, expected 6b5cd46506ce0d2f5d90b659d1134c4f8cca5bbd but is 6e5340c298b8dfc29b1c52c1c54deae77a46d69d
```

brian-brazil:
This breaks several build tools, including
sbt and leiningen.

@prepor @xperimental This should fix #75. I've pushed this as a snapshot release if you'd like to try it out.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -39,12 +39,12 @@
         <dependency>
             <groupId>io.prometheus</groupId>
             <artifactId>client</artifactId>
-            <version>${project.version}</version>
+            <version>0.0.10-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>io.prometheus</groupId>
             <artifactId>simpleclient</artifactId>
-            <version>${project.version}</version>
+            <version>0.0.10-SNAPSHOT</version>
         </dependency>
         <dependency>
           <groupId>com.codahale.metrics</groupId>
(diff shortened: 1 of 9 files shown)


xperimental:
The checksum warnings are gone when I use the snapshot dependency in a maven project. Unfortunately I can not use the snapshot version in my sbt project, because ivy (the dependency resolver that's used by sbt) is not compatible with the timestamped snapshot releases maven uses (which is unfortunate).

Looking at the patch, I would expect it to run ok (I had done a similar change locally already to have something to develop with). From my understanding the variable in the junit dependency should be okay because it can be resolved once the parent pom has been resolved, but I am not 100% sure.

brian-brazil:
Great, I'll merge this then. I'll do a new release once #77 and one small followup is in.

xperimental:
Uh, that was fast. :+1:

brian-brazil:
For tooling I don't use myself, I'm dependant on those who report the issue to verify that the fix works as expected. At least until I can find time to write and maintain automated regression tests with all the standard toolchains :)

PR #77, opened 2015-07-06
Switch to new pushgateway URL.

brian-brazil:
Note: Anyone relying on setting the instance to empty string
will now get an empty instance label, rather than the previous
behaviour of the pushgateway using the IP address of the client.


beorn7:
Weren't you planning to add a convenience function that adds the hostname and port as instance label?

brian-brazil:
@beorn7 Yes, I'll do it in a future PR though as this one is beefy enough as-is.

beorn7:
:+1: (But as usual, Java export review needed.)

brian-brazil:
This has been reviewed within Boxever by Pete.

PR #78, opened 2015-07-09
Add convenience function for setting instance to the IP.

brian-brazil:
The new pushgateway url doesn't support setting the instance to the client IP,
so for those who were doing:
pg.push(registry, "job", "");
they can now do:
pg.push(registry, "job", PushGateway.instanceIPGroupingKey());
instead in many cases.


markreddy:
LGTM.

PR #80, opened 2015-07-13
Adds return values to Timer observations.

frankwis:
While using the client library we ran into the situation where we'd like to expose the measured ```Timer``` observations (e.g. in HTTP response headers). Since the ```Timer::observeDuration()``` methods didn't not return the elapsed time we had to measure internally and call ```observe(double amt)``` on their respective ```Collector```. This PR enables the ```Timer```s to serve such use cases.

@beorn7 @brian-brazil please have look whether you'd like this feature to be merged. Thanks!

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Gauge.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Gauge.java
@@ -97,9 +97,12 @@ public class Gauge extends SimpleCollector<Gauge.Child, Gauge> {
      }
      /**
       * Set the amount of time in seconds since {@link Child#startTimer} was called.
+      * @return Measured duration in seconds since {@link Child#startTimer} was called.
       */
-     public void setDuration() {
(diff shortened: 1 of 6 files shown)


brian-brazil:
This seems fine in principle, can you update the docstrings?

frankwis:
Thanks, @brian-brazil. I updated the docstrings. Please have look whether they suit your standards!

simpleclient/src/main/java/io/prometheus/client/Gauge.java
```diff
@@ -97,9 +97,12 @@ private Timer(Child child) {
      }
      /**
       * Set the amount of time in seconds since {@link Child#startTimer} was called.
+      * @return Measured duration since {@link Child#startTimer} was called.
```
brian-brazil: In seconds

brian-brazil:
Can you squash the commits please?

frankwis:
@brian-brazil Done :-)

PR #83, opened 2015-09-03
Put response body in pushgateway exception too.

### LINKED ISSUE #82: Error messages from pushgateway are swallowed.
https://github.com/prometheus/client_java/blob/master/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java#L272-L274 only reports the URL and the status code in case of an error pushing to pushgateway. However, the body does contain valuable information about the reason for the error. The error message should be reported, too.

brian-brazil:
Fixes #82 

@beorn7

### DIFF
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -268,9 +268,10 @@ public class PushGateway {
         writer.close();
       }
 
-      int response = connection.getResponseCode();
-      if (response != HttpURLConnection.HTTP_ACCEPTED) {
-        throw new IOException("Response code from " + url + " was " + response);
+      int responseCode = connection.getResponseCode();
+      if (responseCode != HttpURLConnection.HTTP_ACCEPTED) {
+        throw new IOException("Response code from " + url + " was " + responseCode
+            + ", message: " + connection.getResponseMessage());
       }
     } finally {
       connection.disconnect();


simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -268,9 +268,10 @@ void doRequest(CollectorRegistry registry, String job, Map<String, String> group
         writer.close();
       }
 
-      int response = connection.getResponseCode();
-      if (response != HttpURLConnection.HTTP_ACCEPTED) {
-        throw new IOException("Response code from " + url + " was " + response);
+      int responseCode = connection.getResponseCode();
+      if (responseCode != HttpURLConnection.HTTP_ACCEPTED) {
+        throw new IOException("Response code from " + url + " was " + responseCode
+            + ", message: " + connection.getResponseMessage());
```
beorn7: This will just print the standardized response message (like "Not Found" for 404 or "Internal Server Error" for 500). The important information, however, is in the body. (getContent() or something?)

brian-brazil: Even better, the `getResponseCode` call throws an exception when there's a 400 so I'm not sure it's even possible to get to the content without pulling in dependencies.

brian-brazil: Yeah, found the relevant code: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/sun/net/www/protocol/http/HttpURLConnection.java#1622

I'm afraid it doesn't look like this is going to be possible.

beorn7: In my example setup, line 273 is reached:

`java.io.IOException: Response code from http://localhost:9091/metrics/job/logmatic_sift was 500, message: Internal Server Error`

However, the body of that response contains, which is never exposed on the client side:

`text format parsing error in line 28: second HELP line for metric name "total_read"`

This is the recommended way of sending error messages in HTTP.  Surely there is a simple way of accessing that message in Java?

brian-brazil: Stackoverflow indicates that it may be JVM dependant. I'm on 1.8.0_51.

> Surely there is a simple way of accessing that message in Java?

You'd think so, wouldn't you?

beorn7: Ugh... in any case, I have filed https://github.com/prometheus/pushgateway/issues/55 to remedy things from the other side.

PR #86, opened 2015-10-10
Improve type inference in SimpleCollector.Builder

### LINKED ISSUE #65: SimpleCollector.Builder register's type inference goes wrong way
This is a problem you can encounter mostly Scala. Because Scala does type inference for parameters, if you do:

```
val myCounter = Counter.build().name("foo").register()
```
it will have type `Nothing`

That's because the type inference introduced in
https://github.com/prometheus/client_java/commit/ed184d8e50c82e98bb2706723fff764424840c3a
cause it to infer the type from the assignment parameter, not from the type of Builder.

I'll send in a patch once I find some time to fix it.

andersschuller:
I have tweaked some of the generics on `SimpleCollector` and `SimpleCollector.Builder` in order to improve the type inference and compile-time guarantees when using builders. This should fix #65, but it actually also fixes an issue for Java users. Currently, the following code compiles, but fails at runtime with a ClassCastException:

    Gauge gauge = Counter.build().name("name").help("help").register();

This is because the return types of `register` and `create` are currently inferred at the call site, rather than being determined by the builder itself. I have therefore added a second type parameter to the builder, which is the type of collector it builds.

I have also removed the second type parameter from `SimpleCollector`, since it was unused. I think the intention of this parameter was to use it as I am using the new builder type parameter, but since the builder classes are static this was not possible.

This change breaks the API for anyone extending `SimpleCollector`, but since the documentation says "you should never need to subclass this class", I'm assuming this is not really an issue.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Counter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Counter.java
@@ -64,13 +64,13 @@ import java.util.Map;
  * These can be aggregated and processed together much more easily in the Promtheus 
  * server than individual metrics for each labelset.
  */
-public class Counter extends SimpleCollector<Counter.Child, Counter> {
+public class Counter extends SimpleCollector<Counter.Child> {
 
   Counter(Builder b) {
     super(b);
   }
 
-  public static class Builder extends SimpleCollector.Builder<Builder> {
+  public static class Builder extends SimpleCollector.Builder<Builder, Counter> {
     @Override
     public Counter create() {
       return new Counter(this);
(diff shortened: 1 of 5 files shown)


brian-brazil:
Thanks for this it looks good to me, @willfleury would you mind taking a look at this as you're more an expert here than I?

Are there any tests we could add to prevent an accidental regression here in the future?

> This change breaks the API for anyone extending SimpleCollector, but since the documentation says "you should never need to subclass this class", I'm assuming this is not really an issue.

Indeed.

willfleury:
Looks good. Much better than we had it originally

brian-brazil:
Great, thanks!

PR #89, opened 2015-10-13
Remove original client.

brian-brazil:


benchmark/pom.xml
```diff
@@ -39,7 +39,7 @@
         <dependency>
             <groupId>io.prometheus</groupId>
             <artifactId>client</artifactId>
-            <version>0.0.11-SNAPSHOT</version>
+            <version>0.0.10</version>
```
juliusv: Could you explain this version downgrade for the uninitiated?

brian-brazil: 0.0.11 isn't released yet, so 0.0.10 is the last (ever) release of the original client.

juliusv: Ok, another question because I have no clue what's going on here. Why is there still some dependency section here at all for the "client", which is being deleted in this PR?

brian-brazil: I still want it in the benchmark. The binaries are still around.

PR #91, opened 2015-10-20
Add Histogram to the list of standard metrics.

brian-brazil:
Fix typo, make it completely clear that SimpleCollector is not
to be subclassed.


beorn7:
:+1:

PR #93, opened 2015-11-13
Add set() to Counters.

iksaif:
While it should generally not be used, having a set() method
for counters is this only way to relay counters from other
monitoring systems. It was implemented in the golang client
for the same reasons.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Counter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Counter.java
@@ -114,6 +114,23 @@ public class Counter extends SimpleCollector<Counter.Child> {
       value.add(amt);
     }
     /**
+     * Set is used to set the Counter to an arbitrary value. It is only used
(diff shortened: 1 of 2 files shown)


brian-brazil:
Counters should not have a set method, and the use of it to proxy from other monitoring systems is an anti-pattern.

If you want to proxy from other monitoring systems you should write a custom collector (or in Go use a ConstMetric).

iksaif:
Should the Go version of this be removed/deprecated then ? See https://github.com/prometheus/client_golang/blob/master/prometheus/counter.go#L33

brian-brazil:
Yes it should be removed, however there's many collectors and exporters out there that need fixing first.

iksaif:
In this cases it would be nice to add some comments there stating that this interface is deprecated. The current comments make it sound like it's still the proper way to do that :) (which got me confused in the first place). Anyway, collectors work for my usecase too.

PR #94, opened 2015-12-11
initialize is a static method

ohookins:
Self-explanatory (hopefully) simple in-line docs fix.

@brian-brazil


brian-brazil:
Thanks!

PR #102, opened 2016-02-01
Added support for basic Thread Counts

robdrysdale:
Added support for basic Thread Counts.  This is similar to what is available from codahale.  Missing from current lib for some reason and much needed for tracking thread leaks.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
@@ -1,0 +1,89 @@
+package io.prometheus.client.hotspot;
+
+import io.prometheus.client.Collector;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadMXBean;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Exports metrics about JVM thread areas.
+ * <p>
+ * Example usage:
+ * <pre>
+ * {@code
+ *   new ThreadExports().register();
+ * }
+ * </pre>
+ * Example metrics being exported:
+ * <pre>
+ *   jvm_threads_current{} 300
+ *   jvm_threads_daemon{} 200
+ *   jvm_threads_peak{} 410
+ *   jvm_threads_started_total{} 1200
+ * </pre>
+ */
+public class ThreadExports extends Collector {
+  private final ThreadMXBean threadBean;
+
+  public static final List<String> EMPTY_LABEL = Collections.emptyList();
+
+  public ThreadExports() {
+    this(ManagementFactory.getThreadMXBean());
+  }
+
+  public ThreadExports(ThreadMXBean threadBean) {
+    this.threadBean = threadBean;
+  }
+
+  void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
+    sampleFamilies.add(
+            new MetricFamilySamples(
+                    "jvm_threads_current",
+                    Type.GAUGE,
+                    "Current thread count of a JVM",
+                    Collections.singletonList(
+                            new MetricFamilySamples.Sample(
+                                    "jvm_threads_current", EMPTY_LABEL, EMPTY_LABEL,
+                                    threadBean.getThreadCount()))));
+
+    sampleFamilies.add(
+            new MetricFamilySamples(
+                    "jvm_threads_daemon",
+                    Type.GAUGE,
+                    "Daemon thread count of a JVM",
+                    Collections.singletonList(
+                            new MetricFamilySamples.Sample(
+                                    "jvm_threads_daemon", EMPTY_LABEL, EMPTY_LABEL,
+                                    threadBean.getDaemonThreadCount()))));
+
+    sampleFamilies.add(
+            new MetricFamilySamples(
+                    "jvm_peak_threads",
+                    Type.GAUGE,
+                    "Peak thread count of a JVM",
+                    Collections.singletonList(
+                            new MetricFamilySamples.Sample(
+                                    "jvm_threads_peak", EMPTY_LABEL, EMPTY_LABEL,
+                                    threadBean.getPeakThreadCount()))));
+
+    sampleFamilies.add(
+            new MetricFamilySamples(
+                    "jvm_threads_started_total",
+                    Type.COUNTER,
+                    "Started thread count of a JVM",
+                    Collections.singletonList(
+                            new MetricFamilySamples.Sample(
+                                    "jvm_threads_started_total", EMPTY_LABEL, EMPTY_LABEL,
+                                    threadBean.getTotalStartedThreadCount()))));
+  }
+
+
+  public List<MetricFamilySamples> collect() {
+    List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
+    addThreadMetrics(mfs);
+    return mfs;
+  }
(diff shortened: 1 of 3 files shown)


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+ * <p>
+ * Example usage:
+ * <pre>
+ * {@code
+ *   new ThreadExports().register();
+ * }
+ * </pre>
+ * Example metrics being exported:
+ * <pre>
+ *   jvm_threads_count{type="total"} 300
```
brian-brazil: This isn't consistent with the rest of the file, the _count suffix should be avoided outside of summaries/histograms, different labelnames is strongly discouraged and a total label such as this will cause problems.

I'd suggest removing this as we don't do this in other collectors.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+    this.threadBean = threadBean;
+  }
+
+  void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
+    List<MetricFamilySamples.Sample> threads = new ArrayList<MetricFamilySamples.Sample>();
+
+    threads.add(
+            new MetricFamilySamples.Sample(
+                    "jvm_threads",
+                    Arrays.asList("type"),
```
brian-brazil: Type is a bad label name, it's very generic. I'd suggest having this as a single gauge called `jvm_daemon_threads`

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+                    Arrays.asList("daemon"),
+                    threadBean.getDaemonThreadCount()));
+    threads.add(
+            new MetricFamilySamples.Sample(
+                    "jvm_threads",
+                    Arrays.asList("type"),
+                    Arrays.asList("nondaemon"),
+                    threadBean.getThreadCount() - threadBean.getDaemonThreadCount()));
```
brian-brazil: Generally don't do math in clients, subtraction is particularly risk due to the potential for negative values due to races.

I'd suggest exposing what the bean exposes more directly,  and have one metrics for each of the relevant calls.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+                    "jvm_threads",
+                    Type.GAUGE,
+                    "Thread count of a JVM",
+                    threads));
+
+    List<MetricFamilySamples.Sample> peakThreads = new ArrayList<MetricFamilySamples.Sample>();
+    peakThreads.add(
+            new MetricFamilySamples.Sample(
+                    "jvm_threads_peak",
+                    Arrays.asList("type"),
```
brian-brazil: This label doesn't add anything, and makes using this metric more difficult. The peak is already in the metric name.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+    List<MetricFamilySamples.Sample> startedThreads = new ArrayList<MetricFamilySamples.Sample>();
+    startedThreads.add(
+            new MetricFamilySamples.Sample(
+                    "jvm_threads_started",
+                    Arrays.asList("type"),
+                    Arrays.asList("started"),
+                    threadBean.getTotalStartedThreadCount()));
+    sampleFamilies.add(
+            new MetricFamilySamples(
+                    "jvm_threads",
```
brian-brazil: This should match the name in the sample.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+                    Type.GAUGE,
```
brian-brazil: This is a counter, and by convention the name should end in _total. jvm_threads_created_total should do it.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,99 @@
...
+                    "jvm_threads_peak",
+                    Type.GAUGE,
+                    "Peak thread count of a JVM",
+                    peakThreads));
+
+    List<MetricFamilySamples.Sample> startedThreads = new ArrayList<MetricFamilySamples.Sample>();
+    startedThreads.add(
+            new MetricFamilySamples.Sample(
+                    "jvm_threads_started",
+                    Arrays.asList("type"),
```
brian-brazil: This label doesn't add anything.

simpleclient_hotspot/src/test/java/io/prometheus/client/hotspot/ThreadExportsTest.java
```diff
@@ -0,0 +1,66 @@
+package io.prometheus.client.hotspot;
+
+import io.prometheus.client.Collector;
+import io.prometheus.client.CollectorRegistry;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.lang.management.MemoryMXBean;
```
brian-brazil: You're not using the memory imports

brian-brazil:
It'll be useful to get this in, you may find https://docs.google.com/document/d/1JapuiRbp-XoyECgl2lPdxITrhm5IyCUq9iA_h6jp3OY/edit useful as general background.

simpleclient_hotspot/src/test/java/io/prometheus/client/hotspot/ThreadExportsTest.java
```diff
@@ -0,0 +1,66 @@
...
+    collectorUnderTest = new ThreadExports(mockThreadsBean).register(registry);
+  }
+
+  @Test
+  public void testThreadPools() {
+    assertEquals(
+            200L,
+            registry.getSampleValue(
+                    "jvm_threads",
+                    new String[]{"type"},
```
Danm72: This is used a few times, could benefit from being a const or defined in the set up. Depends if its necessary

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -0,0 +1,89 @@
...
+ * <p>
+ * Example usage:
+ * <pre>
+ * {@code
+ *   new ThreadExports().register();
+ * }
+ * </pre>
+ * Example metrics being exported:
+ * <pre>
+ *   jvm_total_threads{} 300
```
brian-brazil: These don't all match the code

robdrysdale:
Bump

brian-brazil:
This looks good, can you squash your commits please?

Github only sends me a notification if you comment, not if you push a new version of code.

robdrysdale:
Commits squashed.

brian-brazil:
Thanks!

PR #104, opened 2016-02-13
remove runtime dependency on simpleclient_servlet. fixes #103

### LINKED ISSUE #103: simpleclient_hotspot should not depend on simpleclient_servlet
It seems like the dependency is not necessary.
I use the simpleclient together with [undertow core](https://github.com/undertow-io/undertow) without servlets, so this is a mild inconvenience.

Right now I do something like this:

```xml
<dependency>
    <artifactId>simpleclient_hotspot</artifactId>
    <version>${prometheus.version}</version>
    <exclusions>
        <exclusion>
            <artifactId>simpleclient_servlet</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

zimmi:
Turns out the ExampleExporter in the test packages is using simpleclient_servlet.
So I just changed the scope to `test`.

### DIFF
--- a/simpleclient_hotspot/pom.xml
+++ b/simpleclient_hotspot/pom.xml
@@ -39,13 +39,14 @@
             <artifactId>simpleclient</artifactId>
             <version>0.0.14-SNAPSHOT</version>
         </dependency>
+        
+        <!-- Test Dependencies Follow -->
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #105, opened 2016-02-25
Handle Gauges that return both Numbers and Booleans

tomcz:
Allows us to track Gauges that, for example, reflect circuit-breaker status.


brian-brazil:
Thanks!

PR #106, opened 2016-03-29
Expose Summary.Child.Value and Histogram.Child.Value fields.

pstibrany:
This is a trivial change to make Summary.labels().get() and Histogram.labels().get() somewhat useful for clients, that want to see current values stored in the metrics.


brian-brazil:
Thanks!

PR #107, opened 2016-04-04
Clarify the deprecation of methods using 'instance' parameter

pkubowicz:
- remove copy-and-paste: use the string "instance" in only 1 place
- expose method instanceGroupingKey() so that library users
  don't need to hard-code the string "instance" in their code
- provide a replacement for deprecated methods in their JavaDoc
- remove HashMap as its usage was inefficient

### DIFF
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -9,8 +9,9 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.net.URL;
 import java.net.URLEncoder;
-import java.util.Map;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Map;
 
 import java.io.BufferedWriter;
 import java.io.IOException;
@@ -54,7 +55,7 @@ import java.io.OutputStreamWriter;
 public class PushGateway {
 
   private final String address;
-  private final static int SECONDS_PER_MILLISECOND = 1000;
+  private static final int SECONDS_PER_MILLISECOND = 1000;
   /**
    * Construct a Pushgateway, with the given address.
    * <p>
@@ -182,12 +183,11 @@ public class PushGateway {
    * Pushes all metrics in a registry, replacing all those with the same job and instance.
    * <p>
    * This uses the POST HTTP method.
+   * @deprecated use {@link #push(CollectorRegistry, String, Map)}
   */
   @Deprecated
   public void push(CollectorRegistry registry, String job, String instance) throws IOException {
-    Map<String, String> groupingKey = new HashMap<String, String>();
-    groupingKey.put("instance", instance);
-    doRequest(registry, job, groupingKey, "POST");
+    push(registry, job, Collections.singletonMap("instance", instance));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


brian-brazil:
It's a little more complicated than this, it's not such much that these methods are deprecated than that pushgateway was exposing an API that required an instance to be specified - which is not a good practice for batch jobs.
Pointing users to this function could lead users to continue to provide an instance label where in reality they shouldn't be providing one at all. What you're proposing is not a replacement, this is a utility that could be used with the replacement.
I'm also not sure we should be providing a function for what is a single line of code.

pkubowicz:
Can you explain the intended use of pushing then? If I don't specify the instance, the current Java client code will cause pushing from one instance to wipe metrics of all other instances from the Push Gateway server. Such usage clearly makes no sense.

> I'm also not sure we should be providing a function for what is a single line of code.

"instance" grouping key has a special meaning in Prometheus. You should hide it from library users, otherwise they will hard-code it in their code and you will never be able to change this string. Hide it in a method and you will be able to change the string by changing the library method.

brian-brazil:
> If I don't specify the instance, the current Java client code will cause pushing from one instance to wipe metrics of all other instances from the Push Gateway server. Such usage clearly makes no sense.

For the cluster-level batch jobs for which the pgw is designed, there should be no instance label. The job label is the main thing you'd use to differentiate, and additional grouping keys if needed (e.g. if you had a sharded batch job).

> You should hide it from library users, otherwise they will hard-code it in their code and you will never be able to change this string.

The instance label is widely throughout Prometheus and exposed to users, it's set in stone.

pkubowicz:
> For the cluster-level batch jobs for which the pgw is designed

I am using Push Gateway because Prometheus' pull mode is not applicable to my network topology. I don't have any 'batch jobs'. I want my metrics to look the same in Prometheus GUI, no matter whether I push them or pull them. How can I achieve this if I cannot set 'instance' label when I push?

brian-brazil:
> I am using Push Gateway because Prometheus' pull mode is not applicable to my network topology.

We generally recommend running Prometheus on the same network as what it's monitoring, to avoid crossing failure domains. If there's a NAT layer between your services and Prometheus (which is what I'm presuming here), that's too far away network wise to provide reliable monitoring.

> I want my metrics to look the same in Prometheus GUI, no matter whether I push them or pull them.

This is not possible, there's fundamental differences to the two approaches that'll show up in the scraped data and cause artifacts for push.

> How can I achieve this if I cannot set 'instance' label when I push?

You can set this label, there's nothing stopping you. We even provide the `instanceIPGroupingKey` convenience function to make it a bit easier for this sort of use case.

pkubowicz:
> You can set this label, there's nothing stopping you.

All methods taking `instance` are marked as deprecated, without any reason. Therefore I cannot be sure if they won't be removed in a next Java client version.

brian-brazil:
Those methods will be removed at some future point as they're for the old pushgateway API, however they are not the only way to have an instance label as part of the grouping key.

pkubowicz:
You mean the REST API? But by calling the deprecated method and by calling the method with a grouping key you are currently making the same REST request. So I don't understand what could be removed from the API.

brian-brazil:
The deprecated methods would be removed, as they come from a time when the pgw required job and instance to be provided for all pushes. These days only job is required.

pkubowicz:
Ok, what about the pull request?

I can make `instanceGroupingKey` private and change the deprecation JavaDoc to point to methods taking the `groupingKey` parameter.

brian-brazil:
I'd accept replacing the use of a hashmap with the inlined oneliner, but `instanceIPGroupingKey` needs to stay as-is as users may be mutating the result.

pkubowicz:
Please see the modified version.

brian-brazil:
Thanks!

PR #108, opened 2016-04-05
Cosmetic fixes to increase thread-safety and performance

pkubowicz:
- make all variables final if possible
- use Collections.singletonList() where possible
- reduce visiblity of elements clearly not intended to be public

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java
@@ -9,6 +9,7 @@ import java.io.FileNotFoundException;
 import java.lang.management.RuntimeMXBean;
 import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.logging.Logger;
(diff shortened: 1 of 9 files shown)


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java
```diff
@@ -45,17 +46,13 @@ public StandardExports() {
...
       this.osBean = osBean;
       this.runtimeBean = runtimeBean;
-      if (osBean instanceof UnixOperatingSystemMXBean) {
-        unix = true;
-      }
-      if (osBean.getName().indexOf("Linux") == 0) {
-        linux = true;
-      }
+      this.unix = osBean instanceof UnixOperatingSystemMXBean;
+      this.linux = osBean.getName().indexOf("Linux") == 0;
```
brian-brazil: this is getting a little towards the magic side, I'd put the RHS in parens

pkubowicz: Ok.

brian-brazil:
Thanks!

PR #110, opened 2016-04-06
Make some missed fields private and/or final

pkubowicz:
Cleans up 36d08fd1fd82598636f04

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -19,7 +19,7 @@ public abstract class Collector {
    * Return all of the metrics of this Collector.
    */
   public abstract List<MetricFamilySamples> collect();
-  public static enum Type {
+  public enum Type {
     COUNTER,
     GAUGE,
     SUMMARY,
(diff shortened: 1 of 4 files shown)


simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -140,9 +140,9 @@ public String toString() {
    */
   public static final double MILLISECONDS_PER_SECOND = 1E3;
 
-  protected static final Pattern METRIC_NAME_RE = Pattern.compile("[a-zA-Z_:][a-zA-Z0-9_:]*");
-  protected static final Pattern METRIC_LABEL_NAME_RE = Pattern.compile("[a-zA-Z_][a-zA-Z0-9_]*");
-  protected static final Pattern RESERVED_METRIC_LABEL_NAME_RE = Pattern.compile("__.*");
+  private static final Pattern METRIC_NAME_RE = Pattern.compile("[a-zA-Z_:][a-zA-Z0-9_:]*");
```
brian-brazil: These should probably be public.

pkubowicz: You have protected-level methods to perform those checks, you don't need access to the regexes.

brian-brazil: Ah right, it's been a while since I've wrote this code.

brian-brazil:
Thanks!

PR #111, opened 2016-04-06
Remove duplication in collect() of all built-in collectors

pkubowicz:
Also save some space by precomputing capacity of some arrays.
Also de-duplicate class TimeProvider

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
@@ -1,5 +1,6 @@
 package io.prometheus.client;
 
+import java.util.ArrayList;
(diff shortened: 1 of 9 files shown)


simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -142,6 +143,13 @@ protected void initializeNoLabelsChild() {
    */
   protected abstract Child newChild();
 
+  protected List<MetricFamilySamples> familySamplesList(Collector.Type type, List<MetricFamilySamples.Sample> samples) {
```
brian-brazil: This should probably be package rather than protected

pkubowicz: Ok.

PR #112, opened 2016-05-12
Use a map to store labels for samples

cristiangreco:
This PR refactors the Sample class in order to store labels into a Map instead of two distinct Lists for names and values.

By reducing the creation of small, short-lived objects (especially in DropwizardExports), this change could potentially improve memory efficiency by putting less pressure on gc.

There are other places in the codebase where labels could be stored as a Map instead of Lists, but I tried to keep this PR small in order to get it merged easily. :-)

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -4,6 +4,7 @@ package io.prometheus.client;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.regex.Pattern;
 
 /**
@@ -111,8 +112,26 @@ public abstract class Collector {
 
       @Override
       public String toString() {
-        return "Name: " + name + " LabelNames: " + labels.keySet() + " labelValues: " + labels.values() +
-          " Value: " + value;
+        StringBuilder sb = new StringBuilder();
+        sb.append(name);
+        if (labels != null && !labels.isEmpty()) {
+          sb.append('{');
+          final Map<String, String> sortedLabels = new TreeMap<String, String>(labels);
+          for (Map.Entry<String, String> label : sortedLabels.entrySet()) {
+            sb.append(label.getKey())
+                    .append('=')
+                    .append('\"')
+                    .append(escapeLabelValue(label.getValue()))
+                    .append("\",");
+          }
+          sb.append('}');
+        }
+        sb.append(' ').append(doubleToGoString(value));
+        return sb.toString();
+      }
+
+      private String escapeLabelValue(String s) {
+        return s.replace("\\", "\\\\").replace("\"", "\\\"").replace("\n", "\\n");
       }
     }
   }
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -1,8 +1,9 @@
 
 package io.prometheus.client;
 
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Pattern;
 
 /**
@@ -74,14 +75,18 @@ public abstract class Collector {
    */
     public static class Sample {
       public final String name;
-      public final List<String> labelNames;
-      public final List<String> labelValues;  // Must have same length as labelNames.
+      public final Map<String, String> labels;
       public final double value;
 
-      public Sample(String name, List<String> labelNames, List<String> labelValues, double value) {
+      public Sample(String name, double value) {
         this.name = name;
-        this.labelNames = labelNames;
-        this.labelValues = labelValues;
+        this.labels = Collections.EMPTY_MAP;
+        this.value = value;
+      }
+
+      public Sample(String name, Map<String, String> labels, double value) {
+        this.name = name;
+        this.labels = labels;
         this.value = value;
       }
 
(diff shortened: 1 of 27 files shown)


simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -91,24 +96,22 @@ public boolean equals(Object obj) {
...
+        hash = 37 * hash + labels.hashCode();
         long d = Double.doubleToLongBits(value);
         hash = 37 * hash + (int)(d ^ (d >>> 32));
         return hash;
       }
 
       @Override
       public String toString() {
-        return "Name: " + name + " LabelNames: " + labelNames + " labelValues: " + labelValues +
+        return "Name: " + name + " LabelNames: " + labels.keySet() + " labelValues: " + labels.values() +
```
brian-brazil: Are the keys and values here going to be in the same order?

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -112,8 +112,8 @@ public Double getSampleValue(String name, String[] labelNames, String[] labelVal
     for (Collector.MetricFamilySamples metricFamilySamples: Collections.list(metricFamilySamples())) {
       for (Collector.MetricFamilySamples.Sample sample: metricFamilySamples.samples) {
         if (sample.name.equals(name)
-            && Arrays.equals(sample.labelNames.toArray(), labelNames)
-            && Arrays.equals(sample.labelValues.toArray(), labelValues)) {
+            && Arrays.equals(sample.labels.keySet().toArray(), labelNames)
```
brian-brazil: What ordering is used here?

Might it make more sense to take in a map?

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -107,6 +109,17 @@ protected void initializeNoLabelsChild() {
   }
 
   /**
+   * Return label names and values as map.
+   */
+  protected Map<String, String> labelsMap(List<String> childLabels) {
+    Map<String, String> labels = new HashMap<String, String>();
```
brian-brazil: Isn't this going to be creating objects where we weren't creating them before?

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -19,11 +20,11 @@ public static void write004(Writer writer, Enumeration<Collector.MetricFamilySam
       writer.write("# TYPE " + metricFamilySamples.name + " " + typeString(metricFamilySamples.type) + "\n");
       for (Collector.MetricFamilySamples.Sample sample: metricFamilySamples.samples) {
         writer.write(sample.name);
-        if (sample.labelNames.size() > 0) {
+        if (sample.labels != null && !sample.labels.isEmpty()) {
           writer.write("{");
-          for (int i = 0; i < sample.labelNames.size(); ++i) {
+          for (Map.Entry<String, String> label : sample.labels.entrySet()) {
```
brian-brazil: This should be sorted for efficiency

brian-brazil:
I'm not sure this is saving much, and in general I wouldn't worry about optimising things that happen at export time as that's a relatively infrequent operation.

This changes also breaks all existing custom collectors, and I'd prefer to improve the API for custom collectors along the lines of what I did in Python rather than making minor improvements to what's there.

simpleclient/src/main/java/io/prometheus/client/Collector.java
cristiangreco: I see your point, there is no guaranteed iteration order for keySet() and values() (although if you look at the implementation in AbstractMap the entrySet() iterator is used in both cases).

A better solution could be to the change the string format and print the map as a single object.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
cristiangreco: Same as above, using a map would be a better choice for this test method.

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
cristiangreco: Of course, but I guess we've saved much more Lists here and there in the removed code. :-)

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
cristiangreco: The output can be sorted, but this extra operation won't improve efficiency. :-)

I just had a look at the [Python client](https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L64) and it outputs sorted labels, so we could do the same for better consistency, if you like the idea.

cristiangreco:
I see your point and you're right, these changes are going to break the api for subclasses. The problem here is that the fields in MetricFamilySamples and Sample classes shouldn't be public at all, as this prevents any implementation improvement.

cristiangreco:
I refactored the test methods in CollectorRegistry. The toString() format of Sample is now essentially equal to the export text format and outputs sorted labels.

felixbarny:
Why do you think a `HashMap` is more memory efficient than two `ArrayList`s? Isn't it rather the other way around? A `HashMap` contains two more references via the `HashMap#Node` class per entry than compared to an `ArrayList`. Also, you must use a `LinkedHashMap` because the insertion order is important for the graphite bridge whose memory usage is even higher.

brian-brazil:
There's unclear benefit here in exchange for user breakage, without superb benchmarks I don't think this makes sense.

PR #113, opened 2016-05-17
Add quantile support to Summary

cristiangreco:
This adds support for quantile calculation in Summaries without extra dependencies.

Basic unit tests are included. Benchmarks may be added if needed.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Summary.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Summary.java
@@ -41,24 +41,24 @@ import java.util.Map;
 public class Summary extends SimpleCollector<Summary.Child> {
 
   public static final String QUANTILE_LABEL = "quantile";
-  public static final double[] DEFAULT_QUANTILE_VALUES = new double[] { .5, .95, .98, .99, .999 };
 
   private final double[] quantiles;
 
   Summary(Builder b) {
     super(b);
     this.quantiles = b.quantiles;
+    initializeNoLabelsChild();
   }
 
   public static class Builder extends SimpleCollector.Builder<Builder, Summary> {
-    private double[] quantiles = DEFAULT_QUANTILE_VALUES;
+    private double[] quantiles;
 
     @Override
     public Summary create() {
       if (quantiles != null && quantiles.length > 0) {
         for (double q : quantiles) {
-          if (q < 0 || q > 1) {
-            throw new IllegalArgumentException("quantile value must be in interval [0, 1]");
+          if (!isValidQuantile(q)) {
+            throw new IllegalArgumentException("quantile value must be 0 <= q <= 1");
           }
         }
       }
@@ -69,6 +69,7 @@ public class Summary extends SimpleCollector<Summary.Child> {
         }
       }
 
+      dontInitializeNoLabelsChild = true;
       return new Summary(this);
     }
 
@@ -90,7 +91,8 @@ public class Summary extends SimpleCollector<Summary.Child> {
 
   @Override
   protected Child newChild() {
-    return new Child();
+    final boolean trackValues = quantiles != null && quantiles.length > 0;
+    return new Child(trackValues);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -39,15 +40,45 @@
  */
 public class Summary extends SimpleCollector<Summary.Child> {
 
+  public static final String QUANTILE_LABEL = "quantile";
+  public static final double[] DEFAULT_QUANTILE_VALUES = new double[] { .5, .95, .98, .99, .999 };
```
brian-brazil: The default should be no quantiles per the client library guidelines.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -39,15 +40,45 @@
...
 
   public static class Builder extends SimpleCollector.Builder<Builder, Summary> {
+    private double[] quantiles = DEFAULT_QUANTILE_VALUES;
+
     @Override
     public Summary create() {
+      if (quantiles != null && quantiles.length > 0) {
+        for (double q : quantiles) {
+          if (q < 0 || q > 1) {
+            throw new IllegalArgumentException("quantile value must be in interval [0, 1]");
```
brian-brazil: `[` means different things in different countries. Better to say 0<=x<=1.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -93,11 +124,80 @@ public double observeDuration() {
...
         this.sum = sum;
+        this.values = values;
       }
+
+      /**
+       * Return the φ-quantile estimation of the observed events.
+       * The approximate value is calculated by interpolation
+       * between adjacent elements in the array of observed values.
+       *
+       * @param φ a given quantile, in {@code [0, 1]}
```
brian-brazil: Can we stick with ascii please?

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -93,11 +124,80 @@ public double observeDuration() {
...
+       */
+      public double getQuantile(double φ) {
+        if (φ < 0 || φ > 1) {
+          throw new IllegalArgumentException("quantile value must be in interval [0, 1]");
+        }
+
+        final int length = values.length;
+
+        if (length == 0) {
+          return 0;
```
brian-brazil: Return NaN if there's no data.

simpleclient/src/main/java/io/prometheus/client/UniformSampling.java
```diff
@@ -0,0 +1,80 @@
...
+   * Process a new element.
+   * The first {@code size} elements observed are
+   * immediately elected as candidates for the random
+   * sample. Then, each subsequent observed element gets
+   * a chance to replace an existing candidate.
+   *
+   * @param value the observed element
+   */
+  public void add(double value) {
+    final int c = (int) count.incrementAndGet();
```
brian-brazil: You've a race here if there's a lookup before the set.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -93,11 +124,80 @@ public double observeDuration() {
...
+      }
+
+      /**
+       * Calculate an approximate quantile from the observed
+       * value at {@code index}.
+       *
+       * @param index the index in the array of observed values
+       * @return the approximate quantile value
+       */
+      private double approxQuantile(final double index) {
```
brian-brazil: Splitting this code makes it a bit tricky to verify that the preconditions are satisfied.

brian-brazil:
Thanks for the PR. For other quantiles we use a time-based approach, rather than one that's since the start of the process. I'd suggest looking at what the Go client does.

cristiangreco:
Thanks for the review Brian. If the sliding-window sampling is the only stopper for the merge I'll be happy to implement it (with the other changes you addressed above). :-)

brian-brazil:
If we can get it working similar to Go that'd be good, and ensure if there's no quantiles that you don't pay the performance cost for them.

brian-brazil:
#134 has been accepted with this support.

PR #116, opened 2016-05-22
Fix Javadoc comment for DropwizardExports#fromCounter()

### LINKED ISSUE #115: Wrong type for Dropwizard Metrics Counter?
The Javadoc comment for [`DropwizardExports#fromCounter(String, Counter)`](https://github.com/prometheus/client_java/blob/parent-0.0.14/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java#L27-L34) says "Export counter as prometheus counter" but the actual metric type being used is [Gauge](https://prometheus.io/docs/concepts/metric_types/#gauge).

Either the Javadoc comment is wrong or the implementation is using the wrong [Prometheus metric type](https://prometheus.io/docs/concepts/metric_types/).

joschi:
This PR fixes the incorrect Javadoc comment for [`DropwizardExports#fromCounter(String, Counter)`](https://github.com/prometheus/client_java/blob/parent-0.0.14/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java#L27-L34).

Fixes #115

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -25,7 +25,7 @@ public class DropwizardExports extends io.prometheus.client.Collector {
     }
 
     /**
-     * Export counter as prometheus counter.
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #117, opened 2016-05-24
Remove sum from Summary

felixbarny:
Just summing all snapshot values does not work with dropwizard metrics, because the values array does not hold all values but only a sample of recent values.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -68,10 +68,6 @@ public class DropwizardExports extends io.prometheus.client.Collector {
      *
      */
     List<MetricFamilySamples> fromSnapshotAndCount(String name, Snapshot snapshot, long count, double factor, String helpMessage) {
-        long sum = 0;
-        for (long i : snapshot.getValues()) {
-            sum += i;
-        }
         List<MetricFamilySamples.Sample> samples = Arrays.asList(
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.5"), snapshot.getMedian() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.75"), snapshot.get75thPercentile() * factor),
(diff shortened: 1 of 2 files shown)


brian-brazil:
@dopuskh3 

I could have sworn there was a way to get this data.

simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
```diff
@@ -68,19 +68,14 @@ private static String getHelpMessage(String metricName, Metric metric){
...
-        }
         List<MetricFamilySamples.Sample> samples = Arrays.asList(
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.5"), snapshot.getMedian() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.75"), snapshot.get75thPercentile() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.95"), snapshot.get95thPercentile() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.98"), snapshot.get98thPercentile() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.99"), snapshot.get99thPercentile() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.999"), snapshot.get999thPercentile() * factor),
-                new MetricFamilySamples.Sample(name + "_count", new ArrayList<String>(), new ArrayList<String>(), count),
-                new MetricFamilySamples.Sample(name + "_sum", new ArrayList<String>(), new ArrayList<String>(), sum * factor)
```
dopuskh3: Removing sum will make prometheus histogram useless. Don't we have a way to get this value elsewhere?

felixbarny:
Unfortunately, there actually isn't because the Timer only contains data from approximately the last 5 minutes.

One (very bad) estimation would be `timer.getSnapshot().getMean() * timer.getCount()`. But this value is equally incorrect as the mean is only representative for the last 5 minutes.

Do some prometheus query depend on the sum? Which and why?

dopuskh3:
My understanding is that it will work with uniform reservoirs (http://metrics.dropwizard.io/3.1.0/manual/core/#uniform-reservoirs) but not with exponentially decaying reservoirs (defaults for histogram and timer in dropwizard). In this case (uniform reservoir), getValues(1.0) should work.

felixbarny:
`getValue(1.0)` returns the max value (100th percentile). Summing the values returned by `getValues()` also does not work for the `UniformReservoir`, because it only holds 1028 representative values. This would work for `UniformReservoir`s: timer.getSnapshot().getMean() * timer.getCount(). But like you said, the default is the `ExponentiallyDecayingReservoir` where it is not possible to get the sum.

simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
brian-brazil: We still have the quantiles, so it just stops us getting the average.

brian-brazil:
So shall we go ahead and remove this then?

felixbarny:
I can't say how this would affect prometheus histogram queries if there is no sum but the current sum value is just plain wrong.

PR #118, opened 2016-05-25
Added hikaricp support

tokuhirom:


simpleclient_hikaricp/src/main/java/io/prometheus/client/hikaricp/HikariCPCollector.java
```diff
@@ -0,0 +1,39 @@
...
+    this.labelValues = Collections.singletonList(poolName);
+  }
+
+  @Override
+  public List<MetricFamilySamples> collect() {
+    return Arrays.asList(
+            createSample("hikaricp_active_connections", "Active connections", poolStats.getActiveConnections()),
+            createSample("hikaricp_idle_connections", "Idle connections", poolStats.getIdleConnections()),
+            createSample("hikaricp_pending_threads", "Pending threads", poolStats.getPendingThreads()),
+            createSample("hikaricp_total_connections", "Total connections", poolStats.getTotalConnections())
```
brian-brazil: Is this current connections, or connections created over all time?

simpleclient_hikaricp/src/main/java/io/prometheus/client/hikaricp/HikariCPMetricsTracker.java
```diff
@@ -0,0 +1,37 @@
...
+            .labelNames("pool")
+            .help("Connection timeout count")
+            .register();
+
+    this.counter = counter.labels(poolName);
+  }
+
+  @Override
+  public void recordConnectionAcquiredNanos(long elapsedAcquiredNanos) {
+    super.recordConnectionAcquiredNanos(elapsedAcquiredNanos);
```
brian-brazil: Why not capture these in a summary?

brian-brazil:
Have you asked hikaricp about adding direct instrumentation? That'd be much simpler for users as they'd get all this without any configuration, and there's already dropwizard integration.

simpleclient_hikaricp/src/main/java/io/prometheus/client/hikaricp/HikariCPMetricsTracker.java
tokuhirom: Oh. I don't know about summary. I added code to take these metrics.

tokuhirom:
https://github.com/brettwooldridge/HikariCP/issues/321
HikariCP's author was rejected the feature request :(

simpleclient_hikaricp/src/main/java/io/prometheus/client/hikaricp/HikariCPCollector.java
tokuhirom: It's the number of current connections.

brian-brazil:
I don't read that as a rejection, I see it that a PR would be considered. It's probably worth clarifying.

tokuhirom:
Oops. I miss understood. I'll try to send p-r to hikaricp repo.

simpleclient_hikaricp/src/main/java/io/prometheus/client/hikaricp/HikariCPCollector.java
brian-brazil: I'd suggest removing the `total_` from the name then, and clarifying the docstring.

tokuhirom:
I sen't p-r to prometheus repo.

brian-brazil:
That's in the app itself now.

PR #119, opened 2016-05-25
Make *buckets method of Builder add buckets, not replace them.

LuboVarga:
Now exponentialBuckets, linearBuckets and also buckets of Histogram.Builder adds buckets instead of replacing them.

Changed buckets type from double[] to Double[] due to simplicity and introduced collection usage in process of building them.
Moved sanitizing of buckets from builders create method into separate get method. Now is not possible to use Histogram constructor with builder without sanitized buckets.
Added sorting to buckets as part of build process.
Buckets with equal boundary does not make sense, co modified check function to do not allow them.
This code now does make sense:
```
Histogram h = new Histogram.Builder()
            .exponentialBuckets(16, 4, 6) // 16, 64, 256, 1024, 4096, 16384
            .buckets(32768) // adds one more bound
            .labelNames("rpcCommand")
            .help("Histogram of message sizes. It serves also as counter of incoming/outgoing messages and you can use it for making difference to see" +
                "how many messages are in processing stage.")
            .name(this.name + "_size")
            .namespace(this.namespace)
            .create();
```

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Histogram.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Histogram.java
@@ -54,6 +54,9 @@ public class Histogram extends SimpleCollector<Histogram.Child> {
   Histogram(Builder b) {
     super(b);
     buckets = b.getBuckets();
+    if (buckets.length < 1) {
+      throw new IllegalStateException("Histogram must have at least one bucket.");
+    }
     initializeNoLabelsChild();
   }
 
@@ -63,9 +66,6 @@ public class Histogram extends SimpleCollector<Histogram.Child> {
 
     @Override
     public Histogram create() {
-      if (bucketsList.isEmpty()) {
-          throw new IllegalStateException("Histogram must have at least one bucket.");
-      }
       for (String label: labelNames) {
         if (label.equals("le")) {
             throw new IllegalStateException("Histogram cannot have a label named 'le'.");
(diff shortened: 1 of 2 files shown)


simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -1,8 +1,6 @@
 package io.prometheus.client;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
brian-brazil: Imports should be explicit.

brian-brazil:
Thanks for writing this PR. I'm not sure this is the API we should expose to users as it's more complex and I can see it causing confusion. If a user does want to do a mix of linear and exponential, it's probably best that they do it by hand (and it's probably overengineering). The example you give can be done via the existing API.

I don't see the benefit of changing to `Double`, it just introduces the potential for null pointers.

LuboVarga:
I would use for example two linear (or normal) buckets, if I would like to have possibility to compute two different APDEX scores. It is more readable to have in code something like this:
```
.buckets(12, 48) // adds buckets for APDEX_12
.buckets(20, 80) // adds buckets for APDEX_20
```
than this:
```
// adds buckets for APDEX_12 and also APDEX_20. you can figure out which bucket belongs
// (is required for) to which APDEX
.buckets(12, 20, 48, 80)
```

Would it help, if it would be double, instead of Double? It is achievable.

brian-brazil:
The benefit there seems pretty minor, and comments will generally cover it.

We need to be careful when expanding the API, as any changes we make must be supported forever.

LuboVarga:
Ok, I get it. Thanks for review.

PR #124, opened 2016-06-25
Add reset method to the counter

Gouthamve:
While I know reset should be used very rarely, I think the option should exist for building custom exporters.


brian-brazil:
Sets and resets are not operations that a counter should support. You're probably looking to write a custom collector instead.

PR #125, opened 2016-07-05
Fix URL of Google Java Style

qqvpp:


shmaher2:
Hi

I have implemented generic SQL scrapper written in java for Prometheus. Can I add it to repository?

brian-brazil:
This repository is for the java client, exporters wouldn't go here. This would be linked from the docs repository, though we've already https://github.com/chop-dbhi/prometheus-sql listed.

PR #126, opened 2016-07-05
Added support for pushing with a timestamp

BasPH:
This allows to push to the Pushgateway with a timestamp.

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -12,6 +12,13 @@ public class TextFormat {
    * Write out the text version 0.0.4 of the given MetricFamilySamples.
    */
   public static void write004(Writer writer, Enumeration<Collector.MetricFamilySamples> mfs) throws IOException {
+    write004(writer, mfs, null);
+  }
+
+  /**
+   * Write out the text version 0.0.4 of the given MetricFamilySamples.
+   */
+  public static void write004(Writer writer, Enumeration<Collector.MetricFamilySamples> mfs, String timestamp) throws IOException {
     /* See http://prometheus.io/docs/instrumenting/exposition_formats/
      * for the output format specification. */
     for (Collector.MetricFamilySamples metricFamilySamples: Collections.list(mfs)) {
@@ -27,7 +34,11 @@ public class TextFormat {
           }
           writer.write("}");
         }
-        writer.write(" " + Collector.doubleToGoString(sample.value) + "\n");
+        writer.write(" " + Collector.doubleToGoString(sample.value));
+        if (timestamp != null && !timestamp.isEmpty()) {
+          writer.write(" " + timestamp);
+        }
+        writer.write("\n");
       }
     }
   }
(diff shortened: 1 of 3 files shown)


brian-brazil:
What's your use case for this?

There's no plans to support this until after https://github.com/prometheus/prometheus/issues/398 is implemented, as it'll likely be widely misused.

BasPH:
I'm pushing metrics from Spark jobs (e.g. num. records in a batch, current batch delay, num. waiting batches) to a Pushgateway every 15 seconds which is being scraped also every 15 seconds. The push and scrape interval are intentionally equal so that no metrics are being overwritten and thus lost in the Pushgateway.

The Spark jobs are ephemeral and metrics in the Pushgateway without timestamp will receive the scrape time's timestamp, which was inconvenient since the Prometheus server scrapes at any point in time in the given 15 second interval. I wanted to ensure the metrics are from fixed 15 second intervals.

This is also suggested in the [Pushgateway documentation](https://github.com/prometheus/pushgateway#about-timestamps) ("Essentially, if you push more often than every 5min, you could attach the time of pushing as a timestamp.") so I figured this is a logical thing to implement.

brian-brazil:
> I'm pushing metrics from Spark jobs 

This doesn't require timestamps.

> The Spark jobs are ephemeral and metrics in the Pushgateway without timestamp will receive the scrape time's timestamp, which was inconvenient since the Prometheus server scrapes at any point in time in the given 15 second interval.

This is event logging, not metrics. I recommend converting it to whatever the equivalent of a daemon is for Spark and scraping the job in the usual fashion, or choosing an event-logging monitoring system.

rweverwijk:
@brian-brazil The issue with spark jobs is that they typically run on a Hadoop cluster with possible 200+ servers. When a job is started you cannot predict the host where the job will start running. This makes scraping in the usual fashion (almost) impossible. Using the push gateway seems the right way to solve this issue.

>  This is event logging, not metrics.
The Spark Metrics @BasPH is mentioning just are some examples. Things like totalReceivedRecords, waitingBatches, totalCompletedBatches, schedulingDelay are not event related but are related to the state of an application.

Adding a timestamp to the metric makes sure that the metric is stored as a single sample in the store. When the Spark Job is down there is no update and from Prometheus side it will react the same as samples scraped in the usual fashion.

brian-brazil:
> When a job is started you cannot predict the host where the job will start running. 

This is a problem of service discovery.

Your real problem is that your jobs are too frequent and too short lived, which makes this event logging rather than metrics. I'd recommend doing the metrics a level up at the driver rather than the workers.

rweverwijk:
The jobs that we run are Spark Streaming jobs that are continuously running. The Metrics are pushed from the driver already, not from the workers.

We do Service Discovery for all the services where we connect to. The Spark jobs are no services but ETL processes and we really don't care where these are running as we don't have connections towards the jobs.

If we want to scrape the Spark job by using the JMX for example we will have the following issue: The drivers can be running on different machines, but also multiple on the same machine. Therefore we not only need to discover the machine where a job is running, but we need to bind to different ports as well to expose the metrics. In other words a hell of a job to make stable, dynamic and reliable.

We thought to have found a great solution with the push gateway to have a single place to scrape. This fits the dynamic environment where we are working with and is running stable in production now for about a month.

brian-brazil:
In this case what you want is service discovery so that Prometheus can find the drivers. The push gateway is not appropriate for monitoring daemons.

PR #128, opened 2016-07-12
Implemented Vert.x simple client support

pmlopes:
This is a simple implementation of a Vert.x Web handler for the the simple client

@brian-brazil and @beorn7 I've followed the servlet code in order to implement the Vert.x code, please let me know if this is the correct way.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: java
 
+jdk:
+  - oraclejdk8
+
 script:
 - mvn test
 - mvn javadoc:aggregate
--- a/simpleclient_vertx/pom.xml
+++ b/simpleclient_vertx/pom.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <parent>
+        <groupId>io.prometheus</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.0.15-SNAPSHOT</version>
+    </parent>
+
+    <groupId>io.prometheus</groupId>
+    <artifactId>simpleclient_vertx</artifactId>
+    <packaging>bundle</packaging>
+
+    <name>Prometheus Java Simpleclient Vert.x</name>
+    <description>
+        Vert.x  Web Handler exporter for the simpleclient.
+    </description>
+
+    <licenses>
+        <license>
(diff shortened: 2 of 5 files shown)


simpleclient_vertx/pom.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
```
brian-brazil: We support Java 6

simpleclient_vertx/src/main/java/io/prometheus/client/vertx/MetricsHandler.java
```diff
@@ -0,0 +1,81 @@
+package io.prometheus.client.vertx;
+
+import io.prometheus.client.CollectorRegistry;
+import io.prometheus.client.exporter.common.TextFormat;
+import io.vertx.core.Handler;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.ext.web.RoutingContext;
+
+import java.io.*;
```
brian-brazil: All imports should be listed.

simpleclient_vertx/pom.xml
pmlopes: I'm fine with lowering it to 6, however vert.x requires 8 to run. In this case I can use abstract inner classes. What would you recommend?

brian-brazil: If it only works on 8 then that's okay, but avoid using more 8 syntax than you have to avoid confusion.

pmlopes:
The only java 8 language feature in the code is a call to a default static method on an interface, this interfaces are part of vert.x core and it is the factory for buffers, and on the example the factory for vertx itself and the web router.

```java
Vertx.vertx()
Buffer.buffer()
Router.create(vertx)
```

I can't really avoid them. The only i can avoid is the method reference on the example which i can wrap in a annonymous inner class.

simpleclient_vertx/src/main/java/io/prometheus/client/vertx/MetricsHandler.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public MetricsHandler(CollectorRegistry registry) {
+    this.registry = registry;
+  }
+
+  @Override
+  public void handle(RoutingContext ctx) {
+
+    try {
+
```
brian-brazil: You can trim the vertical whitespace a bit

brian-brazil:
Can you squash your commits please?

pmlopes:
Squashed!

brian-brazil:
The CI isn't liking your use of java 8, you'll need to tweak travis.yml to use jdk8 for this module.

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: java
 
+jdk:
+  - oraclejdk8
```
brian-brazil: Can we still make sure the rest works with 6/7? I want to avoid an accidental regression here, as 2/3 of our users still are on 6.

pmlopes: The top level pom forces 1.6:

https://github.com/prometheus/client_java/blob/master/pom.xml#L63

and only the vert.x one allows 1.8

brian-brazil:
Thanks!

PR #129, opened 2016-07-13
Implemented an vert.x based asynchronous pushgateway client

pmlopes:
This is a simple implementation of an asynchronous pushgateway client using Vert.x.

@brian-brazil and @beorn7 I've extended the base PushGateway class to make it async. I've added it as a seperate module so the original code is untouched and no extra dependencies get pulled. The existing test suite is also run on the async client.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,7 @@
         <module>simpleclient_spring_boot</module>
         <module>simpleclient_jetty</module>
         <module>simpleclient_vertx</module>
+        <module>simpleclient_pushgateway_vertx</module>
         <module>benchmark</module>
     </modules>
     
--- a/simpleclient_pushgateway_vertx/pom.xml
+++ b/simpleclient_pushgateway_vertx/pom.xml
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.prometheus</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.0.15-SNAPSHOT</version>
+    </parent>
+
+    <groupId>io.prometheus</groupId>
+    <artifactId>simpleclient_pushgateway_vertx</artifactId>
+    <packaging>bundle</packaging>
+
+    <name>Prometheus Vert.x Async Simpleclient Pushgateway</name>
+    <description>
+        Async Pushgateway exporter for the simpleclient.
+    </description>
+
+    <licenses>
+        <license>
+            <name>The Apache Software License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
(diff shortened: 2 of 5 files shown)


brian-brazil:
How does the standard pushgateway module not meet your needs?

pmlopes:
The standard gateway is using URLConnection which is a blocking API. Using it on a single thread programming model (reactive) will block the whole application making it unresponsive. Using the reactive model the single thread is not blocked so more work can be performed in parallel.

brian-brazil:
What use case do you have where it matters?

The pushgateway is typically used at the end of the batch job, and you want to block until the pushgateway has accepted the request.

pmlopes:
I don't agree that I'd want to block until the gateway accepted the request. Say that I've run a batch job and at the end i want to push my metrics to the push gateway and perform some logging, clean up, etc... If the job is complete i can fire an async push to the gateway and start logging/cleaning, if an error occurs while pushing, my async exception handler will be called and I can either log it or terminate.

The issue here is once you're working with a reactive framework the rule of thumb is that you should never block the main loop. If we do that we might also be blocking any incoming input as an example.

brian-brazil:
That's going to be a very rare use case, and such a feature is likely to be misused and abused far more likely than it is used correctly.

For example the previous person requesting something along these lines wanted to do regular async pushes of metrics, which is an incorrect usage.

pmlopes:
I've read the documentation and understand you point but let me give some a bit of context how vert.x works. Vert.x is a reactive toolkit for building distributed applications. When dealing with traditional web applications everything is fine since one can use the `simpleclient_vertx` handler, no questions there.

However if one is using vert.x for its distributed capabilities then integrating with prometheus gets a bit more tricky.

In vertx you can create a cluster and scale up or down by adding nodes to the cluster at runtime, there is no central authority so you don't know before starting prometheus where all nodes reside which leads to the case you can't configure prometheus as expected.

On top of that vert.x is a message driven toolkit, so if you want to scale your code you deploy more nodes on the cluster and send a message, at that time there is no knowledge which node will actually perform the task, so it must be that node that pushes the results of the job/metrics to prometheus.

Due to the high availability in vert.x if a specific node crashes it will be respawn on another node so no assumptions about ip addresses can be made.

Finally vert.x eventbus is not HTTP based so again one cannot assume that there is a server running on each node.

I don't have a specific use case to say this should be done this way (or not) as a Vert.x developer I just want to give Vert.x users a simple way to integrate with prometheus.

brian-brazil:
It sounds like you should be using normal direct instrumentation with service discovery. The pushgateway is only for batch jobs, see https://prometheus.io/docs/practices/pushing/

pmlopes:
Vert.x has already a metrics collecting SPI, one implementation is Hawkular:

http://vertx.io/docs/vertx-hawkular-metrics/java/#_vert_x_core_tools_metrics

With time and interest from the community we should implement a exporter (for the same core metrics using prometheus), that would be the next step ;)

This PR was all about having an valid async API for the push gateway. Vert.x API is async so having a compatible push client makes sense to me. Yes we can work around the blocking part of the standard push client by wrapping it with:

http://vertx.io/docs/apidocs/io/vertx/core/Vertx.html#executeBlocking-io.vertx.core.Handler-boolean-io.vertx.core.Handler-

My idea was that since we're already using Netty for high performance IO why not going the whole way and have it compatible with Vert.x too.

brian-brazil:
As it's all JVM, a collector would be better than an exporter.

I don't think I'll be accepting any async APIs for the pushgateway, as all requests for it so far were for incorrect usage. We try as a project to point users down the right path and thus avoid features that may be useful for advanced users but would be widely misused by normal users.

PR #132, opened 2016-07-20
Add a method to fetch all the samples starting with a given prefix

grandbora:
This pr makes testing on CollectorRegistry easier for new developers who are not familiar with internals of different collectors.

The problem we have at SoundCloud is that when new developers want to do assertions on various collectors with the existing `getSampleValue` methods, they need know the details of how each collector works. For example summaries suffix the given name, and histograms adds the `le` label. Even our internal library is adding a label to every collector by default. These are abstracted away from the application developer but when it comes to testing they need to be known.

The method introduced in this pr is a step towards an easier set up. In `listSampleValuesOfPrefix` by Using the collectors name as a prefix developers can access to all samples without having to know the labels and exact metric names. At least this can help developers do the right assertions, or in some cases we can do assertions on the return value of this method.

cc @brian-brazil @beorn7

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -1,10 +1,12 @@
 package io.prometheus.client;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.Enumeration;
 import java.util.Iterator;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
What assertions does this permit that the existing call doesn't?

grandbora:
Point is that it is allowing you not to know about the internal details of collectors. 

A developer who knows how collectors work internally can easily write their tests with the existing interface. However new developers always bump into the problem of "I expect value X but got null", because they don't know that metric name is different than collector name and/or there are other labels added to the metrics and so on.

Aim here is helping new developers get going with java client as fast as possible.

brian-brazil:
Couldn't they figure that out from the /metrics output? If they can't work with that I'm not sure adding an additional API method will help, and you're also losing whatever type information there was.

grandbora:
The /metrics endpoint requires you to run the application and send a request that will increase the metric that you are interested in. This is a friction, ideally we would like to be able write the tests without having to do that. 

I don't particularly like the idea of adding these test methods. If you have any other suggestions I am happy to follow.

brian-brazil:
To step back a bit, why are a significant number of developers you writing unittests? Generally instrumentation should be treated like logging and not unittested, outside of a few core places like RPC libraries where the metrics are an important feature in and of themselves.

grandbora:
That's a good question. At the moment we are writing unit tests for our metrics to makes sure we are reporting the right metric and label names. We have some conventions that need to be adhered by various distributed systems, otherwise our generic dashboards and other aggregations that we do may break.

I see your point with instrumentation should not be tested. But also I can imagine a new developer changes a label name on a metric without being aware of and deploys the application because no tests failed. Then we see a drop in our business dashboards or on another graph. Now we need to go through all of the systems that report that metric and see what has changed. That's also a situation we wouldn't want to be in.

brian-brazil:
A developer that changes the label on the metric will likely update the unittest, and then proceed to break things.


I'm not sure there's anything practical that can be done here for devs who don't have any knowledge of the data model.

grandbora:
> A developer that changes the label on the metric will likely update the unit test

I was thinking about the developers who do that change unintentionally. We have developers from different backgrounds and would like to make systems fool proof.

I am happy to put effort in making testing easier for java client, if you show some direction or have any ideas.

brian-brazil:
#74 is the only outstanding testing bug we have, I can't think of anything that'd really help here as it's an education issue rather than an API issue.

By the way:

> Even our internal library is adding a label to every collector by default. 

This is an anti-pattern. Target labels should be handled on the Prometheus end. This also implies that you've wrapped the client library, please don't do that.

PR #134, opened 2016-08-01
Added quantiles to summary metric (same algorithm as in client_golang).

fstab:
As mentioned in the comment on CKMSQuantiles: The algorithm isn't implemented by me, I just found the implementation on github.com/Netflix/ocelli (apache license) and integrated it. The Netflix implementation looks good, though.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -6,3 +6,6 @@ Boxever Ltd. (http://www.boxever.com/).
 
 This product includes software developed at
 SoundCloud Ltd. (http://soundcloud.com/).
+
+This product includes software developed as part of the
+Ocelli project by Netflix Inc. (https://github.com/Netflix/ocelli/).
--- a/simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
@@ -1,0 +1,293 @@
+package io.prometheus.client;
+
+// Copied from https://raw.githubusercontent.com/Netflix/ocelli/master/ocelli-core/src/main/java/netflix/ocelli/stats/CKMSQuantiles.java
+// Revision d0357b8bf5c17a173ce94d6b26823775b3f999f6 from Jan 21, 2015.
+//
+// This is the original code except for the following modifications:
+//
+//  - Changed the type of the observed values from int to double.
+//  - Removed the Quantiles interface and corresponding @Override annotations.
+//  - Changed the package name.
+//  - Make get() return NaN when no sample was observed.
+//  - Make class package private
+
+/*
+ Copyright 2012 Andrew Wang (andrew@umbrant.com)
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.ListIterator;
+
+/**
+ * Implementation of the Cormode, Korn, Muthukrishnan, and Srivastava algorithm
+ * for streaming calculation of targeted high-percentile epsilon-approximate
+ * quantiles.
+ * 
+ * This is a generalization of the earlier work by Greenwald and Khanna (GK),
+ * which essentially allows different error bounds on the targeted quantiles,
+ * which allows for far more efficient calculation of high-percentiles.
+ * 
+ * 
+ * See: Cormode, Korn, Muthukrishnan, and Srivastava
+ * "Effective Computation of Biased Quantiles over Data Streams" in ICDE 2005
+ * 
+ * Greenwald and Khanna,
+ * "Space-efficient online computation of quantile summaries" in SIGMOD 2001
+ * 
+ */
+class CKMSQuantiles {
+    /**
+     * Total number of items in stream.
+     */
+    private int count = 0;
+
+    /**
+     * Used for tracking incremental compression.
+     */
+    private int compressIdx = 0;
+
+    /**
+     * Current list of sampled items, maintained in sorted order with error
+     * bounds.
+     */
+    protected LinkedList<Item> sample;
+
+    /**
+     * Buffers incoming items to be inserted in batch.
+     */
+    private double[] buffer = new double[500];
+
+    private int bufferCount = 0;
+
+    /**
+     * Array of Quantiles that we care about, along with desired error.
+     */
+    private final Quantile quantiles[];
+
+    public CKMSQuantiles(Quantile[] quantiles) {
+        this.quantiles = quantiles;
+        this.sample = new LinkedList<Item>();
+    }
+
+    /**
+     * Add a new value from the stream.
+     * 
+     * @param value
+     */
+    public synchronized void insert(double value) {
+        buffer[bufferCount] = value;
+        bufferCount++;
+
+        if (bufferCount == buffer.length) {
+            insertBatch();
+            compress();
+        }
+    }
+
+    /**
+     * Get the estimated value at the specified quantile.
+     * 
+     * @param q
+     *            Queried quantile, e.g. 0.50 or 0.99.
+     * @return Estimated value at that quantile.
+     */
+    public synchronized double get(double q) {
+        // clear the buffer
+        insertBatch();
+        compress();
+
+        if (sample.size() == 0) {
+            return Double.NaN;
+        }
+
+        int rankMin = 0;
+        int desired = (int) (q * count);
+
+        ListIterator<Item> it = sample.listIterator();
+        Item prev, cur;
+        cur = it.next();
+        while (it.hasNext()) {
+            prev = cur;
+            cur = it.next();
+
+            rankMin += prev.g;
+
+            if (rankMin + cur.g + cur.delta > desired
+                    + (allowableError(desired) / 2)) {
+                return prev.value;
+            }
+        }
+
+        // edge case of wanting max value
+        return sample.getLast().value;
+    }
+
+    /**
+     * Specifies the allowable error for this rank, depending on which quantiles
+     * are being targeted.
+     * 
+     * This is the f(r_i, n) function from the CKMS paper. It's basically how
+     * wide the range of this rank can be.
+     * 
+     * @param rank
+     *            the index in the list of samples
+     */
+    private double allowableError(int rank) {
+        // NOTE: according to CKMS, this should be count, not size, but this
+        // leads
+        // to error larger than the error bounds. Leaving it like this is
+        // essentially a HACK, and blows up memory, but does "work".
+        // int size = count;
+        int size = sample.size();
+        double minError = size + 1;
+
+        for (Quantile q : quantiles) {
+            double error;
+            if (rank <= q.quantile * size) {
+                error = q.u * (size - rank);
+            } else {
+                error = q.v * rank;
+            }
+            if (error < minError) {
+                minError = error;
+            }
+        }
+
+        return minError;
+    }
+
+    private boolean insertBatch() {
+        if (bufferCount == 0) {
+            return false;
+        }
+
+        Arrays.sort(buffer, 0, bufferCount);
+
+        // Base case: no samples
+        int start = 0;
+        if (sample.size() == 0) {
+            Item newItem = new Item(buffer[0], 1, 0);
+            sample.add(newItem);
+            start++;
+            count++;
+        }
+
+        ListIterator<Item> it = sample.listIterator();
+        Item item = it.next();
+
+        for (int i = start; i < bufferCount; i++) {
+            double v = buffer[i];
+            while (it.nextIndex() < sample.size() && item.value < v) {
+                item = it.next();
+            }
+
+            // If we found that bigger item, back up so we insert ourselves
+            // before it
+            if (item.value > v) {
+                it.previous();
+            }
+
+            // We use different indexes for the edge comparisons, because of the
+            // above
+            // if statement that adjusts the iterator
+            int delta;
+            if (it.previousIndex() == 0 || it.nextIndex() == sample.size()) {
+                delta = 0;
+            } 
+            else {
+                delta = ((int) Math.floor(allowableError(it.nextIndex()))) - 1;
+            }
+
+            Item newItem = new Item(v, 1, delta);
+            it.add(newItem);
+            count++;
+            item = newItem;
+        }
+
+        bufferCount = 0;
+        return true;
+    }
+
+    /**
+     * Try to remove extraneous items from the set of sampled items. This checks
+     * if an item is unnecessary based on the desired error bounds, and merges
+     * it with the adjacent item if it is.
+     */
+    private void compress() {
+        if (sample.size() < 2) {
+            return;
+        }
+
+        ListIterator<Item> it = sample.listIterator();
+        int removed = 0;
+
+        Item prev = null;
+        Item next = it.next();
+
+        while (it.hasNext()) {
+            prev = next;
+            next = it.next();
+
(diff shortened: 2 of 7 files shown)


simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -36,17 +36,64 @@
...
     @Override
     public Summary create() {
-      return new Summary(this);
+        for (String label: labelNames) {
+          if (label.equals("quantile")) {
+            throw new IllegalStateException("Summary cannot have a label named 'quantile'.");
+          }
+        }
+        if (quantiles.isEmpty()) {
+            // use defaults
```
brian-brazil: Per the client guidelines, the default should be no quantiles.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -93,10 +140,12 @@ public double observeDuration() {
     public static class Value {
       public final double count;
       public final double sum;
+      public final CKMSQuantiles quantiles;
```
brian-brazil: Could this be modified by the user?

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -154,6 +210,13 @@ public Timer startTimer() {
     List<MetricFamilySamples.Sample> samples = new ArrayList<MetricFamilySamples.Sample>();
     for(Map.Entry<List<String>, Child> c: children.entrySet()) {
       Child.Value v = c.getValue().get();
+      List<String> labelNamesWithQuantile = new ArrayList<String>(labelNames);
+      labelNamesWithQuantile.add("quantile");
+      for (CKMSQuantiles.Quantile q : this.quantiles) {
```
brian-brazil: Are these sorted?

brian-brazil:
One of the requirements that we have for quantiles is that it's a value over a recent window, not something that goes back forever. When there's no data recently NaN should be returned.

I'd suggest taking the Go client's implementation.

fstab:
Hi Brian, thanks a lot for your comments. I implemented each of your points:

- make Quantiles optional (you are right, it is specified that way in the guidelines. however, the client_golang has default values for quantiles)
- make member variables in Value class private to make sure it's for internal use only
- make sure the Quantiles are sorted (before they were in the order that was used when building the summary)
- implement a sliding time window for quantiles (I wrapped the CKMSQuantiles class into a TimeWindowQuantiles class)

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -0,0 +1,292 @@
+package io.prometheus.client;
+
+// Copied from https://raw.githubusercontent.com/Netflix/ocelli/master/ocelli-core/src/main/java/netflix/ocelli/stats/CKMSQuantiles.java
```
brian-brazil: This providence of this code should be added to the NOTICE file.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -0,0 +1,292 @@
...
+ * 
+ * 
+ * See: Cormode, Korn, Muthukrishnan, and Srivastava
+ * "Effective Computation of Biased Quantiles over Data Streams" in ICDE 2005
+ * 
+ * Greenwald and Khanna,
+ * "Space-efficient online computation of quantile summaries" in SIGMOD 2001
+ * 
+ */
+public class CKMSQuantiles {
```
brian-brazil: This should be internal.

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -186,6 +186,9 @@ public static String doubleToGoString(double d) {
     if (d == Double.NEGATIVE_INFINITY) {
       return "-Inf";
     }
+    if (d == Double.NaN) {
```
brian-brazil: Does a straight comparison work here? NaN isn't usually equal to itself.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -50,50 +51,80 @@
...
  * }
  * </pre>
- * The values above are the default values, so if you don't call <tt>quantile()</tt>,
- * the quantiles above will be used implicitly.
+ *
+ * The quantiles are calculated over a sliding window of time. There are two options to configure this time window:
+ * <ul>
+ *   <li>maxAgeMillis(long): Set the duration of the time window is, i.e. how long observations are kept before they are discarded.
+ *       Default is 10 minutes.
+ *   <li>nBuckets(int): Set the number of buckets used to implement the sliding time window. If your time window is 10 minuts, and you have nBuckets=5,
```
brian-brazil: minutes

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -50,50 +51,80 @@
  *             .register();
  * }
  * </pre>
- * The values above are the default values, so if you don't call <tt>quantile()</tt>,
- * the quantiles above will be used implicitly.
+ *
+ * The quantiles are calculated over a sliding window of time. There are two options to configure this time window:
+ * <ul>
+ *   <li>maxAgeMillis(long): Set the duration of the time window is, i.e. how long observations are kept before they are discarded.
```
brian-brazil: Let's keep to seconds

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -1,8 +1,9 @@
 package io.prometheus.client;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import io.prometheus.client.CKMSQuantiles.Quantile;
+
+import java.util.*;
```
brian-brazil: Imports should be explicit

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -50,50 +51,80 @@
...
-          throw new IllegalArgumentException("Error " + error + " invalid: Expected percentage between 0.0 and 1.0.");
-        }
-        quantiles.add(new CKMSQuantiles.Quantile(quantile, error));
-        return this;
+    public Builder quantile(double quantile, double error) {
+      if (quantile < 0.0 || quantile > 1.0) {
+        throw new IllegalArgumentException("Quantile " + quantile + " invalid: Expected percentage between 0.0 and 1.0.");
+      }
+      if (error < 0.0 || error > 1.0) {
+        throw new IllegalArgumentException("Error " + error + " invalid: Expected percentage between 0.0 and 1.0.");
```
brian-brazil: A percentage is between 0 and 100

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -50,50 +51,80 @@
...
+
+    public Builder maxAgeMillis(long maxAgeMillis) {
+      if (maxAgeMillis <= 0) {
+        throw new IllegalArgumentException("maxAgeMillis cannot be " + maxAgeMillis);
+      }
+      this.maxAgeMillis = maxAgeMillis;
+      return this;
+    }
+
+    public Builder nBuckets(int nBuckets) {
```
brian-brazil: This is called AgeBuckets in Go

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -137,15 +168,15 @@ public double observeDuration() {
    * {@link SimpleCollector#remove} or {@link SimpleCollector#clear}.
    */
   public static class Child {
-    public static class Value {
-      public final double count;
-      public final double sum;
-      public final CKMSQuantiles quantiles;
+    private static class Value {
```
brian-brazil: This is meant to be public

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -137,15 +168,15 @@ public double observeDuration() {
...
+      private final double count;
+      private final double sum;
+      private final TimeWindowQuantiles optionalQuantiles; // optional means can be null.
```
brian-brazil: This should probably be a map of the output quantiles.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -155,12 +186,12 @@ private Value(double count, double sum, CKMSQuantiles quantiles) {
     // This should be reevaluated in the future.
     private final DoubleAdder count = new DoubleAdder();
     private final DoubleAdder sum = new DoubleAdder();
-    private final CKMSQuantiles quantiles;
+    private final TimeWindowQuantiles optionalQuantiles; // optional means can be null.
```
brian-brazil: Best to always have it set

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -155,12 +186,12 @@ private Value(double count, double sum, CKMSQuantiles quantiles) {
...
 
     static TimeProvider timeProvider = new TimeProvider();
 
-    private Child(CKMSQuantiles.Quantile[] q) {
-      quantiles = new CKMSQuantiles(q);
+    private Child(Quantile[] q, long maxAgeMillis, int nBuckets) {
+      optionalQuantiles = q.length > 0 ? new TimeWindowQuantiles(q, maxAgeMillis, nBuckets) : null;
```
brian-brazil: This would be clearer as an if

simpleclient/src/main/java/io/prometheus/client/TimeWindowQuantiles.java
```diff
@@ -0,0 +1,53 @@
+package io.prometheus.client;
+
+import io.prometheus.client.CKMSQuantiles.Quantile;
+
+/**
+ * Wrapper around CKMSQuantiles.
+ *
+ * Maintains a ring buffer of CKMSQuantiles to provide quantiles over a sliding windows of time.
+ */
+public class TimeWindowQuantiles {
```
brian-brazil: This shouldn't be public.

simpleclient/src/test/java/io/prometheus/client/SummaryTest.java
```diff
@@ -44,6 +63,43 @@ public void testObserve() {
...
+  public void testMaxAge() throws InterruptedException {
+    Summary summary = Summary.build()
+            .quantile(0.99, 0.001)
+            .maxAgeMillis(200) // After 200ms, all observations will be discarded.
+            .nBuckets(2)       // We got 2 buckets, so we discard one bucket each 100ms.
+            .name("short_attention_span").help("help").register(registry);
+    summary.observe(8.0);
+    double val = registry.getSampleValue("short_attention_span", new String[]{"quantile"}, new String[]{Collector.doubleToGoString(0.99)}).doubleValue();
+    assertEquals(8.0, val, 0.0); // From bucket 1.
+    Thread.sleep(100);
```
brian-brazil: That's going to be tight, give yourself at least 50ms slack

simpleclient/src/test/java/io/prometheus/client/SummaryTest.java
```diff
@@ -1,24 +1,37 @@
...
   @Before
   public void setUp() {
     registry = new CollectorRegistry();
     noLabels = Summary.build().name("nolabels").help("help").register(registry);
     labels = Summary.build().name("labels").help("help").labelNames("l").register(registry);
+    noLabelsAndQuantiles = Summary.build()
+            .quantile(0.5, 0.05)
+            .quantile(0.9, 0.01)
+            .quantile(0.99, 0.001)
+            .name("noLabelsAndQuantiles").help("help").register(registry);
```
brian-brazil: underscores are the convention for metric names

simpleclient/src/main/java/io/prometheus/client/Summary.java
fstab: I don't really understand what `Value` is used for outside of the `Summary` class. Should it be immutable? The `final double` values are immutable, but in order to make quantiles immutable, I would need to implement a wrapper that exposes only the `get(double quantile)` method but not the `insert(double value)` method. Is `Value` supposed to be used outside of the `Summary` class for read-only access to the metrics?

brian-brazil: It's for users to pull out information. It should be immutable.

A array/map snapshot would be good.

fstab:
Thanks very much for your comments again. I committed the following:

Apart from the straightforward changes, I switched from `final Quantile[] sortedQuantiles` to an unmodifiable list, so it becomes clear that `class Summary` is immutable. Sorting the quantiles was moved to the point where I create the snapshot in `Value`, and I used `SortedMap` as a type, so no need to have the term *sorted* in the variable name.

I find the term `nAgeBuckets` more intuitive for the number of buckets than just `ageBuckets`, but if you want me to remove the `n` to be consistent with client_golang that's fine.

NOTICE
```diff
@@ -6,3 +6,6 @@ Boxever Ltd. (http://www.boxever.com/).
 
 This product includes software developed at
 SoundCloud Ltd. (http://soundcloud.com/).
+
+This product includes software developed at
+Netflix Inc. (https://www.netflix.com).
```
brian-brazil: I'd specify which repo it was copied from, as the rest were developed in this repo.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -186,12 +192,18 @@ private Value(double count, double sum, TimeWindowQuantiles optionalQuantiles) {
...
     static TimeProvider timeProvider = new TimeProvider();
 
-    private Child(Quantile[] q, long maxAgeMillis, int nBuckets) {
-      optionalQuantiles = q.length > 0 ? new TimeWindowQuantiles(q, maxAgeMillis, nBuckets) : null;
+    private Child(List<Quantile> quantiles, long maxAgeSeconds, int nAgeBuckets) {
+      this.quantiles = quantiles;
+      if (quantiles.size() > 0) {
+        quantileValues = new TimeWindowQuantiles(this.quantiles.toArray(new Quantile[]{}), maxAgeSeconds, nAgeBuckets);
+      } else {
+        this.quantileValues = null;
```
brian-brazil: You're not using `this` two lines up.

brian-brazil:
> I find the term nAgeBuckets more intuitive for the number of buckets than just ageBuckets, but if you want me to remove the n to be consistent with client_golang that's fine.

I'd prefer to be consistent with client_golang.

fstab:
I renamed `nAgeBuckets` to `ageBuckets`, removed the unnecessary `this` in Child constructor, mentioned Ocelli repository in NOTICE.

brian-brazil:
Could you squash your commits please?

simpleclient/src/test/java/io/prometheus/client/SummaryTest.java
```diff
@@ -92,17 +148,16 @@ public void testLabels() {
 
   @Test
   public void testCollect() {
-    labels.labels("a").observe(2);
-    List<Collector.MetricFamilySamples> mfs = labels.collect();
```
brian-brazil: This test should remain here

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/TextFormatTest.java
```diff
@@ -56,13 +56,19 @@ public void testCounterOutput() throws IOException {
...
   public void testSummaryOutput() throws IOException {
-    Summary noLabels = (Summary) Summary.build().name("nolabels").help("help").register(registry);
-    noLabels.observe(2);
+    Summary labelsAndQuantiles = Summary.build()
+            .quantile(0.5, 0.05).quantile(0.9, 0.01).quantile(0.99, 0.001)
+            .labelNames("l").name("labelsAndQuantiles").help("help").register(registry);
+    labelsAndQuantiles.labels("a").observe(2);
+    writer = new StringWriter();
     TextFormat.write004(writer, registry.metricFamilySamples());
-    assertEquals("# HELP nolabels help\n"
```
brian-brazil: And this one too.

fstab:
Squashed and restored previous tests.

brian-brazil:
Thanks!

PR #137, opened 2016-08-18
Fix documentation for push and pushAdd methods.

mindprince:
According to https://github.com/prometheus/pushgateway/blob/0.3.0/README.md#post-method
> POST works exactly like the PUT method but only metrics with the same name as
> the newly pushed metrics are replaced (among those with the same grouping
> key).

push uses POST and pushAdd uses PUT.
push replaces only metrics with the same name under a job and grouping key.
pushAdd replaces all metrics under a job and grouping key.

The method names are unfortunate.

### DIFF
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -66,7 +66,7 @@ public class PushGateway {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for finding this. It's actually the implementation that's wrong, could you switch it around?

mindprince:
I thought so too. But I didn't because it would change the behavior for existing users. But if you are okay with doing that, I can update the PR.

mindprince:
Created https://github.com/prometheus/client_java/pull/138 in case we want to do that.

brian-brazil:
Closing in lieu of #138

PR #138, opened 2016-08-18
push and pushAdd logic reversed. #137, #138.

mindprince:
According to https://github.com/prometheus/pushgateway/blob/0.3.0/README.md#post-method

> POST works exactly like the PUT method but only metrics with the same name
as the newly pushed metrics are replaced (among those with the same grouping
key).

### DIFF
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -66,12 +66,12 @@ public class PushGateway {
   }
 
   /**
-   * Pushes all metrics in a registry, replacing all those with the same job as the grouping key.
+   * Pushes all metrics in a registry, replacing all those with the same job and no grouping key.
    * <p>
-   * This uses the POST HTTP method.
+   * This uses the PUT HTTP method.
   */
   public void push(CollectorRegistry registry, String job) throws IOException {
-    doRequest(registry, job, null, "POST");
+    doRequest(registry, job, null, "PUT");
   }
 
   /**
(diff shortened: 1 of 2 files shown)


simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -66,20 +66,20 @@ public PushGateway(String address) {
   }
 
   /**
-   * Pushes all metrics in a registry, replacing all those with the same job as the grouping key.
+   * Pushes all metrics in a registry, replacing all those with the same job and no grouping key.
```
brian-brazil: The comment was correct, the job is part of the grouping key.

mindprince: There are three other methods down below with similar signature that say "same job and no grouping key". I think this is to differentiate with the methods which take groupingKey map as argument.

brian-brazil: That's just offering them as optional parameters. The grouping key comprises the job label plus whatever other optional labels you add.

mindprince: I know that.

There are eight methods:
```
1. push(CollectorRegistry registry, String job)
2. push(Collector collector, String job)

3. push(CollectorRegistry registry, String job, Map<String, String> groupingKey)
4. push(Collector collector, String job, Map<String, String> groupingKey)

5. pushAdd(CollectorRegistry registry, String job)
6. pushAdd(Collector collector, String job)

7. pushAdd(CollectorRegistry registry, String job, Map<String, String> groupingKey)
8. pushAdd(Collector collector, String job, Map<String, String> groupingKey)
```

Methods 1,2,5,6 should have similar documentation. I made the documentation for method 1 similar to what was there for methods 2,5,6.

brian-brazil: That makes sense.

brian-brazil:
Thanks!

Issue #140, opened 2016-09-05
testHistogram in DropWizard client is Locale insensitive.


ovelindstrom:
If the JVM that runs the test is not having a default Locale that uses dots as delimiter in decimal values, such as the Swedish locale, the test will fail.

ovelindstrom:
Created PR with simple solution.

https://github.com/prometheus/client_java/pull/141

PR #141, opened 2016-09-05
Solution for issue 140: Lock down locale.

ovelindstrom:
Lock down the locale to that test will succeed regardless of JVM default Locale.

https://github.com/prometheus/client_java/issues/140

### DIFF
--- a/simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
+++ b/simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
@@ -9,9 +9,7 @@ import org.junit.Test;
 import java.io.IOException;
 import java.util.Arrays;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 public class DropwizardExportsTest {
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for the PR, there are several other calls to this function in the codebase. Are any of the rest of them affected?

ovelindstrom:
Could be. Let me take a look.

ovelindstrom:
As far as I can see, no other calls to the getSampleValue in this codebase is affected. Mostly, you use string representation of the quantiles like "0.99" where you query the registy. In this case, the doubles are used to make the arithmetic easier and make the test case shorter.

Lurking in the code of the History and Sample classes, they never actually mention a label format when it comes to different locales. Maybe it would be a good idea to define the used quantile lables.

brian-brazil:
Looking at the other code, this should probably just be `b.toString`

ovelindstrom:
You are right @brian-brazil. simplifying it to use a Double instead of the primitive and using String.format will make it easier. Committing the change.

brian-brazil:
Could you remove the extra import and squash your commit?

ovelindstrom:
Done.

simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
```diff
@@ -9,9 +9,7 @@
 import java.io.IOException;
 import java.util.Arrays;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
brian-brazil: Please leave imports explicit

brian-brazil:
Thanks

PR #142, opened 2016-09-08
export of Dropwizard metrics should use original name in help message

arnostv:
@brian-brazil : small cleanup in Dropwizard export help message

Dropwizard metric name can contain characters which are not allowed in Prometheus metrics name. For example `my.application.namedTimer1`. This name is then sanitized to `my_application_namedTimer1`.

The change is that help message should provide original metric name:

 `Generated from Dropwizard metric import (metric=my.application.namedTimer1, type=com.codahale.metrics.Timer)`

instead of

`Generated from dropwizard metric import (metric=my_application_namedTimer1, type=com.codahale.metrics.Timer)`

 Metric name in help message was not  correct.


brian-brazil:
Thanks!

PR #143, opened 2016-09-20
Changed visibility of Summary.Child.Value back to public.

pstibrany:
This gives clients access to current values of summary, if they need it.
Also added simple tests to avoid breaking this API in Summary and Histogram classes in the future.


brian-brazil:
Thanks!

Issue #139, opened 2016-08-22
Expose /prometheus as a Spring Boot Mgmt Endpoint


eliezio:
I missed a simpleclient submodule that creates a Spring Boot Management Endpoint to expose the prometheus metrics. 
The existing simpleclient_servlet module requires additional glue-code to accomplish this task.

eliezio:
Can I contribute with a new module based on the implementation drawn at
http://blog.monkey.codes/actuator-and-prometheus/?

maust:
+1 (with an Spring-style annotation like @EnablePrometheusEndpoint would be nice)

ovelindstrom:
+1

maust:
Here is a starting point if somebody wants to use prometheus with spring boot: https://github.com/maust/prometheus-spring-boot

sdomula:
I tried out the implementation of @maust. It works well.

brian-brazil:
If someone sends a PR I'm willing to merge.

eliezio:
Feature released on v0.0.18

PR #146, opened 2016-10-13
added /prometheus as a Spring Boot Endpoint

eliezio:
Here goes the code that implements the extension discussed on #139. Credits also go to Marco Aust that started the implementation.

### DIFF
--- a/simpleclient_spring_boot/pom.xml
+++ b/simpleclient_spring_boot/pom.xml
@@ -31,10 +31,24 @@
             <name>Tokuhiro Matsuno</name>
             <email>tokuhirom@gmail.com</email>
         </developer>
+        <developer>
+            <name>Marco Aust</name>
+            <email>github@marcoaust.de</email>
+            <organization>private</organization>
+            <organizationUrl>https://github.com/maust</organizationUrl>
+        </developer>
+        <developer>
+            <id>eliezio</id>
+            <name>Eliezio Oliveira</name>
+            <email>eliezio.oliveira@gmail.com</email>
+        </developer>
     </developers>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+        <spring.framework.version>4.3.3.RELEASE</spring.framework.version>
+        <spring.boot.version>1.4.1.RELEASE</spring.boot.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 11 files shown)


simpleclient_spring_boot/pom.xml
```diff
@@ -62,15 +87,21 @@
...
+            <groupId>org.cthul</groupId>
+            <artifactId>cthul-matchers</artifactId>
+            <version>1.1.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
-            <version>1.3.3.RELEASE</version>
+            <version>${spring.boot.version}</version>
```
brian-brazil: Stick with literal versions, some build tools don't support variables.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/EnablePrometheusEndpoint.java
```diff
@@ -0,0 +1,13 @@
+package io.prometheus.client.spring.boot;
+
+import org.springframework.context.annotation.Import;
+
+import java.lang.annotation.*;
```
brian-brazil: Be explicit in what you're importing

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+  }
+
+  @Configuration
+  @ConditionalOnClass(DefaultExports.class)
+  @ConditionalOnProperty(prefix = "prometheus.hotspot", name = "enabled", matchIfMissing = true)
+  static class HotspotMetricsExporter {
+    public HotspotMetricsExporter() {
+      DefaultExports.initialize();
```
brian-brazil: Similarly, this isn't the responsibility of exposition. The module should be exposing the default registry, and that's it.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+class PrometheusConfiguration {
+
+  @Configuration
+  static class PublicMetricsCollector {
+    @Autowired(required = false)
+    List<PublicMetrics> publicMetrics = Collections.emptyList();
+
+    @Bean
+    public SpringBootMetricsCollector springBootMetricsCollector() {
+      SpringBootMetricsCollector springBootMetricsCollector = new SpringBootMetricsCollector(publicMetrics);
```
brian-brazil: This isn't the responsibility of exposition

PR #147, opened 2016-10-13
Expose /prometheus as a Spring Boot Mgmt Endpoint (2nd PR)

eliezio:
Here is the revised code after applying your recommendations. I only kept the code related to the endpoint exposition.


eliezio:
Two minor issues I only saw after sending the PR:
1) I forgot to rename the method PrometheusEndpointTest#testMixedMetricsExportedThroughPrometheusEndpoint
2) it would be better to first handle the migration to SB 1.4 on the 1st commit and only then implement the endpoint exposition

PR #148, opened 2016-10-13
Expose /prometheus as a Spring Boot Mgmt Endpoint (3rd Rev)

eliezio:
Here is the revised code after following your recommendations. I only kept the code related to the endpoint exposition.

### DIFF
--- a/simpleclient_spring_boot/pom.xml
+++ b/simpleclient_spring_boot/pom.xml
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-actuator</artifactId>
-            <version>1.3.3.RELEASE</version>
+            <version>1.4.1.RELEASE</version>
         </dependency>
 
         <!-- Test Dependencies Follow -->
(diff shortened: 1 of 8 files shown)


simpleclient_spring_boot/pom.xml
```diff
@@ -64,13 +64,13 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
-            <version>1.3.3.RELEASE</version>
+            <version>1.4.1.RELEASE</version>
```
brian-brazil: 1.4 has only been out a few months, it it reasonable to presume that almost everyone has upgraded already?

I like to keep support for a version that covers the most users, and depend on backwards compatibility for anyone using something newer.

simpleclient_spring_boot/src/test/java/io/prometheus/client/matchers/CustomMatchers.java
```diff
@@ -0,0 +1,43 @@
+package io.prometheus.client.matchers;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.core.IsCollectionContaining;
+
+/**
+ * @author <a href="http://stackoverflow.com/users/4483548/bretc">BretC</a>
```
brian-brazil: What license is this under?

brian-brazil:
Javadocs showing how to use all this would be good.

simpleclient_spring_boot/src/test/java/io/prometheus/client/matchers/CustomMatchers.java
eliezio: Since Feb-1-2016 all code posted on StackOverflow is licensed under MIT License. I'll append this notice.

simpleclient_spring_boot/pom.xml
eliezio: I'll revert to SB 1.3 and send a new PR if no major problem arises.

eliezio:
I'll document the @EnablePrometheusEndpoint showing how to use it.

PR #151, opened 2016-10-17
added /prometheus as a Spring Boot Endpoint

eliezio:
fixed doc issues

### DIFF
--- a/simpleclient_spring_boot/pom.xml
+++ b/simpleclient_spring_boot/pom.xml
@@ -31,6 +31,17 @@
             <name>Tokuhiro Matsuno</name>
             <email>tokuhirom@gmail.com</email>
         </developer>
+        <developer>
+            <name>Marco Aust</name>
+            <email>github@marcoaust.de</email>
+            <organization>private</organization>
+            <organizationUrl>https://github.com/maust</organizationUrl>
+        </developer>
+        <developer>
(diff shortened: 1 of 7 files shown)


brian-brazil:
Thanks!

fiunchinho:
When will this released? Thanks!

brian-brazil:
Next week sometime is the plan.

fiunchinho:
I was trying to test it locally just copy and pasting the files from this pull request, but I can't get it to work properly. After adding the `@EnablePrometheusEndpoint()` annotation to my Application, I can visit the `/prometheus` endpoint but with empty result.
The `/metrics` endpoint is returning the usual response.

prometheus

Do I need to do anything else?

eliezio:
The metrics you are seeing through the /metrics endpoint are those that were auto-configured by Spring Boot.
To see these metrics also through /prometheus endpoint, you need to instantiate a SpringBootMetricsCollector bean as stated on its documentation (see https://prometheus.io/client_java/io/prometheus/client/spring/boot/SpringBootMetricsCollector.html).
Maybe there should be a second annotation (i.e.: @ExportPublicMetricsToPrometheus) to simplify this task.

fiunchinho:
You are right. It works now, thank you very much

PR #152, opened 2016-10-24
Add annotation to easily export prometheus metrics through the prometheus

fiunchinho:
@eliezio @brian-brazil As [discussed here](https://github.com/prometheus/client_java/pull/151#issuecomment-255700815), it may be useful to have an annotation to easily export metrics to prometheus.
I'm open to suggestions on how to improve the code.

### DIFF
--- a/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/EnablePrometheusEndpoint.java
+++ b/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/EnablePrometheusEndpoint.java
@@ -38,7 +38,7 @@ import java.lang.annotation.Target;
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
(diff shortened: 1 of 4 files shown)


simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/ExportMetricsToPrometheus.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.springframework.context.annotation.Import;
+
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Import(PrometheusMetricsConfiguration.class)
+public @interface ExportMetricsToPrometheus {
```
brian-brazil: This name doesn't really capture what it does. `EnableSpringBootMetricsCollector` would be better I think.

PR #154, opened 2016-10-31
Add class loading metrics to default exports

cfrantsen:


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ClassLoadingExports.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public ClassLoadingExports(ClassLoadingMXBean clBean) {
+    this.clBean = clBean;
+  }
+
+  void addClassLoadingMetrics(List<MetricFamilySamples> sampleFamilies) {
+    sampleFamilies.add(
+            new MetricFamilySamples(
+                    "jvm_classloading_loaded_current",
```
brian-brazil: I'd drop the _current

I think jmv_classes_... might also read better here.

cfrantsen:
Updated with suggested changes

brian-brazil:
Thanks!

PR #155, opened 2016-11-08
Add https and basic auth capability for PushGateway

Eloo:


simpleclient_pushgateway/src/test/java/io/prometheus/client/exporter/ExamplePushGateway.java
```diff
@@ -12,7 +12,7 @@
    * Example of how to use the pushgateway, pass in the host:port of a pushgateway.
    */
   public static void main(String[] args) throws Exception {
-    PushGateway pg = new PushGateway(args[0]);
+    PushGateway pg = new PushGateway(args[0], args[1]);
```
brian-brazil: This will throw an exception if there's only one arg.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,22 +50,53 @@
...
    * <p>
-   * @param address  host:port or ip:port of the Pushgateway.
+   *
+   * @param scheme   url scheme of the Pushgateway.
+   * @param address  host or ip of the Pushgateway.
+   * @param port     port of the Pushgateway.
+   * @param username username for basic authorization.
+   * @param password password for basic authorization.
    */
-  public PushGateway(String address) {
```
brian-brazil: You're removing this function, which will break all existing users.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,22 +50,53 @@
...
+  public PushGateway(String scheme, String address, String port, String username, String password) {
```
brian-brazil: You're presuming that basic auth is the only way that a use would like to authenticate. Someone might want ssl, kerberos or something else. Can you do something more generic?

Eloo:
Add more generic way to authenticate by passing a java.net Authenticator implementation.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -4,18 +4,13 @@
...
-import java.net.URL;
-import java.net.URLEncoder;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-
 import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
+import java.net.*;
```
brian-brazil: Please leave imports explicit.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,22 +49,69 @@
...
+  }
+
+  /**
+   * Construct a Pushgateway, with the given url and port.
+   * <p>
+   *
+   * @param url host or ip of the Pushgateway.
+   * @param port    port of the Pushgateway.
+   */
+  public PushGateway(String url, String port) {
```
brian-brazil: A port is part of a URL

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,22 +49,69 @@
...
+  /**
+   * Construct a Pushgateway, with the given address.
+   * <p>
+   *
+   * @param scheme   url scheme of the Pushgateway.
+   * @param url  host or ip of the Pushgateway.
+   * @param port     port of the Pushgateway.
+   * @param authenticator authenticator for the Pushgateway.
+   */
+  public PushGateway(String scheme, String url, String port, Authenticator authenticator) {
```
brian-brazil: Scheme is also part of a URL

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,22 +49,69 @@
...
+    this.scheme = scheme;
+    this.url = url;
+    this.port = port;
+    Authenticator.setDefault(authenticator);
```
brian-brazil: Isn't this setting a global? This should only apply to this object.

Eloo:
Don't know why the test is failing, has nothing to do with my changes :/

brian-brazil:
There's one flaky http server test. I mightn't get to reviewing this for a week as I'm travelling.

Eloo:
Any updates here?

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -50,28 +61,112 @@
...
+ * <a href="http://www.programcreek.com/java-api-examples/index.php?source_dir=jenkins-client-master/src/main/java/com/offbytwo/jenkins/client/JenkinsHttpClient.java">JenkingsHttpClient</a>
+ * <a href="http://www.programcreek.com/java-api-examples/index.php?source_dir=jenkins-client-master/src/main/java/com/offbytwo/jenkins/client/PreemptiveAuth.java">PreemptiveAuth</a>
+ * <p>
  * See <a href="https://github.com/prometheus/pushgateway">https://github.com/prometheus/pushgateway</a>
  */
 public class PushGateway {
 
-  private final String address;
-  private static final int SECONDS_PER_MILLISECOND = 1000;
+  private static final int TIMEOUT = 10000;
```
brian-brazil: This needs a unit

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -50,28 +61,112 @@
  * }
  * </pre>
  * <p>
+ * Example usage with Basic authentication and SSL:
```
brian-brazil: I'd cut this example down to just the auth bits

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -50,28 +61,112 @@
...
    * Construct a Pushgateway, with the given address.
    * <p>
-   * @param address  host:port or ip:port of the Pushgateway.
+   *
+   * @param scheme     address scheme of the PushGateway.
+   * @param address    host or ip of the PushGateway.
+   * @param port       port of the PushGateway.
+   * @param httpClient httpClient for the connection to the PushGateway
+   */
+  public PushGateway(String scheme, String address, String port, HttpClient httpClient) {
```
brian-brazil: I don't think we need this many consructors. Just one with a URL, and one with the url and client should be sufficient,

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -216,69 +316,79 @@ public void pushAdd(CollectorRegistry registry, String job, String instance) thr
...
         url += "/" + entry.getKey() + "/" + URLEncoder.encode(entry.getValue(), "UTF-8");
       }
     }
-    HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
-    connection.setRequestProperty("Content-Type", TextFormat.CONTENT_TYPE_004);
-    if (!method.equals("DELETE")) {
-      connection.setDoOutput(true);
-    }
-    connection.setRequestMethod(method);
+    customizeRequestConfig(request);
```
brian-brazil: This is only used once, inline.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -135,9 +232,11 @@ public void pushAdd(Collector collector, String job) throws IOException {
    * Pushes all metrics in a registry, replacing only previously pushed metrics of the same name, job and grouping key.
    * <p>
    * This uses the POST HTTP method.
-  */
+   */
   public void pushAdd(CollectorRegistry registry, String job, Map<String, String> groupingKey) throws IOException {
-    doRequest(registry, job, groupingKey, "POST");
+    HttpPost request = new HttpPost();
+    request.setEntity(createOutputEntity(registry));
```
brian-brazil: I think this code and function would be better inlined  in doRequest

brian-brazil: I mean move this code out of this function.

Eloo: Okay but i think its not a good solution to put this in doRequest, because it depends on the request if we ne a HttpPost, HttpPut or HttpDelete.
May we could move the setEntity part to doRequest, pass the registry and check if we got a registry object in doRequest.
What do you think?

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
brian-brazil: Now the example refers to variables that aren't defined. Can you clean it up?

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -50,28 +61,89 @@
...
+    this("http", authority, HttpClients.createDefault());
+  }
+
+  /**
+   * Construct a Pushgateway, with the given authority, scheme and a {@link HttpClient}.
    * <p>
-   * @param address  host:port or ip:port of the Pushgateway.
+   *
+   * @param scheme     address scheme of the PushGateway.
+   * @param authority  host:port or ip:port of the Pushgateway.
```
brian-brazil: I've never seen this called an authority.  Take in a URL as one argument.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -50,28 +61,89 @@
...
 
-  private final String address;
-  private static final int SECONDS_PER_MILLISECOND = 1000;
+  private static final int TIMEOUT_IN_MILLISECONDS = 10000;
+
+  private final HttpClient httpClient;
+
+  private final String urlBase;
+
+  private BasicHttpContext localContext;
```
brian-brazil: This isn't threadsafe, so you either need a lock or to create it on the fly.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
brian-brazil: That's roughly what I'm thinking.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
Eloo: Please check out RFC2396 (https://www.ietf.org/rfc/rfc2396.txt) part 3. URI Syntactic Components

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
Eloo: Which field do you mean? localContext? If yes, why do you think we need threadsafety for this field?

Eloo: Thats not 100% correct.
As mentioned in https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/protocol/BasicHttpContext.html the BasicHttpContext can be not threadsafe if the parent context isn't threadsafe. But we don't use a parent context so this basic context is threadsafe.
Do you agree?

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -50,28 +61,92 @@
...
+ * <p>
  * See <a href="https://github.com/prometheus/pushgateway">https://github.com/prometheus/pushgateway</a>
  */
 public class PushGateway {
 
-  private final String address;
-  private static final int SECONDS_PER_MILLISECOND = 1000;
+  private static final int TIMEOUT_IN_MILLISECONDS = 10000;
+
+  private static final String INSTANCE_GROUPING_KEY = "instance";
```
brian-brazil: Leave this inlined.

Eloo: This is not a good idea because the instance_grouping_key is used multiple times and have this usages "inlined" wouldn't be nice code and fragile for change errors.

Eloo: But this is really bad coding style to use multiple string literals.

brian-brazil: I optimise my code for being read by a human.

marcusvetter: Excuse me, but this is ridiculous. You should absolutely write code for being read by a human and, in particular write CLEAN CODE. Inlining string constants is one of the worst things you could do and violates all good practices in software engineering.

marcusvetter: Ok you're right. CLEAN CODE is my personal preference ;-)

dhpizza: Clean code is personal preference of Uncle Bob, too.

Eloo:
Merge?

brian-brazil:
There are still outstanding comments.

Eloo:
Hope you didn't mean the instance string literal thing

brian-brazil:
I mean that and other comments.

Eloo:
Could you specify which other comments?

brian-brazil:
There's still a separate argument for scheme.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -110,20 +110,19 @@
...
   /**
    * Construct a Pushgateway, with the given authority, scheme and a {@link HttpClient}.
    * <p>
    *
-   * @param scheme     address scheme of the PushGateway.
-   * @param authority  host:port or ip:port of the Pushgateway.
+   * @param urlBase    combined by &lt;scheme&gt;://&lt;host&gt;:&lt;port&gt;
    * @param httpClient httpClient for the connection to the PushGateway
    */
-  public PushGateway(String scheme, String authority, HttpClient httpClient) {
```
brian-brazil: baseUrl is more common

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -110,20 +110,19 @@
    * @param authority host:port or ip:port of the Pushgateway.
    */
   public PushGateway(String authority) {
```
brian-brazil: This should be "host" or similar, as most users will never have heard of "authority"

Eloo:
So this is my last try. Changed the constructor to use a single string argument with scheme.
The instance thing will not be change due to the fact that i'm not willing to consciously contribute code smells!

brian-brazil:
> The instance thing will not be change due to the fact that i'm not willing to consciously contribute code smells!

This change will not be merged until this unnecessary and unrelated change to the code is undone.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -216,69 +295,87 @@ public void pushAdd(CollectorRegistry registry, String job, String instance) thr
...
+  }
+
+  /**
+   * Creates the output entity for the POST or PUT requests based on the passed registry
+   *
+   * @param registry the metric collector registry
+   * @return the {@link StringEntity} for the request
+   * @throws IOException
+   */
+  private StringEntity createOutputEntity(CollectorRegistry registry) throws IOException {
```
brian-brazil: Please inline this function

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -216,69 +295,87 @@ public void pushAdd(CollectorRegistry registry, String job, String instance) thr
...
-      }
-    }
-    HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
-    connection.setRequestProperty("Content-Type", TextFormat.CONTENT_TYPE_004);
-    if (!method.equals("DELETE")) {
-      connection.setDoOutput(true);
-    }
-    connection.setRequestMethod(method);
+  void doRequest(String job, Map<String, String> groupingKeys, HttpRequestBase request) throws IOException {
+    String url = buildUrl(job, groupingKeys);
```
brian-brazil: and this one

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -216,69 +295,87 @@ public void pushAdd(CollectorRegistry registry, String job, String instance) thr
...
 
-    connection.setConnectTimeout(10 * SECONDS_PER_MILLISECOND);
-    connection.setReadTimeout(10 * SECONDS_PER_MILLISECOND);
-    connection.connect();
+    customizeRequestConfig(request);
```
brian-brazil: and this one

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -216,69 +295,87 @@ public void pushAdd(CollectorRegistry registry, String job, String instance) thr
...
+   * @param request the http request
+   */
+  private void customizeRequestConfig(HttpRequestBase request) {
+    RequestConfig config = request.getConfig();
+    RequestConfig.Builder builder;
+    if (config == null) {
+      builder = RequestConfig.custom();
+    } else {
+      builder = RequestConfig.copy(config);
+
```
brian-brazil: extra blank line here

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
Eloo: But in fact this isn't the host!

Eloo:
This is ridiculous!

PR #157, opened 2016-11-14
allow to register default JVM collectors with the given registry

arnostv:
@brian-brazil 

Prometheus Java API in general allows to do so, this is just to make DefaultExports to follow this idea.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/DefaultExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/DefaultExports.java
@@ -1,5 +1,7 @@
 package io.prometheus.client.hotspot;
 
+import io.prometheus.client.CollectorRegistry;
+
 /**
  * Registers the default Hotspot collectors.
  * <p>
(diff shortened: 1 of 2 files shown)


brian-brazil:
Why do you want to do this?

arnostv:
In my application I use the Prometheus Client API with explicitly defined  CollectorRegistry and then passing where needed (e.g. https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Collector.java#L126 or https://github.com/prometheus/client_java/blob/master/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java#L30).

This change just allows to stick to this style with DefaultExports (or I could copy&paste stuff from DefaultExports but that's not the point of DefaultExports)

brian-brazil:
It's recommended to use the default registry in cases like that, so that for example if you pull in a 3rd party library that's instrumented you'll get the metrics automatically.

Custom registries are usually for ad-hoc short-lived use cases.

arnostv:
Oh, I see it is actually documented here  - https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L16

I don't think it would really be a major problem with 3rd party library. It would either support passing  CollectorRegistry as parameter, if not I could make new pull request do improve it. And if this pull request was merged, I could use it as a reference to show that it makes sense to do so ;)

But if you prefer to stick to CollectorRegistry.defaultRegistry, then please close this pull request, I'll find some way to work with this.

brian-brazil:
> It would either support passing CollectorRegistry as parameter, if not I could make new pull request do improve it.

Consider the case where the library was a transitive dependency. Getting every intermediate library to support this could be challenging, and the end result of all that effort would be the exact same as if you'd used the default registry in the first place.

PR #160, opened 2016-11-21
Make Timer classes Closeable

russgray:
If the timer classes are closeable they can be used with java8's
try-with-resources, which allows for slightly more concise code as you don't
need to write the finally block.

No tests here because a) the new method directly delegates to an existing method
with no new logic, and b) because the test project doesn't target java8 and so
cannot use try-with-resources anyway.

@brian-brazil @beorn7


brian-brazil:
Gauge also has a Timer.

Isn't this a Java7 feature?

russgray:
> Isn't this a Java7 feature?

brian-brazil:
Thanks!

Issue #158, opened 2016-11-14
Spring Boot improvements


pidster:
The Spring Boot support could be improved in the following ways:

- Use AutoConfiguration, so just adding the jar to the classpath is enough to launch the actuator classes
- Add a single `@EnablePrometheus` annotation, with parameters to enable/disable other configuration
- Use `@ConditionalXXX` to override the auto-configuration classes, where users would need to customise things
- Expose the common collector/exporters

I think there may be more opportunities, but if there's an interest in doing this I'll provide a PR for review.

arnostv:
If you are going to work on Spring Boot integration,  I'd also suggest improving documentation a bit. For example to me it was not even clear if Spring Boot support we have right now is trying to:

- expose Prometheus metric in Spring Boot way - i.e. `GET host/metrics` shows JSON and will include Prometheus counters, timers, ...
- expose Spring metrics on Prometheus - i.e. `GET host/prometheus` shows metrics in text format and it will include some additional metrics from Spring Boot
- it just allows Spring Boot application to show only Prometheus metrics using Prometheus text format

Maybe all of those options?

brian-brazil:
In terms of documentation for this repo generally, I'm considering doing similar to the python client and having examples in the README.md. While all our docs are currently in java doc, which classes are relevant is no longer really organically discoverable as the number of modules have grown.

pidster:
@arnostv We could certainly make it more specific.

`GET host/metrics` will hit the standard Spring Boot 'metrics actuator' endpoint, it doesn't know about Prometheus stuff.

pidster:
I will open a separate issue for improvements to docs after the above PR is reviewed.

arockiaraj1994:
@arnostv worked fine for me, bit struggle to find the solution. It would be great if place in documentation 

Thanks

PR #163, opened 2016-11-25
Springify some more things & use auto-configuration.

pidster:
Fixes https://github.com/prometheus/client_java/issues/158.

- Some minor improvements to the build.
- Tests are improved
- More Spring-like processing of Collector beans
- More flexible `@Enable` annotations
- Adds Auto-Configuration, which will enable things just by adding the Jar to the classpath.


simpleclient_spring_boot/pom.xml
```diff
@@ -42,32 +42,48 @@
...
+        <spring.core.version>4.3.4.RELEASE</spring.core.version>
+        <spring.boot.version>1.4.2.RELEASE</spring.boot.version>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>io.prometheus</groupId>
             <artifactId>simpleclient</artifactId>
-            <version>0.0.19-SNAPSHOT</version>
+            <version>${prometheus.client.version}</version>
```
brian-brazil: This needs to be left as-is. Not all tools dealing with poms support this.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * An {@literal @}Enable style annotation that provides CollectorRegistry related functions.
+ *
+ * <pre><code>
+ *{@literal @}Configuration
+ *{@literal @}EnablePrometheusCollectorRegistration(useDefault = true, register = true)
+ * public class MyConfiguration {
+ *
+ *  {@literal @}Bean
+ *   public Counter counter() {
```
brian-brazil: I don't think this is a usage we should be encouraging. A standard `.register()` on a static member will do what the user wants.

This however has the potential for bugs as it's not static.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/HotspotMetricCollectorsConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+import io.prometheus.client.hotspot.ThreadExports;
+
+/**
+ * A {@link Configuration} class that imports the Prometheus
+ * JVM management metrics Collectors from the <code>simpleclient_hotspot</code> package.
+ *
+ * @author pidster
+ */
+@Configuration
+public class HotspotMetricCollectorsConfiguration {
```
brian-brazil: Don't duplicate what's already handled in DefaultExports.

simpleclient_spring_boot/pom.xml
pidster: Really?! 
I guess that explains why it's duplicated...
May I ask which ones (for my own future reference)?

brian-brazil: I forget which one it was, the scala or clojure one I think.

pidster: Oh dear. I will push another commit to address this.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
pidster: My recommendation/advice is to let Spring deal with this, because relationships & dependencies in Spring are best managed by the dependency injection container. It is static calls & references that are more likely to lead to bugs here.

The 'natural' way to build things in Spring is to just create the bean, and any other things that are interested in it will find & use it.

The associated BeanPostProcessor will see every Collector bean and ensure it is registered once, and once only.

Which potential bugs are you concerned about?

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/HotspotMetricCollectorsConfiguration.java
pidster: Yes, I did see that class - but per the above comment regarding registration/relationships in Spring, I was letting the container handle it.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
brian-brazil: Duplicate metrics, incorrect uses of metrics, inappropriate use of custom registries.

The original java client had code like this, it was over complicated and unnecessary.

Direct instrumentation should use the java client in the usual fashion.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/HotspotMetricCollectorsConfiguration.java
brian-brazil: All you should need to do is ensure DefaultExports gets initialised at least once. I'm not going to maintain that list in two places.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
pidster: Classic. No sooner do I say it than the build fails.

On reviewing it, (and noting that I'm a little rusty), I think that neither the static `CollectorRegistry.defaultRegistry` field, nor the `DefaultExports.initialize()` method are entirely thread safe.

pidster:
OK, it's failing now because it can't see ClassLoadingExports. Which is weird because my build can see that and the project hasn't changed here, since the first build of this PR - which did see that class.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
brian-brazil: That sounds like a bug in Java.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/HotspotMetricCollectorsConfiguration.java
brian-brazil: You shouldn't have to care about this for tests, as the call would be in your main (and has to be, in case you're not using hotspot).

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
pidster: > Duplicate metrics, incorrect uses of metrics, inappropriate use of custom registries.
> The original java client had code like this, it was over complicated and unnecessary.
> Direct instrumentation should use the java client in the usual fashion.

Understood.

The way Spring wants to do things is often different to the way non-Spring apps work. It's the nature of "inversion of control".

It's been my experience that more problems emerge when one doesn't try to do things the Spring way, than the other.

Can I ask you to consider it, & perhaps some other people with Spring knowledge can comment?

brian-brazil:
You're likely hitting that there's no SNAPSHOT with that change. Try -20 in a few minutes, I'm in the middle of a release.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/HotspotMetricCollectorsConfiguration.java
pidster: Spring does not need a main for tests, e.g. for Boot, main is just used to call a utility that starts the Spring container - which then handles the rest of the class lifecycle. 

The tests in this module now don't have a main class, because a complete Spring & Servlet container can be started on a random-ish port and run inside the test class by Springs JUnit integration. See `io.prometheus.client.spring.boot.AutoConfigurationTests` for an example.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/EnablePrometheusCollectorRegistration.java
pidster: Given, a Spring application that processes orders and a Counter that counts them.

If the whole application is started for a test, the first time it uses the static field the Counter is registered in the static CollectorRegistry (using `counter.register()`).

If the JVM is not recycled (it is usually not recycled during builds), the next test that starts the whole application will also try to register the same Counter in the CollectorRegistry - which will fail, because the other Counter is still there - incidentally causing a small memory leak.

brian-brazil: That's not a usual way of doing testing, that's going to break any singletons implemented as static final. Are you sure those are the semantics of how Spring does tests, as that sounds very fragile?

> the first time it uses the static field the Counter is registered in the static CollectorRegistry

This happens the first time the class is loaded, not when the field is used.

pidster: Apologies for being imprecise. 

> That's not a usual way of doing testing, that's going to break any singletons implemented as static final.

It is the usual way of testing Spring applications.
NB Spring can mark any object/bean as a singleton and has a registry, so it doesn't need to use a static final field reference to ensure a given object is accessible to other classes.

Static field singletons are somewhat notorious for causing memory leaks in containers with multiple classloaders, especially when they're used as a registry. 

If classes loaded in different classloaders are added to the static singleton, their classloader can't be garbage collected if it is otherwise ready to be dereferenced - unless the registry clears their reference, which it may not be able to do.

> Are you sure those are the semantics of how Spring does tests, as that sounds very fragile?

Yes, I'm sure. Perhaps it's just a side-effect of a conflict of styles.

> That sounds like a bug in Java.

Well... funny you should mention that, but even the double-checked-locking idiom was broken until Java 5. In this case, I don't think it's a bug in Java. Here's a good write up about singletons & safety: https://shipilev.net/blog/2014/safe-public-construction/#_singletons_and_singleton_factories

brian-brazil: > Static field singletons are somewhat notorious for causing memory leaks in containers with multiple classloaders, especially when they're used as a registry.

In this context that's not a memory leak, that's correct behaviour. The Collector should continue to exist and be registered until the process shutdowns.

> Here's a good write up about singletons & safety

I don't see any indication there that the existing client_java code is incorrect.

arnostv:
@pidster Currently I have Prometheus metrics in Spring Boot application and it is exposed in very simple way:
```
  @Bean
  public ServletRegistrationBean prometheusServletRegistrationBean() {
    DefaultExports.initialize();
    MetricsServlet metricsServlet = new MetricsServlet(CollectorRegistry.defaultRegistry);
    return new ServletRegistrationBean(metricsServlet, "/prometheus-metrics");
  }
```
At the beginning the application was using some Dropwizard Metrics and it is pretty easy to expose those
```
  @Bean
  public Collector dropwizardPrometheusExports(MetricRegistry registry) {
    DropwizardExports dropwizardExports = new DropwizardExports(registry);
    dropwizardExports.register();
    return dropwizardExports;
  }
```

Would it be possible if this "Prometheus - Spring Boot integration" supported this way of integration as well? Do you think this works using current design? Basically I'd like to register `SpringPublicMetricsCollector` to expose Spring Boot Metrics, but I don't want to have any other Spring Boot automatics activated.

simpleclient_spring_boot/src/test/java/io/prometheus/client/spring/boot/AutoConfigurationTests.java
```diff
@@ -0,0 +1,74 @@
...
+    ApplicationContext context;
+
+    @Autowired
+    SecurityProperties securityProperties;
+
+    @LocalServerPort
+    int port;
+
+    @Value("${endpoints.prometheus.path}")
+    String prometheusPath;
```
arnostv: This testing is failing for me if I run it directly in IntelliJ IDEA (2016.2.5), but it works for me in Maven. @pidster , @brian-brazil - maybe one of you is using IntelliJ, could you check if this works for you? It my be just my local problem, but it may also hit some other users...

I'm getting exception `Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'endpoints.prometheus.path' in string value "${endpoints.prometheus.path}"`

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusProperties.java
```diff
@@ -0,0 +1,50 @@
+package io.prometheus.client.spring.boot;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author pidster
+ */
+@ConfigurationProperties(prefix = "endpoints.prometheus")
+public class PrometheusProperties {
```
arnostv: I'm wondering how  this class is used - I can see it only in `@EnableConfigurationProperties({PrometheusProperties.class})` in `PrometheusAutoConfiguration`, but I don't see class properties used anywhere. For example I'm trying to find how the `path` is used (I assume this is for path in URL), but it is not clear (to me :) how this is applied.

simpleclient_spring_boot/src/test/java/io/prometheus/client/spring/EnableCollectorRegistrationAnnotationTests.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Test
+    public void testCollectorRegistryAware() {
+        TestCollectorRegistryAware aware = context.getBean(TestCollectorRegistryAware.class);
+        assertNotNull(aware.collectorRegistry);
+
+        CollectorRegistry registry = context.getBean(CollectorRegistry.class);
+        assertNotNull(registry);
+        assertNotEquals(CollectorRegistry.defaultRegistry, registry);
```
arnostv: What is the `CollectorRegistryAware` needed for actually?  As we show here the `CollectorRegistry` can be received either from `CollectorRegistryAware` or from `context`. In case of `context`, I assume it should be possible to use simple `@Autowired CollectorRegistry collectorRegistry`. 
I'm wondering if the `CollectorRegistryAware`  is actually required in some cases or if this is just an alternative to `@Autowired`.

simpleclient_spring_boot/src/test/java/io/prometheus/client/spring/boot/AutoConfigurationTests.java
pidster: @brian-brazil @arnostv I think I'm going to close this PR and break it up into smaller units. There's too much going on and it's too complex to debate the details of each thing in a single PR thread.

pidster:
> Basically I'd like to register SpringPublicMetricsCollector to expose Spring Boot Metrics, but I don't want to have any other Spring Boot automatics activated.

All of the PublicMetrics come from auto configuration.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/HotspotMetricCollectorsConfiguration.java
unguiculus: I'm currently integrating Prometheus in our Spring Boot app and am experiencing pain the static stuff causes. As @pidster pointed out, static state is bad. One of the reasons why you use DI frameworks such as Spring is to get rid of static state. 

You cannot use Spring Devtools with the current static implementation. Spring tries to reload the context but the static stuff stays around. As a result, the context does not come back up because already registered metrics survive the reload. I have to call `clear()` to get around this. Clearing the registry then causes the next problem. The default exports cannot be re-registered because they track initialization in a static field which is still `true` because it also survives the reload. As a result, I have this in my Spring configuration which  is ugly:

```java
@Autowired
void initializeDefaultsExports() {
    // The default registry, which is also used by the Spring Boot integration, 
    // is static. It survives a reload with Spring Devtools. As a result the 
    // Spring context does not come back up again, because the registry  
    // throws an exception because we try to register metrics which are 
    // already there. Thus, we just clear the registry to be on the safe side.
    CollectorRegistry.defaultRegistry.clear();

    // We cannot use this: DefaultExports.initialize()
    // It tracks initialization which is still true after a reload with Spring  
    // Devtools causing the exports not to be re-registered.
    new StandardExports().register();
    new MemoryPoolsExports().register();
    new GarbageCollectorExports().register();
    new ThreadExports().register();
    new ClassLoadingExports().register();
}
```

Thanks for your effort, @pidster. I'd very much like to see a solution that let's me opt out from static state.

@brian-brazil Could you please consider looking at this again? Maybe we can go for a simpler solution with less Spring magic, if that suits you better. All I ask for is to be able to provide a non-static CollectorRegistry. I'd be happy to help.

PR #164, opened 2016-11-30
Collertor registery listener

jfbreault:


brian-brazil:
Why do you want to ad this? The abstraction is such that you shouldn't care.

jfbreault:
I want to be able to add some sort of jmx reporter to have live metrics from the server. Something like the dropwizard jmx reporter. It need to know when the metric added and removed to unregister or register it correctly.  It's the first part of it and I think it can be usefull in other case

jfbreault:
That's not how jmx object registration work. I would like to use it like that but cannot do.

brian-brazil:
The only correct way is to use `metricFamilySamples()`, as collectors are not required to pre-declare what metrics they return. I'd suggest looking for a way having a single bean that returns everything.

jfbreault:
A single jmx object?
Have you ever work with jmx?, it's unusable this way.
A metric Object should be something like that
`
Prometheup:Name=requests_latency_seconds,Type=HISTOGRAM,label1=name1,...
  help=Request latency in seconds.
  bucket1=value1
  bucket2=value2
  ...
  count=value
  sum=value
`
The attribute should be generated live(aka Dynamic MBean) for histogram,summary and untyped as they have undefined number of quantile/bucket. Gauge and counter can work as standard Mbean

brian-brazil:
> But exporter that change the collect return over time(different MetricFamilySamples number), I don't see how jmx can ever work with it. Like dropwizard exporter can do weird thing

This is what you need to deal with, and if dropwizard has a solution I suggest following their lead.

jfbreault:
Dropwizard use the listener to register and unregister mbean

brian-brazil:
That approach won't work with Prometheus, as we're more generic,

PR #167, opened 2016-12-01
dropwizard histogram: export min, max, mean, median, stdDev

atorstling:
I noticed that some stats from Dropwizard histograms weren't exported, and simply added them. I was primarily interested in the historical extremes (min, max), but added the others while at it.


brian-brazil:
These were explicitly not included as they're not very useful statistically.

atorstling:
Ok. So how do you propose deriving max? I'm using a histogram to measure a queue size, so the max is very important for understanding how to size the queue. 

While I can understand the viewpoint that a max measurement isn't particularly useful in general for understanding the characteristics of a statistical distribution, it's very useful for understanding worst case and identifying outliers. This especially when your assumed statistical model doesn't hold very well. It's very good for identifying spurious problems like GC, dropped packages etc.

brian-brazil:
> I'm using a histogram to measure a queue size, so the max is very important for understanding how to size the queue.

Queue size would be using a gauge, not a histogram. There's no general way to do this, as we don't know what time period is involved.

atorstling:
Ok, but a Gauge doesn't give me any statistics, whereas a Histogram does. This is why I switched. It's also interesting for us to see how many items were queued totally over time, how much it fluctuates etc. So a Dropwizard Histogram is actually really useful in this case. According to the Dropwizard docs, a Histogram is just providing `...the statistical distribution of values in a stream of data`. If you regard the sampling of the queue size as a stream of data, this should be a totally valid use case as far as I can see. Could you not incorporate the min and max somehow?

brian-brazil:
There's no way we can do this statistically, this would break as soon as a process restarts.

PR #168, opened 2016-12-01
Add AOP-based spring method timing metrics, servlet filter for path-based metrics.

andrewstuart:
I'm not entirely certain that the spring_boot module is the best place for the AOP annotation, since it should work fine in any other spring application, where users may not want to import all the baggage that spring boot will drag along with it, but I thought I'd get this PR submitted at least, and get input on where that belongs.

/cc: @beorn7 / @brian-brazil


simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+ * The MetricsFilter class exists to provide a high-level filter that enables tunable collection of metrics for Servlet
+ * performance.
+ *
+ * By default, this filter will provide metrics that distinguish 3 levels deep for the request path
+ * (including servlet context path).
+ */
+public class MetricsFilter implements Filter {
+    public static final String PATH_COMPONENT_PARAM = "path-components";
+    public static final int DEFAULT_PATH_COMPONENTS = 3;
+    public static final String FILTER_NAME = "servlet_request_latency";
```
brian-brazil: Inline this.

The word "seconds" should appear in the name

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+    private static int pathComponents = DEFAULT_PATH_COMPONENTS;
+
+    int getPathComponents() {
+        return pathComponents;
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+         if (filterConfig == null) {
+            pathComponents = DEFAULT_PATH_COMPONENTS;
```
brian-brazil: This is 3 levels of indirection to find the value, please simplify.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * By default, this filter will provide metrics that distinguish 3 levels deep for the request path
+ * (including servlet context path).
+ */
+public class MetricsFilter implements Filter {
+    public static final String PATH_COMPONENT_PARAM = "path-components";
+    public static final int DEFAULT_PATH_COMPONENTS = 3;
+    public static final String FILTER_NAME = "servlet_request_latency";
+
+    private static Histogram servletLatency = Histogram.build()
```
brian-brazil: final

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+            return;
+        }
+        if (!StringUtils.isEmpty(filterConfig.getInitParameter(PATH_COMPONENT_PARAM))) {
+            pathComponents = Integer.valueOf(filterConfig.getInitParameter(PATH_COMPONENT_PARAM));
+        }
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        if (servletLatency == null || !(servletRequest instanceof HttpServletRequest)) {
```
brian-brazil: servletLatency can't be null

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class MetricsFilter implements Filter {
+    public static final String PATH_COMPONENT_PARAM = "path-components";
+    public static final int DEFAULT_PATH_COMPONENTS = 3;
+    public static final String FILTER_NAME = "servlet_request_latency";
+
+    private static Histogram servletLatency = Histogram.build()
+            .name(FILTER_NAME)
+            .help("The time taken fulfilling uportal requests")
+            .labelNames("path", "verb")
```
brian-brazil: It's usually method rather than verb.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+        String path = request.getRequestURI();
+        int lastSlash = StringUtils.ordinalIndexOf(path, "/", pathComponents+1);
+
+        Histogram.Timer timer = servletLatency
+            .labels(lastSlash == -1 ? path : path.substring(0, lastSlash), request.getMethod())
+            .startTimer();
+
+        try {
+            filterChain.doFilter(servletRequest, servletResponse);
```
brian-brazil: If this is in an arbitrary position in an arbitrary chain then the user needs to provide a prefix for the metric name.

simpleclient_servlet/src/test/java/io/prometheus/client/filter/MetricsFilterTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Test;
+
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.*;
```
brian-brazil: Please specify imports explicitly.

simpleclient_servlet/src/test/java/io/prometheus/client/filter/MetricsFilterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Created by andrewstuart on 11/28/16.
```
brian-brazil: Please remove this

simpleclient_servlet/src/test/java/io/prometheus/client/filter/MetricsFilterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @Test
+    public void init() throws Exception {
+        FilterConfig cfg = mock(FilterConfig.class);
+        when(cfg.getInitParameter(MetricsFilter.PATH_COMPONENT_PARAM)).thenReturn("4");
+
+        f.init(cfg);
+
+        assertEquals(f.getPathComponents(), 4);
+
+                HttpServletRequest req = mock(HttpServletRequest.class);
```
brian-brazil: Formatting is off here

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/MethodTimer.java
```diff
@@ -0,0 +1,38 @@
...
+ * This class automatically times (via aspectj) the execution of methods by their signature, if it's been enabled via {@link EnablePrometheusTiming}
+ * and in methods annotated (or within classes annotated) with {@link PrometheusTimeMethods}
+ *
+ * @author Andrew Stuart <andrew.stuart2@gmail.com>
+ */
+@Aspect
+@ControllerAdvice
+@Component
+public class MethodTimer {
+    public static final String METRIC_NAME = "prometheus_method_timing";
```
brian-brazil: Inline, should contain the word "seconds"

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/MethodTimer.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public static final Histogram hist = Histogram.build()
+            .name(METRIC_NAME)
+            .help("Automatic annotation-driven method timing")
+            .labelNames("signature")
+            .register();
+
+    @Around("within(@PrometheusTimeMethods *) || execution(@PrometheusTimeMethods * *.*(..))")
+    public Object timeMethod(ProceedingJoinPoint pjp) throws Throwable {
+        Histogram.Timer t = hist.labels(pjp.getSignature().toShortString()).startTimer();
```
brian-brazil: There could be rather a lot of these, I'd go with a quantileless summary by default.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusTimeMethods.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.context.annotation.Import;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Enable Spring-AOP-based automated method timing for the annotated method or all methods in an annotated class.
+ * The timings will be recorded in a Histogram with the metric name of {@code prometheus_method_timing}, and a label
```
brian-brazil: It may be better if the user provided a metric name for each use

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/SpringBootMetricsCollector.java
```diff
@@ -35,12 +35,23 @@ public SpringBootMetricsCollector(Collection<PublicMetrics> publicMetrics) {
   public List<MetricFamilySamples> collect() {
     ArrayList<MetricFamilySamples> samples = new ArrayList<MetricFamilySamples>();
     for (PublicMetrics publicMetrics : this.publicMetrics) {
+
```
brian-brazil: Please revert the formatting changes here, they're only adding noise to the PR

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
andrewstuart: Also just noticed some copypasta: "uportal" is not relevant here.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
andrewstuart: You'll notice in my updated PR that this is still not final, but that it's now being initialized in the `init` method. That was my original intent, and forgot to finalize when I temporarily moved it back out of `init`.

Let me know if there's a better solution that I'm not thinking of.

andrewstuart:
Okay, I think I've addressed most of your comments and added a few features, as well as repackaging the spring-web annotations to their own package to avoid the spring boot dependencies.

Let me know if you have any other suggestions (especially whether HashMap will be performant enough in the MethodTimer class to keep the promise of super low overhead).

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        if (!(servletRequest instanceof HttpServletRequest)) {
+            filterChain.doFilter(servletRequest, servletResponse);
+            return;
+        }
+
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+
+        String path = request.getRequestURI();
```
jfbreault: why not using `request.getServletPath()` or `request.getPathInfo()` if first is empty. You don't want the full uri in the metrics

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,64 @@
...
+            .name(METRIC_NAME)
+            .help("Automatic annotation-driven method timing")
+            .labelNames("signature")
+            .register();
+
+    @Around("@annotation(annot)")
+    public Object timeMethod(ProceedingJoinPoint pjp, PrometheusTimeMethods annot) throws Throwable {
+        Summary summary = defaultSummary;
+        String name = annot.value();
+        if (!StringUtils.isEmpty(name)) {
```
jfbreault: This whole block need to be thread safe

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.HashMap;
+
+/**
+ * This class automatically times (via aspectj) the execution of methods by their signature, if it's been enabled via {@link EnablePrometheusTiming}
+ * and in methods annotated (or within classes annotated) with {@link PrometheusTimeMethods}
+ *
+ * @author Andrew Stuart <andrew.stuart2@gmail.com>
+ */
+@Aspect
+@ControllerAdvice
```
jfbreault: Why `@ControllerAdvice` ?

andrewstuart: My experience was that without this annotation, Spring ignored the `@PrometheusMethodTiming` annotation anywhere inside controllers. This annotation enabled that, which to me is the most applicable use case for high-level monitoring.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
andrewstuart: getRequestURI is only the path, but always includes the whole path and doesn't omit the servlet context path. It's not the proper URI in the [RFC3986](https://tools.ietf.org/html/rfc3986) sense. See the [example table](https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/http/HttpServletRequest.html#getRequestURI()) in the docs. :smile:

I did try the methods you suggest, and ended up landing on `getRequestURI()` as it provides the most helpful fully-contextual info from just looking at the labels in a dashboard.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
andrewstuart: Good catch. I swapped out HashMap for ConcurrentHashMap.

jfbreault: ConcurentHashMap only will not do, it handle only concurent atomic operation. You can end up creating 2 metrics. The only way is to use a syncronize block or using a Lock. Personal I would extract the if in a synchronized method returning the summary and with that you can keepthe original HashMap.

andrewstuart: Alright, I updated this with a ReentrantReadWriteLock since 99% of the time we will probably only be reading the map and this will offer proper multithreaded access for those cases and thus keep this from becoming a bottleneck.

Double (or triple) check my implementation, though, since I'd rather not be introducing deadlocks.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/EnablePrometheusTiming.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Enable the use of {@link PrometheusTimeMethods} annotation on classes or methods.
+ *
+ * Usage: Add this annotation to any Spring {@link org.springframework.context.annotation.Configuration} class to enable
+ * the use of the {@link PrometheusTimeMethods} annotation.
+ *
+ *  * <pre><code>
+ * {@literal @}SpringBootApplication
```
brian-brazil: Are you sure about this?

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,96 @@
...
+ * This class automatically times (via aspectj) the execution of methods by their signature, if it's been enabled via {@link EnablePrometheusTiming}
+ * and in methods annotated (or within classes annotated) with {@link PrometheusTimeMethods}
+ *
+ * @author Andrew Stuart
+ */
+@Aspect
+@ControllerAdvice
+@Component
+public class MethodTimer {
+    public static final String METRIC_NAME = "prometheus_method_timing_seconds";
```
brian-brazil: Inline this.

The prometheus_ prefix implies this is a metric of the prometheus server

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public static final Summary defaultSummary = Summary.build()
+            .name(METRIC_NAME)
+            .help("Automatic annotation-driven method timing")
+            .labelNames("signature")
+            .register();
+
+    private Summary getSummary(PrometheusTimeMethods annot) {
+        Summary summary;
+        String name = annot.value();
```
brian-brazil: Can you make the value the actual Summary?

You don't want something like a hashmap or mutex on this codepath, they're relatively slow.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,96 @@
...
+            return summary;
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    @Around("@annotation(annot)")
+    public Object timeMethod(ProceedingJoinPoint pjp, PrometheusTimeMethods annot) throws Throwable {
+        Summary summary = defaultSummary;
+        if (!StringUtils.isEmpty(annot.value())) {
```
brian-brazil: The suggestion was that the name is required, not optional

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,96 @@
...
+            // and one of them got here first
+            if (collectors.containsKey(name)) {
+                return collectors.get(name);
+            }
+
+            // Only one thread may get here, as writeLock is fully mutually exclusive
+            summary = Summary.build()
+                    .name(name)
+                    .help(annot.help())
+                    .labelNames("signature")
```
brian-brazil: If you have the name being unique, then you don't need this label.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * The Histogram buckets can be configured with a {@code buckets} init parameter whose value is a comma-separated list of valid {@code double} values.
+ *
+ * The Histogram name itself can also be configured with a {@code metric-name} init parameter, whose value will be used directly.
+ */
+public class MetricsFilter implements Filter {
+    public static final String PATH_COMPONENT_PARAM = "path-components";
+    public static final String METRIC_NAME_PARAM = "metric-name";
+    public static final String BUCKET_CONFIG_PARAM = "buckets";
+    public static final String DEFAULT_FILTER_NAME = "servlet_request_latency_seconds";
```
brian-brazil: It's incorrect to have a default here, as if there's multiple uses that'll throw and exception and semantically each use would mean something different.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,98 @@
...
+    public static final int DEFAULT_PATH_COMPONENTS = 3;
+
+    private static Histogram servletLatency = null;
```
brian-brazil: Are you sure this should be static? There's going to be more than one of these

andrewstuart: Right, good point.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
andrewstuart: If it's a class being annotated, we still need to differentiate between each method within the class. My goal in that design choice is that users can still get granular instrumentation without having to annotate each method individually.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
andrewstuart: What would you think about making that optional and a recommended best practice? 

My preference would be toward making instrumentation dead simple to get started with (and increase the positive feedback loop to encourage users onward), while still making it easy to supply your own Summary where performance is desired.

brian-brazil: I mean that you store the Summary as part of the annotation automatically. ConcurrentHashMaps and mutexes are slow, you don't want those as a default.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
brian-brazil: That's tending towards profiling more than metrics, what sort of metric volume are you thinking here?

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
andrewstuart: I guess I'll have to read up on Spring AOP a little more.  My understanding was that it's typically a singleton handling the methods, but it appears there can be advice instances. Any specific suggestions on implementation there?

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
andrewstuart: Well, for one, if it's performant enough I don't see why profiling in production would be a bad thing. Getting that level of visibility into your application's real performance, and the ability to drill down, seems like a win. But that said, my goal was really just to make it easy to set up whatever level of granularity users want, and still provide meaningful output.

In my use case, it's great to have whole spring `@Controller` and `@Service` classes annotated and get metrics on individual method performance, at least at those levels.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
brian-brazil: I've never implemented one of these

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,98 @@
...
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        if (!(servletRequest instanceof HttpServletRequest)) {
+            filterChain.doFilter(servletRequest, servletResponse);
+            return;
+        }
+
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+
+        String path = request.getRequestURI();
+        int lastSlash = StringUtils.ordinalIndexOf(path, "/", pathComponents+1);
```
andrewstuart: The reason for pathComponents is to allow easy limiting of the number of TimeSeries metrics being tracked, at the expense of resolution. Thinking about it, though, 3 is a silly default. I've updated the default to disable that.

Yeah, I could add a request matcher property that can be set, but I'd prefer to get at least something out there before I tack on more complexity.

andrewstuart:
Okay, I think this is ready for another review. There's still some thread safety I had to address, but after the first method call or two it will be read only, and so there's only the potential for thread blocking on the first few calls. The Aspect is now instantiated per target (per `@PrometheusTimeMethods` annotation, if I understand correctly), but after a whole lot of trying, I couldn't figure out a sensible way to access the annotation or join point at Aspect instantiation time. Nor could I have Spring/AspectJ automatically pass me the Annotation. So there's a bit more complexity now thanks to the AOP APIs, or at least what I can understand of them.

Let me know what you think.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/PrometheusTimeMethods.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Enable Spring-AOP-based automated method timing for the annotated method or all methods in an annotated class.
+ * The timings will be recorded in a Histogram with the default metric name of {@code prometheus_method_timing},
+ * which can be overridden by providing a {@code value} parameter in the annotation, and optionally a {@code help}
+ * parameter and a single label of {@code signature} whose values will be {@code ClassOrInterfaceName.methodName(..)}.
+ *
+ * To properly work, {@link EnablePrometheusTiming} must be specified somewhere in your application configuration.
+ *
+ *  <pre><code>
+ * {@literal @}PrometheusTimeMethods("my_app_timer_seconds")
```
brian-brazil: I was talking to a user who had their devs instrumenting all functions and using Prometheus effectively for profiling, causing performance issues. Thus I think it'd be wisest to stick just with method annotations rather than class annotations.

andrewstuart: So were they seeing performance issues in the application itself, due to Prometheus overhead? I guess I was assuming that the metrics overhead (for metrics instances themselves) was pretty negligible compared to the methods being profiled.

I think my main concern is that the primary point of the annotation approach is ease of use, and I'm not sure we do much more than annoy users and force more typing if they really want to annotate every method of a class. If they do want to, they'll still just use the annotation on every method, which just forces them to type more and come up with more help phrases and metric names, and then make sure they maintain the code any time new methods are added in case they want those methods profiled too.

Honestly, I'm not firmly set on allowing class annotations, but to me it feels like the better experience, as long as it goes with the warning that "hey, use metrics sparingly," which seems like it applies regardless of whether the metrics were created with annotations, or by adding all the requisite code themselves.

Thoughts?

brian-brazil: > So were they seeing performance issues in the application itself, due to Prometheus overhead? 

That wasn't reported to me, but I presume not. Though with this level of instrumentation there's likely an inner loop somewhere where it'd cause a problem.

> like the better experience, as long as it goes with the warning that "hey, use metrics sparingly," which seems like it applies regardless of whether the metrics were created with annotations, or by adding all the requisite code themselves.

The problem is that metrics are basically free, as long as you don't use them in an industrial fashion. Class level annotations tip the scales for me where it's too easy to run yourself into problems.

If you really want per-function data for every single function, that's more the job for a profiling tool than a metrics one.

andrewstuart: You know, somehow it hadn't really occurred to me that class methods would be calling each other in the same class, within loops, recursively, etc. I'm too used to languages like Go or JavaScript where methods tend to call functions not associated with an instance.

So yeah, per-method granularity makes a lot more sense.

That should make things a lot more performant too, since as you mentioned, we can cache the child and avoid a lookup.

jfbreault: Not true in Spring AOP, method calling each other in the same class is ignored by aop. Aop is a proxy based stuff.
see https://docs.spring.io/spring/docs/current/spring-framework-reference/html/aop.html#aop-understanding-aop-proxies.

note: it could work with cglib base proxy but i never tested it myself

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Pointcut("annotatedMethod()")
+    public void timeable() {}
+
+    private PrometheusTimeMethod getAnnotation(ProceedingJoinPoint pjp) throws NoSuchMethodException {
+        assert(pjp.getSignature() instanceof MethodSignature);
+
+        MethodSignature signature = (MethodSignature) pjp.getSignature();
+
+        PrometheusTimeMethod annot =  AnnotationUtils.findAnnotation(pjp.getTarget().getClass(), PrometheusTimeMethod.class);
```
brian-brazil: extra space here after the =

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new IllegalStateException("Annotation could not be found for pjp \"" + pjp.toShortString() +"\"", e);
+        }
+
+        assert(annot != null);
+
+        summaryChild = Summary.build()
+                .name(annot.name())
+                .help(annot.help())
+                .register().labels();
+//                .labelNames("signature")
```
brian-brazil: Please remove the debug code.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,94 @@
...
+        } catch (NullPointerException e) {
+            throw new IllegalStateException("Annotation could not be found for pjp \"" + pjp.toShortString() +"\"", e);
+        }
+
+        assert(annot != null);
+
+        summaryChild = Summary.build()
+                .name(annot.name())
+                .help(annot.help())
+                .register().labels();
```
brian-brazil: No labels is special cased already inside the metric, so you can use the metric normally.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,105 @@
...
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+/**
+ * The MetricsFilter class exists to provide a high-level filter that enables tunable collection of metrics for Servlet
```
andrewstuart: Yeah, no problem. The root README, or create a new README in the `simpleclient_servlet` package?

simpleclient_servlet/pom.xml
```diff
@@ -67,5 +67,16 @@
             <version>8.1.7.v20120910</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.4</version>
```
andrewstuart: This is updated; not sure why GitHub hasn't shown the change.

andrewstuart:
With the new change to `pertarget()` I'm now unable to get the code to run in an actual application. I'd like to get a test together to reproduce the issue obviously, but in case anybody (/cc @jfbreault, since you seem to have AOP experience) has some insight on the best fix for this, here's the error I'm seeing:

`Caused by: java.lang.IllegalArgumentException: Bean with name 'methodTimer' is a singleton, but aspect instantiation model is not singleton`

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
brian-brazil: The root README.

The others should have examples too.

andrewstuart: Looks like I was a bit behind `master`, so I'm rebasing. I'll add the examples there as well.

andrewstuart:
Alright, I have that issue fixed (by adding `@Scope("prototype")`), but I'm struggling a bit to even get AOP testing to work with mockMvc, much less reproducing the original issue...

Is that something you'd definitely want to have before merging?

brian-brazil:
This stuff is not something I have experience with, so tests to prevent accidental breakage with would be good.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
checketts: I have a similar filter that I'm using that does the following:
```
String handlerUrl = (String) req.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
if(handlerUrl == null){
  handlerUrl = req.getRequestURI();
}
```

This way the URI matches the `RequestMapping` definition that was chosen. Unfortunately it means you would have to look it up after calling `doFilter` and allowing Spring to add the attribute.

I also will allow you to avoid the path components/matching approach.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,122 @@
...
+    public void init(FilterConfig filterConfig) throws ServletException {
+        Histogram.Builder builder = Histogram.build()
+                .help("The time taken fulfilling servlet requests")
+                .labelNames("path", "method");
+
+        if (filterConfig == null) {
+            throw new ServletException("Please provide a configuration object, even for testing.");
+        }
+
+        if (StringUtils.isEmpty(filterConfig.getInitParameter(METRIC_NAME_PARAM))) {
```
checketts: Please add a constructor that allows passing in any items used in the `FilterConfig`. This allows services using Spring Java Config to add a filter to be able to use compile time checks to provide useful values.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -46,11 +46,16 @@
     public static final String METRIC_NAME_PARAM = "metric-name";
     public static final String BUCKET_CONFIG_PARAM = "buckets";
 
-    private Histogram servletLatency = null;
+    private Histogram histogram = null;
 
     // Package-level for testing purposes
     int pathComponents = 0;
 
+    public MetricsFilter(Histogram histogram, int pathComponents) {
```
checketts: You'll need a default/no args constructor as well.

andrewstuart:
So. A bit of an update on what I've found about Spring AOP that's going to be the deciding factor on the implementation here.

For one, we cannot have a separate instance of MetricsTimer for every method annotated. At the most granular, we get one per class instance (i.e. only one for singleton components in spring) unless we move to AspectJ and require consumers to use the weaver in their compilation process.

So it looks like if we want to handle multiple potential annotations within the same class, we will have to either move to properly handling the chance of concurrency ourselves, and we won't be able to cache the metric instances in the desired way.

Any thoughts on whether to proceed? I am thinking the most sensible way to proceed is basically using the code I have now protecting a single field, but change it to use a hashmap to look up the metric based on the signature.

andrewstuart:
I just pushed a test case that demonstrates what I mean about not being able to cache the metric in the case where a single component is annotated multiple times. The first-registered metric will be used for all the subsequent calls.

brian-brazil:
A ConcurrentHashMap sounds like the way to go, there's a 100-150ns hit though.

andrewstuart:
In an attempt to get below the 100ns barrier, I went with a ReentrantReadWriteLock since 99.9% of the time we should be reading. We only need to actually write on the first time we hit a method that hasn't been called before, so the extra time taken for the lock call should be worth it in the long run since now multiple threads can access the map for reading at the same time.

I'd really like to look at optimizing this out completely if there's a way to just do all the work at instantiation time and stuff things into a completely immutable map so we can just drop synchronization, but I figured this was a good first pass at least.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,121 @@
...
+
+        // When target is an AOP interface proxy but annotation is on class method (rather than Interface method).
+        final String name = signature.getName();
+        final Class[] parameterTypes = signature.getParameterTypes();
+
+        return AnnotationUtils.findAnnotation(pjp.getTarget().getClass().getDeclaredMethod(name, parameterTypes), PrometheusTimeMethod.class);
+    }
+
+    private Summary ensureSummary(ProceedingJoinPoint pjp, String key) throws IllegalStateException {
+        // Only one thread may get here, since this method is synchronized
```
brian-brazil: Be consistent, have a period at the end of all your comments.

brian-brazil: This method isn't synchronized

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,121 @@
...
+                return summary;
+            }
+
+            // Now we know for sure that we have never before registered
+            summary = Summary.build()
+                    .name(annot.name())
+                    .help(annot.help())
+                    .register();
+
+            // Even a rehash will
```
brian-brazil: This comment seems incomplete

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/PrometheusTimeMethod.java
```diff
@@ -0,0 +1,46 @@
+package io.prometheus.client.spring.web;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Enable Spring-AOP-based automated method timing for the annotated method or all methods in an annotated class.
```
brian-brazil: it's just the annotated method now

simpleclient_spring_web/src/test/java/io/prometheus/client/spring/web/MethodTimerTest.java
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+    }
+
+    private interface Time2 {
+        void timeMe() throws Exception;
+        void aSecondMethod() throws Exception;
+    }
+
+//    @PrometheusTimeMethod(name = "test_two", help = "help two")
```
brian-brazil: Don't leave commented out code lying about

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -82,9 +78,10 @@ private Summary ensureSummary(ProceedingJoinPoint pjp, String key) throws Illega
             summary = Summary.build()
                     .name(annot.name())
                     .help(annot.help())
-                    .register();
+                    .register();f
```
brian-brazil: f?

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -0,0 +1,118 @@
...
+            summary = summaries.get(key);
+            if (summary != null) {
+                return summary;
+            }
+
+            // Now we know for sure that we have never before registered
+            summary = Summary.build()
+                    .name(annot.name())
+                    .help(annot.help())
+                    .register();f
```
brian-brazil: f?

andrewstuart: Arg. IntelliJ autosave...

README.md
```diff
@@ -303,6 +303,26 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
...
+the process of manually instaniating and registering your metrics classes. To
+use the metrics annotations, simply add `simpleclient_spring_web` as a
+dependency, annotate a configuration class with `@EnablePrometheusTiming`, then
+annotate your Spring components as such:
+
+```java
+@Controller
+public class MyController {
+  @RequestMapping("/")
+  @PrometheusTimeMethod(name = "main_path_seconds", help = "Some helpful info here")
```
brian-brazil: Per a discussion over in the ruby client, duration_seconds is what we're doing with convention-wise.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -0,0 +1,128 @@
+package io.prometheus.client.filter;
+
+import io.prometheus.client.Histogram;
+import org.apache.commons.lang3.StringUtils;
```
brian-brazil: simpleclient_servlet is for exposing metrics to Prometheus, it shouldn't be pulling in something like apache commons.

This probably belongs in a separate module.

brian-brazil:
I just noticed that this is touching simpleclient_servlet (after, uhm, 3 months), I had thought this was just a Spring-related PR. That'll need some care, as we can't be pulling in unnecessary deps there, as it's the "default" way users expose Prometheus metrics and we don't want to drop them in dependency hell.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
andrewstuart: I can definitely see not wanting to pull in Apache commons. I only needed two functions from there, and I can easily write/pull in both for our use cases.

As far as location, I'm not sure where else it would go. It's a simple servlet metrics filter that has all the same dependencies as the servlet exporter filter, except that I've pulled in Mockito for testing.

Let me know if there is a better place and I can try to update that.

andrewstuart:
Yeah, I think that was there since the beginning but I forgot about it when I was titling the PR.

At this point there should be no dependencies added in the simpleclient_servlet package apart from Mockito for testing, so hopefully that's acceptable there. :-)

Let me know if I'm forgetting anything else.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
brian-brazil: If we can avoid the dep then it can stay here, we'll just need to be clear in the readme where this comes from.

The other consideration is that we'd want a fairly new jetty for the servlet in order to fix whatever bugs/issues there are, however we'd want to have as old a jetty as practical to ensure we're not unintentionally using new features that users stuck on an old version wouldn't have. Try and find a good old stable version I guess.

README.md
```diff
@@ -303,6 +303,71 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
...
+init parameter is not required, although it is highly recommended.  The number
+of buckets is overridable, and can be configured by passing a comma-separated
+string of doubles as the `buckets` init parameter. The granularity of path
+measuring is also configurable, via the `path-components` init parameter. By
+default, the servlet filter will record each path differently, but by setting an
+integer here, you can tell the filter to only record up to the Nth slashes. That
+is, all reqeusts with greater than N "/" characters in the servlet URI path will
+be measured in the same bucket and you will lose that granularity.
+
+Example XML config:
```
brian-brazil: It'd help if you tell the use where to place this

README.md
```diff
@@ -303,6 +303,71 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
...
+    <param-name>metric-name</param-name>
+    <param-value>webapp_metrics_filter</param-value>
+  </init-param>
+  <init-param>
+    <param-name>help</param-name>
+    <param-value>This is the help for your metrics filter</param-value>
+  </init-param>
+  <init-param>
+    <param-name>buckets</param-name>
+    <param-value>.001,.002,.005,.010,.020,.040,.080,.120,.200</param-value>
```
brian-brazil: Stick with the default buckets in examples

README.md
```diff
@@ -303,6 +303,71 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
...
+  </init-param>
+  <init-param>
+    <param-name>path-components</param-name>
+    <param-value>5</param-value>
```
brian-brazil: As this is a histogram which already has a cardinality of 10, we should limit this down to one component to be on the safe side.

README.md
```diff
@@ -303,6 +303,71 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
 SessionFactory sessionFactory = entityManagerFactory.unwrap(SessionFactory.class);
 ```
 
+### Servlet Filter
```
brian-brazil: This should have another #

And below too.

brian-brazil:
Thanks!

There's a few minor style nits, but this review has taken long enough.

PR #170, opened 2016-12-08
Fixes #169 - add thread deadlock metrics

### LINKED ISSUE #169: Add JVM deadlock metrics
dropwizard JVM monitoring has support for "Breakdown of thread states, including deadlocks": http://metrics.dropwizard.io/3.1.0/manual/jvm/
else prometheus is on par with dropwizard metrics - I prefer using prometheus - if this gap could be closed it would be great.

davidkarlsen:
@brian-brazil Here you go.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
@@ -62,21 +62,20 @@ public class ThreadExports extends Collector {
           "Started thread count of a JVM",
           threadBean.getTotalStartedThreadCount()));
 
-
     sampleFamilies.add(
         new GaugeMetricFamily(
         "jvm_threads_deadlocked",
-        "Deadlocked thread count of a JVM",
+        "Cycles of JVM-threads that are in deadlock waiting to acquire object monitors or ownable synchronizers",
         nullSafeArrayLength(threadBean.findDeadlockedThreads())));
 
     sampleFamilies.add(
         new GaugeMetricFamily(
-        "jvm_threads_monitor_deadlocked",
-        "Deadlocked monitor thread count of a JVM",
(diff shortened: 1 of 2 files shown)


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -61,8 +61,24 @@ void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
...
+        nullSafeArrayLength(threadBean.findDeadlockedThreads())));
+
+    sampleFamilies.add(
+        new GaugeMetricFamily(
+        "jvm_threads_monitor_deadlocked",
+        "Deadlocked monitor thread count of a JVM",
+        nullSafeArrayLength(threadBean.findMonitorDeadlockedThreads())));
   }
 
+  private double nullSafeArrayLength(long[] array) {
```
brian-brazil: This should be static technically.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -61,8 +61,24 @@ void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
...
+
+    sampleFamilies.add(
+        new GaugeMetricFamily(
+        "jvm_threads_deadlocked",
+        "Deadlocked thread count of a JVM",
+        nullSafeArrayLength(threadBean.findDeadlockedThreads())));
+
+    sampleFamilies.add(
+        new GaugeMetricFamily(
+        "jvm_threads_monitor_deadlocked",
```
brian-brazil: deadlocked_monitor would be more typical.

I'd suggest expanding the help a bit too so users will get idea of the difference between these two metrics.

brian-brazil:
Thanks!

Issue #165, opened 2016-11-30
Expose `SummaryTimer` logic so clients can use it when calling `observe`


checketts:
The built in timer is very convenient for recording measurement when you know the label before hand. However when you don't, like if `responseStatusCode` is a label, and its value isn't known until the request is finishing up, then the user has to fall back to using `observe`

Too often I've seen users fall back to using `millis` or using Guava's stopwatch (which also returns longs, encouraging millis)

It would be great to expose the Timer as a standalone timer I can read elapsed time after the fact so I can get a double that has the right units (meaning seconds).

A simpler approach that would also go along way would be to expose the nanos to seconds conversion as a static method (https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Summary.java#L159)

Then an observation could be as simple as:

```
long startNanos = System.nanoTime();
//Do work

double durationSeconds = Summary.Timer.durationSeconds(startNanos);
//Record observation
```

brian-brazil:
I don't think we should presume that longs are ns, they could be ms if someone used currentTimeMillis. Both of those functions use different bases too (nanoTime isn't guaranteed to use the unix epoch). Thus a different approach is needed.

> if responseStatusCode is a label, and its value isn't known until the request is finishing up, then the user has to fall back to using observe

It's generally recommended not to export latencies in this way, as there's a good chance a human will misinterpret the data.

checketts:
> It's generally recommended not to export latencies in this way, as there's a good chance a human will misinterpret the data.

Are you recommending against recording `responseStatusCode` as a label? I would definitely want to learn more about the recommended way if you have any links or examples.

I agree regarding treating longs as nanos. I prefer the `Timer` class to keep those details hidden.

brian-brazil:
> Are you recommending against recording responseStatusCode as a label?

On latency metrics, yes.

> I would definitely want to learn more about the recommended way if you have any links or examples.

One metric counting requests broken out by response code, another that's just latency.

checketts:
Closed by #173

PR #172, opened 2016-12-29
Expose `SimpleTimer` to keep timing logic that returns elapsed seconds

checketts:
This is regards to https://github.com/prometheus/client_java/issues/165

As team mates have begun using `Summary` objects, they find that they can't use the built in timer as they have a label like `success="true|false"` they want to include when recording the observation.

I've noticed 2 results: they use something like Guava's `StopWatch` (https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Stopwatch.html) class that returns longs (ie using `Duration.Seconds` truncates) or do the math in elapsed milliseconds as longs. I'm convinced of the Prometheus doc recommendations to use seconds (and @brian-brazil article at https://www.robustperception.io/who-wants-seconds/) so I wanted to create a very simple option that supports seconds as doubles.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
@@ -1,0 +1,69 @@
+package io.prometheus.client;
+
+/**
+ * SimpleTimer, to measure elapsed duration in seconds as a double.
+ *
+ * <p>
+ * This is a helper class intended to measure latencies and encapsulate the conversion to seconds without losing precision.
+ *
+ * <p>
+ * Keep in mind that preferred approaches avoid using this mechanism if possible, since latency metrics broken out by
+ * outcome should be minimized; {@link Summary#startTimer()} and {@link Histogram#startTimer()} are preferred.
+ * Consider moving outcome labels to a separate metric like a counter.
+ *
+ * <p>
+ * Example usage:
+ * <pre>
+ * {@code
+ *   class YourClass {
+ *     static final Summary requestLatency = Summary.build()
+ *         .name("requests_latency_seconds")
+ *         .help("Request latency in seconds.")
+ *         .labelNames("aLabel")
+ *         .register();
+ *
+ *     void processRequest(Request req) {
+ *        SimpleTimer requestTimer = new SimpleTimer();
+ *        try {
+ *          // Your code here.
+ *        } finally {
+ *          requestTimer.labels("aLabelValue").observe(requestTimer.elapsedSeconds());
+ *        }
+ *     }
+ *   }
+ * }
+ * </pre>
+ *
+ */
+public class SimpleTimer {
+  private final long start;
+  static TimeProvider defaultTimeProvider = new TimeProvider();
+  private final TimeProvider timeProvider;
+
+  static class TimeProvider {
+    long nanoTime() {
+      return System.nanoTime();
+    }
+  }
+
+  // Visible for testing.
+  SimpleTimer(TimeProvider timeProvider) {
+    this.timeProvider = timeProvider;
+    start = timeProvider.nanoTime();
+  }
+
+  public SimpleTimer() {
+    this(defaultTimeProvider);
+  }
+
+  /**
+   * @return Measured duration in seconds since {@link SimpleTimer} was constructed.
+   */
+  public double elapsedSeconds() {
+    return elapsedSecondsFromNanos(start, timeProvider.nanoTime());
+  }
+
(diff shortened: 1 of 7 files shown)


simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -138,22 +138,44 @@ protected Child newChild() {
     return new Child(quantiles, maxAgeSeconds, ageBuckets);
   }
 
+
+  public static class SimpleTimer {
+    private final long start;
+    private final TimeProvider timeProvider;
+
+    public SimpleTimer(TimeProvider timeProvider) {
```
brian-brazil: This is only for unitests, it shouldn't be public.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -138,22 +138,44 @@ protected Child newChild() {
...
    * Represents an event being timed.
    */
   public static class Timer {
     private final Child child;
-    private final long start;
+    private final SimpleTimer innerTimer;
     private Timer(Child child) {
       this.child = child;
-      start = Child.timeProvider.nanoTime();
+      innerTimer = new SimpleTimer(Child.timeProvider);
```
brian-brazil: An additional new object on each event (of which there are likely many for a single request) seems like it might get expensive.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -138,22 +138,44 @@ protected Child newChild() {
     return new Child(quantiles, maxAgeSeconds, ageBuckets);
   }
 
+
+  public static class SimpleTimer {
```
brian-brazil: This would apply to Histograms too. It probably makes the most sense for it to be a top-level class.

It should also have javafoc showing correct usage, and pointing users to the simpler methods they probably actually want.

checketts:
@brian-brazil  Thanks for the feedback. I attempted to implement it.

simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
```diff
@@ -0,0 +1,70 @@
...
+ *          // Your code here.
+ *        } finally {
+ *          requestTimer.labels("aLabelValue").observe(requestTimer.elapsedSeconds());
+ *        }
+ *     }
+ *   }
+ * }
+ * </pre>
+ *
+ * Keep in mind that preferred approaches avoid using this mechanism if possible, since latency labels should be minimized and
```
brian-brazil: I'd put this up above the example, as it won't be read otherwise.

Latency labels would imply the latency is going in the label. I'd say latency metrics broken out by outcome.

simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
```diff
@@ -0,0 +1,70 @@
...
+  static TimeProvider defaultTimeProvider = new TimeProvider();
+  private final TimeProvider timeProvider;
+
+  static class TimeProvider {
+    long nanoTime() {
+      return System.nanoTime();
+    }
+  }
+
+  //Visible for testing
```
brian-brazil: Space after //, period at end.

simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
```diff
@@ -0,0 +1,70 @@
+package io.prometheus.client;
+
+import static io.prometheus.client.Collector.NANOSECONDS_PER_SECOND;
```
brian-brazil: Let's avoid static imports.

simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
```diff
@@ -0,0 +1,69 @@
+package io.prometheus.client;
+
+/**
+ * SimpleTimer, to measure elapsed duration as secondas a double.
```
brian-brazil: seconds as

checketts:
Fixed typo

simpleclient/src/main/java/io/prometheus/client/SimpleTimer.java
```diff
@@ -0,0 +1,69 @@
+package io.prometheus.client;
+
+/**
+ * SimpleTimer, to measure elapsed duration as second as a double.
```
brian-brazil: This should be:

to measure elapsed duration in seconds as a double.

brian-brazil:
Thanks!

PR #173, opened 2016-12-30
Add `time()` methods that remove needs for `finally` blocks

checketts:
This allows for lambda syntax friendly timings.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Gauge.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Gauge.java
@@ -181,6 +181,26 @@ public class Gauge extends SimpleCollector<Gauge.Child> implements Collector.Des
     public Timer startTimer() {
       return new Timer(this);
     }
+
+    /**
+     * Executes runnable code (i.e. a Java 8 Lambda) and observes a duration of how long it took to run.
+     *
+     * @param timeable Code that is being timed
+     * @return Measured duration in seconds for timeable to complete.
+     */
+    public double setToTime(Runnable timeable){
+      Timer timer = startTimer();
+
+      double elapsed;
+      try {
+        timeable.run();
+      } finally {
+        elapsed = timer.setDuration();
+      }
+
+      return elapsed;
+    }
+
     /**
      * Get the value of the gauge.
      */
@@ -239,6 +259,16 @@ public class Gauge extends SimpleCollector<Gauge.Child> implements Collector.Des
    */
   public Timer startTimer() {
     return noLabelsChild.startTimer();
+  }
+
+  /**
+   * Executes runnable code (i.e. a Java 8 Lambda) and observes a duration of how long it took to run.
+   *
+   * @param timeable Code that is being timed
+   * @return Measured duration in seconds for timeable to complete.
+   */
+  public double setToTime(Runnable timeable){
+    return noLabelsChild.setToTime(timeable);
   }
 
   @Override
--- a/simpleclient/src/main/java/io/prometheus/client/Histogram.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Histogram.java
@@ -36,6 +36,13 @@ import java.util.Map;
  *          requestTimer.observeDuration();
  *        }
  *     }
+ *
+ *     // Or if using Java 8 lambdas.
+ *     void processRequestLambda(Request req) {
+ *        requestLatency.time(() -> {
+ *          // Your code here.
+ *        });
+ *     }
  *   }
  * }
  * </pre>
@@ -172,6 +179,25 @@ public class Histogram extends SimpleCollector<Histogram.Child> implements Colle
    * {@link SimpleCollector#remove} or {@link SimpleCollector#clear}.
    */
   public static class Child {
+
+    /**
+     * Executes runnable code (i.e. a Java 8 Lambda) and observes a duration of how long it took to run.
+     *
+     * @param timeable Code that is being timed
+     * @return Measured duration in seconds for timeable to complete.
+     */
+    public double time(Runnable timeable) {
+      Timer timer = startTimer();
+
+      double elapsed;
+      try {
+        timeable.run();
+      } finally {
+        elapsed = timer.observeDuration();
+      }
+      return elapsed;
+    }
+
     public static class Value {
       public final double sum;
       public final double[] buckets;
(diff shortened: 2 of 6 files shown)


simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -247,6 +254,20 @@ public Timer startTimer() {
     return noLabelsChild.startTimer();
   }
 
+  /**
+   * Executes runnable code and observes a duration of how long it took to run
```
brian-brazil: Sentence should end with a period.

I'd also mention in here that this is intended for use with Java 8+

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -247,6 +254,20 @@ public Timer startTimer() {
...
+   *
+   * @param timeable Code the is being timed
```
brian-brazil: that is

checketts:
Thanks for the feedback. I made the updates.

simpleclient/src/main/java/io/prometheus/client/Gauge.java
```diff
@@ -181,6 +181,21 @@ public void setToCurrentTime() {
     public Timer startTimer() {
       return new Timer(this);
     }
+
+    /**
+     * Executes runnable code (ie Java 8 Lambda) and observes a duration of how long it took to run.
```
brian-brazil: i.e. a

i.e. is short for "id est", "that is" in Latin.

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -36,6 +36,13 @@
  *          requestTimer.observeDuration();
  *        }
  *     }
+ *
+ *     //Or if using Java 8 lambda
```
brian-brazil: Space after //, period at the end

lambdas

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -247,6 +254,20 @@ public Timer startTimer() {
     return noLabelsChild.startTimer();
   }
 
+  /**
+   * Executes runnable code (ie Java 8 Lambda) and observes a duration of how long it took to run.
+   *
+   * @param timeable Code that is being timed
+   */
+  public void time(Runnable timeable){
```
brian-brazil: We had requests for the existing timers to return the duration for use in logging. These should probably do the same.

simpleclient/src/main/java/io/prometheus/client/Gauge.java
```diff
@@ -181,6 +181,21 @@ public void setToCurrentTime() {
...
+     *
+     * @param timeable Code that is being timed
+     */
+    public void time(Runnable timeable){
```
brian-brazil: I'm not sure on the naming here, as we use setDuration as against observeDuration for the existing timers. This is to reduce the chances that someone will incorrectly use this where they should use a Summary/Historgram.

`setToTime` maybe?

brian-brazil: These need to be added to the convenience functions too, so they'll work as expected for metrics without labels.

checketts: Sorry, I'm not sure which convenience functions you have in mind. Do you just mean the one on `Gauge`? I only see that one I missed.

brian-brazil: There's a pile of them just below this section.

checketts: I should have noted the comment 😄 

So you would like an `inc(Runnable runnable)` method that increments when it is finished? I'm not sure how it would behave for `dec` or `set` though...

```
// Convenience methods.
  /**
   * Increment the gauge with no labels by 1.
   */
  public void inc() {
    inc(1);
  }
  /**
   * Increment the gauge with no labels by the given amount.
   */
  public void inc(double amt) {
    noLabelsChild.inc(amt);
  }
  /**
   * Increment the gauge with no labels by 1.
   */
  public void dec() {
    dec(1);
  }
  /**
   * Decrement the gauge with no labels by the given amount.
   */
  public void dec(double amt) {
    noLabelsChild.dec(amt);
  }
  /**
   * Set the gauge with no labels to the given value.
   */
  public void set(double val) {
    noLabelsChild.set(val);
  }
```

brian-brazil: The important thing about those functions is the noLabelsChild. Each of those mirrors the methods on a Child, so you need to add a single `time` function here.

checketts: OK I think I added the right one.

brian-brazil: Now it's missing from the Child. The convenience functions should be simple wrappers over the core code in the Child.

simpleclient/src/main/java/io/prometheus/client/Gauge.java
```diff
@@ -181,6 +181,26 @@ public void setToCurrentTime() {
...
       return new Timer(this);
     }
+
+    /**
+     * Executes runnable code (ie a Java 8 Lambda) and observes a duration of how long it took to run.
+     *
+     * @param timeable Code that is being timed
+     * @return Measured duration in seconds for timeable to complete.
+     */
+    public double setToTime(Runnable timeable){
```
checketts: Here is the `Guage.Child` implementation

simpleclient/src/main/java/io/prometheus/client/Gauge.java
```diff
@@ -241,6 +261,16 @@ public Timer startTimer() {
...
   }
 
+  /**
+   * Executes runnable code (ie a Java 8 Lambda) and observes a duration of how long it took to run.
+   *
+   * @param timeable Code that is being timed
+   * @return Measured duration in seconds for timeable to complete.
+   */
+  public double setToTime(Runnable timeable){
+    return noLabelsChild.setToTime(timeable);
```
checketts: Here is the convenience method

brian-brazil: Okay, can you add them to Summary and Histogram too?

"i.e." still needs changing in the comment

brian-brazil:
If you could address the remaining comments we can get this submitted and do a release.

checketts:
OK I think I addressed all the comments.

simpleclient/src/main/java/io/prometheus/client/Gauge.java
brian-brazil: That should be i.e. in the comment

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -40,7 +40,15 @@
  *          requestTimer.observeDuration();
  *        }
  *     }
- *   }
+ *
+ *     // or if using Java 8 and lambdas
```
brian-brazil: Capital letter and full stop.

brian-brazil:
Thanks!

PR #178, opened 2017-01-11
Make creating Gauge children Java 8 Lambda friendly

checketts:
This is related to http://stackoverflow.com/questions/41491587/simplest-way-to-export-executor-queue-size-to-prometheus?noredirect=1#comment70189241_41491587

This PR add the ability for  usage like so:

    private static final ReportingGaugeCollector connectedGauge = new ReportingGaugeCollector(
            "rabbitmq_listener_connected",
            "Indicator that device is connected to rabbitmq listener ",
            Arrays.asList("deviceId")).register();

Then in the code for each listener:

    connectedGauge.add(new ReportingGauge() {
            @Override
            public Sample get() {
                return Sample.withLabels(name).value(listener.isConnected() ? 1 : 0);
            }
        });

*I realize that the code in the PR doesn't follow the current builder and javadocs patterns.* If the concept is acceptable, I will spend time refining it. As it is now, this is just mimicking my quick implementation in my codebase.


brian-brazil:
This seems like a use case for either direct instrumentation (update it when isConnected changes), or a custom collector. This doesn't require a new class.

checketts:
I've tried the custom collector and while it can meet the need (mostly). I find myself opening the `GaugeMetricFamily` and copying the JavaDoc. It isn't short or intuitive.

The one case it doesn't address is if I have a metric that has the same name but different labels in multiple instances without resorting the the static `CollectorRegistry`. (And the static `CollectorRegistry.defaultRegistry` causes headaches with unit tests.)

Lastly I want a simple gauge that I can just override the `get()` method of similar to http://metrics.dropwizard.io/3.1.0/getting-started/#gauges

brian-brazil:
> The one case it doesn't address is if I have a metric that has the same name but different labels in multiple instances

This is addressed by direct instrumentation. Have one static metric, and then each instance has its child.

checketts:
> This is addressed by direct instrumentation. Have one static metric, and then each instance has its child.

OK this is good input. What are your thoughts about avoiding the `static` metric and allowing a collector to be looked up via a `CollectorRegistry.get(metricName)` call?

brian-brazil:
> What are your thoughts about avoiding the static metric and allowing a collector to be looked up via a CollectorRegistry.get(collectorName) call?

I consider that to be anti-pattern. Having it as a static member of the class is much simpler.

checketts:
Regarding this PR, I would still like a simple way to add custom collectors that simplifies it down to a single call.

Simplify this:
```
   class YourCustomCollector extends Collector {
      List<MetricFamilySamples> collect() {
        List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
        // With labels
        CounterMetricFamily labeledCounter = new CounterMetricFamily("my_other_counter_total", "help", Arrays.asList("labelname"));
        labeledCounter.addMetric(Arrays.asList("foo"), 4);
        mfs.add(labeledCounter);
        return mfs;
      }
    }
```

Into a one line call (removing the `List<MetricFamilySamples>` essentially). Perhaps calling it `SingleMetricFamilyCollector` so the only implementation needed is to return a `Sample` that contains the labels and value.

brian-brazil:
That's not really possible in Java  (it's a bit simpler in Python).

checketts:
That is the point of this PR: to make it possible.

With Java 8 Lambdas using it would be as simple as:
```
connectedGauge.add(() ->  Sample.withLabels(name).value(listener.isConnected() ? 1 : 0));
```

checketts:
Now that I've wrapped my head around the was `Child` objects are used. I can update the PR to enhance the existing `Gauge` class so usage would be like:
```
someGauge.newChild(() ->  Sample.withLabels("active").value(executor.getActiveCount()))
```

brian-brazil:
I think you're looking for the existing setChild. This is better handled via direct instrumentation I suspect though.

checketts:
Ok! This is nearly exactly what I wanted:
```
Gauge.build().create().setChild(new Gauge.Child() {
            @Override
            public double get() {
                return executor.getActiveCount();
            }
        },"active");
```

It isn't Lambda friendly so I can't shorten it to the one-line syntax. I'll update this PR to show what would be needed.

One missing feature is that the labels can't be dynamic as part of the reporting, but in looking at my use cases that hasn't been a problem. So I'm fine with that.

brian-brazil:
> One missing feature is that the labels can't be dynamic as part of the reporting, but in looking at my use cases that hasn't been a problem. So I'm fine with that.

That can only happen with a custom collector, and is something you would have to be very careful with (https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics).

checketts:
How do you feel about this change?

brian-brazil:
Any support like this belongs up in simplecollector, and I wouldn't wrap the collector.

This is an advanced and very niche use case. I'm satisfied that it's possible (which it is via custom collectors) and wary of making it too easy as it may be abused.

checketts:
I disagree it is advanced. It is the typical Gauge implementation people coming from DropWizard are familiar with. (See http://metrics.dropwizard.io/3.1.0/getting-started/#gauges)

It definitely doesn't pertain to `Counter`s, `Histogram`s, or `Summary`s. So moving it into `SimpleCollector` wouldn't  be a good idea.

I'm not sure what you mean about 'wrapping the collector'.

brian-brazil:
> I disagree it is advanced. It is the typical Gauge implementation people coming from DropWizard are familiar with. (See http://metrics.dropwizard.io/3.1.0/getting-started/#gauges)

The vast majority of uses I've seen of that are cases for either standard direct instrumentation, or a custom collector.

As Prometheus client libraries offer a richer API, use of this should be quite rare.

> It definitely doesn't pertain to Counters, Histograms, or Summarys. So moving it into SimpleCollector wouldn't be a good idea.

Why wouldn't it apply to them? This is just a shorter way of writing simple Custom Collectors.

checketts:
Turns out you were completely right. As I've been progressing through converting more of my metrics, I came across a `Counter` where this approach would also have worked.

So at this point would you like me to move my PR code out of `Gauge` and into the `SimpleCollector` or scrap the idea?

brian-brazil:
I am wondering why you feel you need this so much. I'm having difficulty imagining a case where this would make it so much easier compared to direct instrumentation or a full custom collector.

checketts:
The existing API suits my needs perfectly, so I'm fine closing the PR. At this point it merely was to see if the lambda friendly enhancement would be worthwhile. So I'm perfectly fine closing the PR.

> Any support like this belongs up in simplecollector

Your comment made me wonder if you were requesting me to adjust the PR. That was my only confusion.

Thanks so much for your clear explanations in this PR. It has really helped me understand how to use the java client much better.

PR #181, opened 2017-01-19
Use string interpolation to avoid string +

pjfanning:
@brian-brazil Could you review this change?

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -15,8 +15,8 @@ public class TextFormat {
     /* See http://prometheus.io/docs/instrumenting/exposition_formats/
      * for the output format specification. */
     for (Collector.MetricFamilySamples metricFamilySamples: Collections.list(mfs)) {
-      writer.write("# HELP " + metricFamilySamples.name + " " + escapeHelp(metricFamilySamples.help) + "\n");
-      writer.write("# TYPE " + metricFamilySamples.name + " " + typeString(metricFamilySamples.type) + "\n");
+      writer.write(String.format("# HELP %s %s\n", metricFamilySamples.name, escapeHelp(metricFamilySamples.help)));
+      writer.write(String.format("# TYPE %s %s\n", metricFamilySamples.name, typeString(metricFamilySamples.type)));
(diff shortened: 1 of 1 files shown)


brian-brazil:
What are the benefits of this?

pjfanning:
The code change does not change the functionality but it maintains a consistent coding standard in the class as there is string interpolation already used inside the inner for loop.
Generally, string interpolation is regarded as a better practice than using + to concat strings, eg
http://stackoverflow.com/questions/925423/is-it-better-practice-to-use-string-format-over-string-concatenation-in-java
http://stackoverflow.com/questions/6389827/string-variable-interpolation-java

brian-brazil:
Per those links it looks like + is 10-15x faster, so we should probably be going the other way. Doing the math it's not negligible.

The advantage listed for format is that it's easier to localise, which is irrelevant here.

pjfanning:
If performance is a bigger concern than code readability, then maybe https://github.com/pjfanning/client_java/commit/c8c6cfc3c9f55e44d2006a0a030ef2a77a46ef45 is worth considering. That change avoids creating extra String object instances at the cost of more lines of code.

brian-brazil:
Do you have a vague idea how much that saves vs the +? The + is enough to get us into negligible territory even on relatively crazy setups, but the readability loss may be worth it.

pjfanning:
I'll try to set up a basic benchmark test for the code changes.
@brian-brazil I have a question about TextFormat.java. I noticed that the ',' is added after label name/value pair (line 25 in trunk code) - including the last pair. Is this intentional? It's just that when comma separating list entries, it's usual practice not to have the trailing ','.

brian-brazil:
Yes it's intentional as the code is simpler that way. This is permitted by the format.

pjfanning:
@brian-brazil In a fairly unscientific perf test, the interpolation code is nearly twice as slow as the existing code.
However, the 2nd version, where I avoid the string concats at the expense of more verbose code, is twice as fast as the existing code. 
I can discuss this with you next week if you have time.

*performant version*
https://github.com/pjfanning/client_java/blob/text-format-refactor/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java

*interpolation version*
https://github.com/pjfanning/client_java/blob/text-format-string-interpolation/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java

brian-brazil:
I was expecting a bigger difference against the existing code, so we probably want to go with the optimal version.

Issue #133, opened 2016-07-30
Improve documentation/README and add examples.


oliver006:
I recently had the pleasure to instrument an old Java app and I found it hard to get started as the docs are missing examples.
The README has a large part on the maintenance of the library which isn't really helpful for most people that are just looking for a few simple copy-pastable snippets of code, the docs for the python library[1] is a lot better in this regard.

[1] https://github.com/prometheus/client_python

brian-brazil:
The docs are in javadoc, which you can find at https://prometheus.io/client_java/

oliver006:
Yeah, I eventually found that link but it also doesn't give a good, tutorial style instructions about how to get started with instrumenting my app. All the relevant information is hidden with the respective classes and not easily found.
I also found it lacking with with information about the `client.hotspot` package, it's not quite clear what e.g. the difference is between `StandardExports` and `DefaultExports`.

oliver006:
Let me see if I can find some time in the next few days to get a better README doc cobbled together.

oneiros-de:
The home page says "There are canonical examples defined in the class definition Javadoc of the client packages." - this might be true for _some_ packages, but e.g. the log4j and the jetty package have _zero_ javadoc.

checketts:
I'm interested in taking this on. @oliver006 Do you have any in-flight work you would like me to build on?

oliver006:
Hey @checketts - I started making changes, in particular I moved examples to the README so users can find them faster (similar to the docs of the Python client).
You can find them here: https://github.com/oliver006/client_java/tree/oh_improve_readme 

Please feel free to go ahead and re-use them or discard them as you see fit, I'm a bit pressed for time these days and won't be able to get to this any time soon.

checketts:
@oliver006 Thanks for the response. I did a spike last night. (https://github.com/checketts/client_java/tree/readme)

If you don't mind commenting on #182 if anything appears to be missing I'll be happy to tackle it. (I understand if you are too short on time)

oliver006:
Had a brief look, I really like what you did there!

PR #182, opened 2017-01-20
Mimic python_client readme displaying JavaDoc examples

checketts:
Addresses #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,8 @@ There are Servlet, SpringBoot, and Vert.x integrations included in the client li
 To add Prometheus exposition to an existing HTTP server, see the `MetricsServlet`. 
 It also serves as a simple example of how to write a custom endpoint.
 
+To expose the metrics used in your code, you would add the Prometheus servlet to your Jetty server:
+
 ```java
 context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
 ```
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Github Project Page](http://prometheus.github.io/client_java).
 
 ## Instrumenting
 
-Four types of metric are offered: Counter, Gauge, Summary and Histogram.
+Four types of metrics are offered: Counter, Gauge, Summary and Histogram.
 See the documentation on [metric types](http://prometheus.io/docs/concepts/metric_types/)
 and [instrumentation best practices](http://prometheus.io/docs/practices/instrumentation/#counter-vs.-gauge,-summary-vs.-histogram)
 on how to use them.
--- a/README.md
+++ b/README.md
@@ -93,10 +93,7 @@ class YourClass {
 There are utilities for common use cases:
 
 ```java
-gauge.setToCurrentTime(); //Set to current unixtime
-
-# Increment when entered, decrement when exited. (Not implemented yet)
-inprogressRequests.trackInProgress(() -> doWork());
+gauge.setToCurrentTime(); // Set to current unixtime.
 ```
 
 A Gauge can also take its value from a callback:
@@ -114,10 +111,6 @@ void processRequest() {
       }
   })  
 }
-
-//or (Not implemented yet)
-queueDepth.setFunction(() -> executorService.getQueue().size());
-queueDepth.setFunction(() -> executorService.getQueue().size(), "label");
 ```
 
 ### Summary
@@ -153,8 +146,16 @@ class YourClass {
     .name("requests_latency_seconds").help("Request latency in seconds.").register();
   
   void processRequest(Request req) {
+    requestLatency.timer(new Runnable() {
+      public abstract void run() {
+        // Your code here.    
+      }
+    });  
+      
+      
+    // Or the Java 8 lambda equivalent   
     requestLatency.timer(() -> {
-        // Your code here.
+      // Your code here.
     });
   }
 }
@@ -182,7 +183,7 @@ class YourClass {
 ```
 
 The default buckets are intended to cover a typical web/rpc request from milliseconds to seconds.
-They can be overridden with the `buckets()` method on the `Histogram.Builder`.
+They can be overridden with the `buckets()` method on the [Histogram.Builder](https://prometheus.io/client_java/io/prometheus/client/Histogram.Builder.html#buckets-double...-).
 
 There are utilities for timing code:
 
@@ -192,6 +193,14 @@ class YourClass {
      .name("requests_latency_seconds").help("Request latency in seconds.").register();
   
   void processRequest(Request req) {
+    requestLatency.timer(new Runnable() {
+      public abstract void run() {
+        // Your code here.    
+      }
+    });  
+      
+      
+    // Or the Java 8 lambda equivalent  
     requestLatency.time(() -> {
       // Your code here.  
     });
@@ -221,16 +230,33 @@ class YourClass {
 }
 ```
 
-### Included Exporters
+### Included Collectors
 
-The Java client includes exporters for Garbage Collection, Memory Pools, JMX, and Thread counts.
-
+The Java client includes collectors for garbage collection, memory pools, JMX, classloading, and thread counts.
+These can be added individually or just use the `DefaultExports` to conveniently register them. 
 
 ```java
-new GarbageCollectorExports().register();
-new MemoryPoolsExports().register();
-new StandardExports().register();
-new ThreadExports().register()
+DefaultExports.initialize();
+```
+
+####Logging
+
+There are logging collectors for log4j and logback.
+
+To register the Logback collector can be added to the root level like so:
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <include resource="org/springframework/boot/logging/logback/base.xml"/>
+
+    <appender name="METRICS" class="io.prometheus.client.logback.InstrumentedAppender" />
+
+    <root level="INFO">
+        <appender-ref ref="METRICS" />
+    </root>
+
+</configuration>
 ```
 
 ## Exporting
@@ -245,6 +271,7 @@ There are Servlet, SpringBoot, and Vert.x integrations included in the client li
 
 To add Prometheus exposition to an existing HTTP server, see the `MetricsServlet`. 
 It also serves as a simple example of how to write a custom endpoint.
+
 
 
 ## Exporting to a Pushgateway
@@ -307,7 +334,7 @@ thread.join();
 Sometimes it is not possible to directly instrument code, as it is not
 in your control. This requires you to proxy metrics from other systems.
 
-To do so you need to create a custom collector (which will require to be registered as a normal metric), for example:
+To do so you need to create a custom collector (which will need to be registered as a normal metric), for example:
 
 ```java
 class YourCustomCollector extends Collector {
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Prometheus JVM Client
 It supports Java, Clojure, Scala, JRuby, and anything else that runs on the JVM.
+ 
+[![Build Status](https://travis-ci.org/prometheus/client_java.png?branch=master)](https://travis-ci.org/prometheus/client_java)
 
 ## Using
 ### Assets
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+}
+```
+
+There are utilities for common use cases:
+
+```java
+gauge.setToCurrentTime(); //Set to current unixtime
+
+# Increment when entered, decrement when exited. (Not implemented yet)
+inprogressRequests.trackInProgress(() -> doWork());
```
checketts: I noted this function exists exists in Python. Should I add a parallel in the Java Client?

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+      public double get() {
+          //executorService is a ThreadPoolExecutor
+          return executorService.getQueue().size();
+      }
+  })  
+}
+
+//or (Not implemented yet)
+queueDepth.setFunction(() -> executorService.getQueue().size());
+queueDepth.setFunction(() -> executorService.getQueue().size(), "label");
```
checketts: I noted this function exists exists in Python. Should I add a parallel in the Java Client?

README.md
brian-brazil: I'd like to wait and see how the lambdas go before we start adding lambda-only functionality. The base variant of this doesn't make sense in Java 6.

README.md
brian-brazil: setChild is the Java equivalent.

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
 
 
 [![Build Status](https://travis-ci.org/prometheus/client_java.png?branch=master)](https://travis-ci.org/prometheus/client_java)
+
+## Instrumenting
```
brian-brazil: This should start a bit higher up I think, code examples are the most interesting thing to new users I expect.

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+    // Your code here.
+    inprogressRequest.dec();
+  }
+}
+```
+
+There are utilities for common use cases:
+
+```java
+gauge.setToCurrentTime(); //Set to current unixtime
```
brian-brazil: Space after //, period at end of comment

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+
+# Increment when entered, decrement when exited. (Not implemented yet)
+inprogressRequests.trackInProgress(() -> doWork());
+```
+
+A Gauge can also take its value from a callback:
```
brian-brazil: I'd just put a brief mention of the existence of this function here, as those used to DropWizard may use it where they should be using normal direct instrumentation.

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+
+```java
+class YourClass {
+  static final Summary requestLatency = Summary.build()
+    .quantile(0.5, 0.05)   // Add 50th percentile (= median) with 5% tolerated error
+    .quantile(0.9, 0.01)   // Add 90th percentile with 1% tolerated error
+    .name("requests_latency_seconds").help("Request latency in seconds.").register();
+  
+  void processRequest(Request req) {
+    requestLatency.timer(() -> {
```
brian-brazil: Mention this is java 8 only

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+      // Your code here.
+    } finally {
+      receivedBytes.observe(req.size());
+      requestTimer.observeDuration();
+    }
+  }
+}
+```
+
+There are utilities for timing code and support quantiles:
```
brian-brazil: You should link here the discussion on quantiles, as they're not aggregatable and relatively slow.

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+      // Your code here.
+    } finally {
+      requestTimer.observeDuration();
+    }
+  }
+}
+```
+
+The default buckets are intended to cover a typical web/rpc request from milliseconds to seconds.
+They can be overridden with the `buckets()` method on the `Histogram.Builder`.
```
brian-brazil: Should we link down to the javadoc on these?

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+Taking a counter as an example:
+
+```java
+class YourClass {
+  static final Counter requests = Counter.build()
+     .name("requests_total").help("Total requests.")
+     .labelNames("method").register();
+  
+  void processGetRequest() {
+    requests.labels("get").inc();
```
brian-brazil: Because the label value here is static, what you really should be doing is putting the labels() result in a class variable and then calling inc on it. This is so it gets initialised, and it's also faster.

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+    // Your code here.
+  }
+}
+```
+
+### Included Exporters
```
brian-brazil: Collectors

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+
+The Java client includes exporters for Garbage Collection, Memory Pools, JMX, and Thread counts.
```
brian-brazil: Just point to DefaultExports

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+new ThreadExports().register()
+```
+
+## Exporting
+
+There are several options for exporting metrics.
+
+### HTTP
+
+Metrics are usually exposed over HTTP, to be read by the Prometheus server.
```
brian-brazil: I'd include a minimal servlet example, and link off to the docstrings for the rest.

README.md
brian-brazil: There's the logging ones too

README.md
```diff
@@ -47,3 +47,297 @@ Details for those wishing to develop the library can be found on the [wiki](http
...
+thread.interrupt();
+thread.join();
+```
+
+## Custom Collectors
+
+Sometimes it is not possible to directly instrument code, as it is not
+in your control. This requires you to proxy metrics from other systems.
+
+To do so you need to create a custom collector (which will require to be registered as a normal metric), for example:
```
brian-brazil: will need to be

README.md
checketts: This wouldn't be lambda only, it would be Java 6 compatible, just using `Runnables` as in:

```
inprogressRequests.trackInProgress(new Runnabe() {
  run() {
    //do work
  }
}
```

I'll spin that idea into another pull request so it evolves separately and doesn't hold up the readme update.

README.md
checketts: I added the setChild example, though I still think a first class method like Python exposes is deserved. I'll remove the `setFunction` callout.

README.md
checketts: Are you saying we should steer people away from this technique? I think I'm still missing what you mean by 'direct instrumentation' and how that is preferable to the DropWizard approach (I'm one of those used to DropWizard).

README.md
checketts: Do you mean this one? https://prometheus.io/docs/practices/histograms/#quantiles

README.md
checketts: This example just mimics the JavaDoc. I would like to create a separate section about registering and efficient usage on a followup PR that discusses why to use `static final` and registering metrics and what `Child` is. Is it OK if we leave this as is?

README.md
checketts: OK Added a logback example.

README.md
checketts: I'm not sure what the minimal servlet example would look like in a readme. It would just be a `web.xml` mapping a servlet. Even the DropWizard equivalent just says '... you can map the servlet to whatever path you see fit.' (http://metrics.dropwizard.io/3.1.0/getting-started/#reporting-via-http)

checketts:
I think I've addressed all the comments. I want to do a follow up PR with the following topics:

- Efficiently creating/registering a metrics (`static final` class members)
- What 'Child' means in regards to a metric
- Summarization and links to the metrics naming and recommendations around tracking success failure in a Counter and not a Histogram/Summary

README.md
brian-brazil: Yes

README.md
brian-brazil: Ok

README.md
brian-brazil: Borrow from https://www.robustperception.io/instrumenting-java-with-prometheus/

README.md
brian-brazil: I'd probably not mention setChild so prominently, as it may be used in appropriately. A brief mention that it exists for advanced use cases and a link to the javadoc should be enough.

README.md
brian-brazil: Yes, this was originally added to make a certain Spark use case easier.

The problem with the DW approach is that if often leaves things like concurrency handling up to the user, where they should really just use a Gauge (or DW Counter, which is a Prometheus Gauge).

README.md
```diff
@@ -47,3 +47,324 @@ Details for those wishing to develop the library can be found on the [wiki](http
 
 
 [![Build Status](https://travis-ci.org/prometheus/client_java.png?branch=master)](https://travis-ci.org/prometheus/client_java)
+
+## Instrumenting
+
+Four types of metric are offered: Counter, Gauge, Summary and Histogram.
```
checketts: Thanks for the proofread!

README.md
```diff
@@ -34,16 +36,342 @@ version can be found on in the maven repository for
...
+
+Metrics are usually exposed over HTTP, to be read by the Prometheus server.
+
+There are Servlet, SpringBoot, and Vert.x integrations included in the client library.
+
+To add Prometheus exposition to an existing HTTP server, see the `MetricsServlet`. 
+It also serves as a simple example of how to write a custom endpoint.
+
+```java
+context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
```
brian-brazil: Include the few lines before this, enough to get it running.

checketts: You mean this?

> Setup
>
> There’s a small bit of setup work that need to be done once, no matter how many metrics you have. To expose the metrics used in your code, we add the Prometheus servlet to our Jetty server:

brian-brazil: I mean the few lines of code you need with that to create the server and context.

README.md
```diff
@@ -34,16 +36,344 @@ version can be found on in the maven repository for
...
+}
+```
+
+There are utilities for common use cases:
+
+```java
+gauge.setToCurrentTime(); // Set to current unixtime.
+```
+
+As an advanced use case, a `Gauge` can also take its value from a callback. Keep in mind that the default `inc()`, `dec()` and `set()`
```
brian-brazil: As already discussed, link to the javadoc here rather than providing an example.

checketts: Are you requesting I remove the `setChild` example?  There is no example on Gauge in its JavaDocs and no one will be hunting in parent classes for the example. I know I wouldn't have found it without your help in #178 . 

I'll remove the example, but I still can't wrap my head around why you think it shouldn't have up front documentation (aside from the python equivalent `setFunction` being a bit better named than `setChild`) and reminding people that they need to account for concurrency.

brian-brazil: > There is no example on Gauge in its JavaDocs and no one will be hunting in parent classes for the example.

Link to the example we have.

> I still can't wrap my head around why you think it shouldn't have up front documentation

Basically dropwizard has taught some developers that this is the way to handle gauges, whereas with Prometheus client libraries this is rarely the appropriate way to do things. Thus I want to make it less likely that users will tie themselves up in knots trying to use this dropwizard-style.

checketts:
I believe all points have been addressed.

brian-brazil:
Great! Thanks for working through this, it'll make life easier for a lot of people.

oliver006:
🎉

PR #185, opened 2017-01-23
MetricFamily#addMetric() should return this to allow builder like usage

chkal:
Currently all `addMetric()` methods on `MetricFamily` implementations have a `void` return type. A typical usage would look like this:

```java
CounterMetricFamily counter =
    new CounterMetricFamily("foobar", "Foobar", Arrays.asList("label"));
counter.addMetric(Arrays.asList("value1"), 4);
counter.addMetric(Arrays.asList("value2"), 5);
```

This PR changes the return type of the `addMetric()` methods to `this` which now allows builder like usage to create metric families:

```java
CounterMetricFamily counter =
    new CounterMetricFamily("foobar", "Foobar", Arrays.asList("label"))
        .addMetric(Arrays.asList("value1"), 4)
        .addMetric(Arrays.asList("value2"), 5);
```

This is especially useful when implementing `Collector.collect()` and creating many metric families.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CounterMetricFamily.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CounterMetricFamily.java
@@ -48,10 +48,11 @@ public class CounterMetricFamily extends Collector.MetricFamilySamples {
     this.labelNames = labelNames;
   }
 
-  public void addMetric(List<String> labelValues, double value) {
+  public CounterMetricFamily addMetric(List<String> labelValues, double value) {
     if (labelValues.size() != labelNames.size()) {
       throw new IllegalArgumentException("Incorrect number of labels.");
     }
     samples.add(new Sample(name, labelNames, labelValues, value));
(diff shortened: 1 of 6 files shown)


brian-brazil:
Can you add unittests for these?

chkal:
I added tests for the builder style usage.

A small note: `SummaryMetricFamilyTest.java` was using tabs instead of spaces. So I'm also using tabs in the new test, which looks pretty weird in GitHub. But at least it is consistent in the file. ;-)

brian-brazil:
That's odd, it should all be spaces.

You can also reduce the vertical whitespace in the tests.

chkal:
I removed the blank lines.

I kept the tabs for now so the file is formatted consistently. I could send a separate PR for converting the whole file from tabs to spaces if you like.

brian-brazil:
Thanks!

chkal:
You're welcome!

PR #186, opened 2017-01-24
New Hibernate integration module

chkal:
This pull requests provides a new module for integration with [Hibernate](http://hibernate.org/orm/). Hibernate provides various statistics using the [Statistics](https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/stat/Statistics.html) interface which can be accessed though the `SessionFactory`.

A few notes:
 * The module depends on the oldest Hibernate version which is still supported. IMO this is reasonable to support Hibernate versions which are typically used in real world projects.
 * I followed the "Best Practices" guide regarding the metric & label naming. However, every feedback is welcome.
 * The unit test mocks the Hibernate Statistics interface to check that all metrics are resolved and named correctly.
 * The help texts for the metrics are copied from the corresponding Javadocs of the `Statistics` interface.

Any kind of feedback is welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,22 +247,22 @@ To register the Logback collector can be added to the root level like so:
 #### Hibernate
 
 There is a collector for Hibernate which allows to collect metrics from one or more 
-`SessionFactory` instances. To use this collector add the following dependency:
+`SessionFactory` instances. 
 
-```xml
-<dependency>
-  <groupId>io.prometheus</groupId>
-  <artifactId>simpleclient_hibernate</artifactId>
-  <version>0.0.20</version>
-</dependency>
+If you want to collect metrics from a single `SessionFactory`, you can register
+the collector like this:
+
+```java
+new HibernateStatisticsCollector(sessionFactory, "myapp").register();
 ```
 
-Now you can create the collector, add your `SessionFactory` instances and register 
-the collector.
+In some situations you may want to collect metrics from multiple factories. In this
+case just call `add()` on the collector for each of them.
 
 ```java
 new HibernateStatisticsCollector()
-    .add(sessionFactory, "myapp")
+    .add(sessionFactory1, "myapp1")
+    .add(sessionFactory2, "myapp2")
     .register();
 ```
 
--- a/README.md
+++ b/README.md
@@ -266,12 +266,11 @@ new HibernateStatisticsCollector()
     .register();
 ```
 
-If you are using Hibernate in a JPA environment and only have access to the `EntityManager`,
-you can use this code to access the underlying `SessionFactory`:
+If you are using Hibernate in a JPA environment and only have access to the `EntityManager`
+or `EntityManagerFactory`, you can use this code to access the underlying `SessionFactory`:
 
 ```java
-SessionFactory sessionFactory =
-    entityManager.getEntityManagerFactory().unwrap(SessionFactory.class);
+SessionFactory sessionFactory = entityManagerFactory.unwrap(SessionFactory.class);
 ```
 
 ## Exporting
--- a/README.md
+++ b/README.md
@@ -244,6 +244,36 @@ To register the Logback collector can be added to the root level like so:
 </configuration>
 ```
 
+#### Hibernate
+
+There is a collector for Hibernate which allows to collect metrics from one or more 
+`SessionFactory` instances. To use this collector add the following dependency:
+
+```xml
+<dependency>
+  <groupId>io.prometheus</groupId>
+  <artifactId>simpleclient_hibernate</artifactId>
+  <version>0.0.20</version>
+</dependency>
+```
+
+Now you can create the collector, add your `SessionFactory` instances and register 
+the collector.
+
+```java
+new HibernateStatisticsCollector()
+    .add(sessionFactory, "myapp")
+    .register();
+```
+
+If you are using Hibernate in a JPA environment and only have access to the `EntityManager`,
+you can use this code to access the underlying `SessionFactory`:
+
+```java
+SessionFactory sessionFactory =
+    entityManager.getEntityManagerFactory().unwrap(SessionFactory.class);
+```
+
 ## Exporting
 
 There are several options for exporting metrics.
--- a/simpleclient_hibernate/pom.xml
+++ b/simpleclient_hibernate/pom.xml
@@ -38,7 +38,7 @@
         <dependency>
             <groupId>io.prometheus</groupId>
             <artifactId>simpleclient</artifactId>
-            <version>0.0.20-SNAPSHOT</version>
+            <version>0.0.21-SNAPSHOT</version>
         </dependency>
 
         <!-- We support Hibernate versions >= 4.2 -->
--- a/simpleclient_hibernate/pom.xml
+++ b/simpleclient_hibernate/pom.xml
@@ -45,7 +45,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
-            <version>4.2.21.Final</version>
+            <version>4.2.0.Final</version>
             <scope>provided</scope>
         </dependency>
 
--- a/simpleclient_hibernate/pom.xml
+++ b/simpleclient_hibernate/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.prometheus</groupId>
         <artifactId>parent</artifactId>
-        <version>0.0.20-SNAPSHOT</version>
+        <version>0.0.21-SNAPSHOT</version>
     </parent>
 
     <groupId>io.prometheus</groupId>
--- a/simpleclient_hibernate/pom.xml
+++ b/simpleclient_hibernate/pom.xml
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.prometheus</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.0.20-SNAPSHOT</version>
+    </parent>
+
+    <groupId>io.prometheus</groupId>
+    <artifactId>simpleclient_hibernate</artifactId>
+    <packaging>bundle</packaging>
+
+    <name>Prometheus Java Simpleclient Hibernate</name>
+    <description>
+        Collectors of data from Hibernate.
+    </description>
+
+    <licenses>
+        <license>
+            <name>The Apache Software License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <developers>
+        <developer>
+            <id>chkal</id>
+            <name>Christian Kaltepoth</name>
+            <email>christian@kaltepoth.de</email>
+        </developer>
+    </developers>
+
+    <dependencies>
+        
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient</artifactId>
+            <version>0.0.20-SNAPSHOT</version>
+        </dependency>
+
+        <!-- We support Hibernate versions >= 4.2 -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>4.2.21.Final</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Testing -->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>2.6.8</version>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+</project>
--- a/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
+++ b/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
@@ -16,10 +16,15 @@ import org.hibernate.stat.Statistics;
 /**
  * Collect metrics from one or more Hibernate SessionFactory instances.
  * <p>
- * Usage example:
+ * Usage example for a single session factory:
+ * <pre>
+ * new HibernateStatisticsCollector(sessionFactory, "myapp").register();
+ * </pre>
+ * Usage example for multiple session factories:
  * <pre>
  * new HibernateStatisticsCollector()
- *     .add(sessionFactory, "myapp")
+ *     .add(sessionFactory1, "myapp1")
+ *     .add(sessionFactory2, "myapp2")
  *     .register();
  * </pre>
  * If you are in a JPA environment, you can obtain the SessionFactory like this:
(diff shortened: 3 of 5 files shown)


brian-brazil:
> The module depends on the oldest Hibernate version which is still supported. IMO this is reasonable to support Hibernate versions which are typically used in real world projects.

That depends on the support policy. That'd not be sane for Java or Ruby for example. 4.2 came out about 4 years ago, so that should be safe in this case. Supporting only 4.2.21 would be a problem, as that's only 14 months old.

simpleclient_hibernate/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient</artifactId>
+            <version>0.0.20-SNAPSHOT</version>
+        </dependency>
+
+        <!-- We support Hibernate versions >= 4.2 -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>4.2.21.Final</version>
```
brian-brazil: Use an older version to be sure we're compatible all the way back to 4.2.0

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,282 @@
...
+import java.util.List;
+import org.hibernate.SessionFactory;
+import org.hibernate.stat.Statistics;
+
+/**
+ * Collect metrics from Hibernate statistics.
+ * <p>
+ * Usage example:
+ * <pre>
+ * new HibernateStatisticsCollector(sessionFactory).register();
```
brian-brazil: Can there be multiple session factories in a jvm?

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,282 @@
...
+        createGauge(
+            "hibernate_natural_id_query_execution_seconds_max",
+            "Get the global maximum query time for naturalId queries executed against the database",
+            statistics.getNaturalIdQueryExecutionMaxTime() / 1000.0
+        )
+
+    );
+  }
+
+  private CounterMetricFamily createCounter(String name, String help, double value) {
```
brian-brazil: These aren't adding much, inline them.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,282 @@
...
+    metrics.addAll(getEntityMetrics());
+    metrics.addAll(getQueryExecutionMetrics());
+    return metrics;
+  }
+
+  private List<MetricFamilySamples> getSessionMetrics() {
+    return Arrays.<MetricFamilySamples>asList(
+
+        createCounter(
+            "hibernate_session_open_total",
```
brian-brazil: opened

"open" implies the current number of open sessions


Similarly for the other counters.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,282 @@
...
+
+        createCounter(
+            "hibernate_session_close_total",
+            "Global number of sessions closed",
+            statistics.getSessionCloseCount()
+        ),
+
+        createCounter(
+            "hibernate_flush_total",
+            "Get the global number of flush executed by sessions",
```
brian-brazil: s/Get /

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,282 @@
...
+    metrics.addAll(getQueryExecutionMetrics());
+    return metrics;
+  }
+
+  private List<MetricFamilySamples> getSessionMetrics() {
+    return Arrays.<MetricFamilySamples>asList(
+
+        createCounter(
+            "hibernate_session_open_total",
+            "Global number of sessions opened",
```
brian-brazil: It wouldn't hurt to put the function name in the help string, it'd save users debugging a hop through this code.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,282 @@
...
+
+        createCounter(
+            "hibernate_query_execution_total",
+            "Get global number of executed queries",
+            statistics.getQueryExecutionCount()
+        ),
+
+        createGauge(
+            "hibernate_query_execution_seconds_max",
+            "Get the time in milliseconds of the slowest query.",
```
brian-brazil: Over what time period? We generally drop metrics like this as they're not particularly useful statistically.

simpleclient_hibernate/pom.xml
chkal: Ok, I'll change that.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: Yes. Especially in application servers. Actually I was thinking about adding a label identifying the `SessionFactory` to the metrics. But it looks like there isn't something like a "name" of a `SessionFactory`.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: Ok, I changed the names where it makes sense.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: Ok, I see. That's due to the fact that I copied the texts from the original javadocs. I'll change that.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: You are talking about the method name used to gather the metric value? Something like:

    Global number of sessions opened (getSessionOpenCount)

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: AFAIK this is since the session factory startup. I wasn't sure whether it makes sense to add these when I wrote the code, so I've no problem with dropping them.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: Ok! I wasn't sure whether there will be some common label for all metrics. That's why I added this method. But I'll drop it.

chkal:
I pushed the changes! Any further feedback welcome! :-)

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
brian-brazil: The user will need to specify the name then when creating the collector.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,261 @@
...
+
+  private List<MetricFamilySamples> getCacheMetrics() {
+    return Arrays.<MetricFamilySamples>asList(
+
+        new CounterMetricFamily(
+            "hibernate_second_level_cache_hit_total",
+            "Global number of cacheable entities/collections successfully retrieved from the cache",
+            statistics.getSecondLevelCacheHitCount()
+        ),
+        new CounterMetricFamily(
```
brian-brazil: if you're going to remove the vertical whitespace here, remove it everywhere

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,261 @@
...
+
+        new CounterMetricFamily(
+            "hibernate_flushed_total",
+            "The global number of flush executed by sessions",
+            statistics.getFlushCount()
+        ),
+
+        new CounterMetricFamily(
+            "hibernate_connect_total",
+            "The global number of connections asked by the sessions",
```
brian-brazil: either "asked for" or "requested"

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,261 @@
...
+
+        new CounterMetricFamily(
+            "hibernate_session_closed_total",
+            "Global number of sessions closed",
+            statistics.getSessionCloseCount()
+        ),
+
+        new CounterMetricFamily(
+            "hibernate_flushed_total",
+            "The global number of flush executed by sessions",
```
brian-brazil: flushes

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
chkal: I added a name to the constructor. This name is then added as a label.

chkal:
I just pushed the following changes:

 * I removed the blank lines
 * The user can now specify a name which is added as a label to the metrics
 * I renamed the metrics as requested
 * I added the Hibernate `Statistics` method names to the help text.

Any more feedback?

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,249 @@
...
+ * <pre>
+ * SessionFactory sessionFactory =
+ *     entityManager.getEntityManagerFactory().unwrap(SessionFactory.class);
+ * </pre>
+ *
+ * @author Christian Kaltepoth
+ */
+public class HibernateStatisticsCollector extends Collector {
+
+  private static final List<String> LABEL_NAMES = Collections.singletonList("name");
```
brian-brazil: Name is a bit generic.

Considering this is likely a per-db connection pool, I'd suggest database or pool.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,249 @@
...
+            "Global number of sessions closed (getSessionCloseCount)",
+            statistics.getSessionCloseCount()
+        ),
+        createCounter(
+            "hibernate_flushes_total",
+            "The global number of flush executed by sessions (getFlushCount)",
+            statistics.getFlushCount()
+        ),
+        createCounter(
+            "hibernate_connection_requested_total",
```
brian-brazil: I was talking more about the help text than the metric name.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,249 @@
...
+ * </pre>
+ *
+ * @author Christian Kaltepoth
+ */
+public class HibernateStatisticsCollector extends Collector {
+
+  private static final List<String> LABEL_NAMES = Collections.singletonList("name");
+
+  private final Statistics statistics;
+  private final String name;
```
brian-brazil: Thinking a bit this isn't going to work with 2 collectors, as they're both outputting the same metric families.

What we probably need is that there's one collector that can handle multiple factories, so in the simple case it'll just have one and then there's another function to add more as needed. You'd then have to iterate through them at collection time.

chkal: Ah, OK. I wasn't aware that different collectors cannot produce the same metric name. But I just found the corresponding code in the registry.

Is there any specific reason why this isn't supported? I could imagine that if a single collector is responsible for monitoring "a thing" (like a DB connection pool), that it would make sense that a collector monitors just a single instance and that users can register the collector multiple times with different configuration. In this case the metric name would of cause be the same and the collector instances would just add different labels. Is there any reason why this isn't supported?

However, if you don't want to change the current behavior, I could implement something like:

```java
new HibernateStatisticsCollector()
    .add(sessionFactory1, "unit1")
    .add(sessionFactory2, "unit2")
    .register();

File: simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java (hunk 12 above)
Comment (chkal) [inline simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java]: Ah, OK... Sorry about that. I'll undo the change.

File: simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java (hunk 13 above)
Comment (brian-brazil) [inline simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java]: Usually with direct instrumentation this doesn't come up, as you'd be sharing the metrics across instances as class variables. The majority of custom collectors also only monitor one thing, so the Java client keeps all the handling simple.

If you could change your code that'd be great.

File: simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java (hunk 11 above)
Comment (brian-brazil) [inline simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java]: Okay, if you think that's best.

File: simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java (hunk 13 above)
Comment (chkal) [inline simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java]: Okay. I'll add support for multiple session factories to the collector. I'll come back to you as soon as this is done.

**Comment by brian-brazil** [comment]
The docs change has also been merged in now, so please add an example to the README.md.

**Comment by chkal** [comment]
I just pushed the following changes:

 * I fixed the help texts as requested
 * The collector now supports multiple session factories
 * The `name` label has been renamed to `unit`
 * I added a section about the collector to the new README file 

A small note: I had to rebase this branch against master to get the current version of the README.

File: README.md [hunk 14]
```diff
@@ -244,6 +244,36 @@ To register the Logback collector can be added to the root level like so:
 </configuration>
 ```
 
+#### Hibernate
+
+There is a collector for Hibernate which allows to collect metrics from one or more 
+`SessionFactory` instances. To use this collector add the following dependency:
+
+```xml
+<dependency>
```
brian-brazil: We haven't included this with the others, and I'd prefer not to have another place to manually update version numbers.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,446 @@
...
+ * </pre>
+ * If you are in a JPA environment, you can obtain the SessionFactory like this:
+ * <pre>
+ * SessionFactory sessionFactory =
+ *     entityManager.getEntityManagerFactory().unwrap(SessionFactory.class);
+ * </pre>
+ *
+ * @author Christian Kaltepoth
+ */
+public class HibernateStatisticsCollector extends Collector {
```
brian-brazil: I think an explicit constrcutor would be clearer to those reading the code, and a convenience constructor that also did the add would be useful too.

chkal:
I've added two constructors and updated the README as requested.

chkal:
I just added a minor improvement for the documentation regarding how to use this collector in a JPA environment.

So all your comments are addressed now. Let me know if there is anything left I can do to get this merged.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -0,0 +1,472 @@
...
+ *     entityManagerFactory.unwrap(SessionFactory.class);
+ * </pre>
+ *
+ * @author Christian Kaltepoth
+ */
+public class HibernateStatisticsCollector extends Collector {
+
+  private static final List<String> LABEL_NAMES = Collections.singletonList("unit");
+
+  private final Map<String, SessionFactory> sessionFactories = new LinkedHashMap<String, SessionFactory>();
```
brian-brazil: Hmm, let's make this a ConcurrentHashMap so that different threads can add themselves without having to worry about synchronisation.

chkal: Okay, makes sense! I replaced it with a ConcurrentHashMap.

brian-brazil:
Thanks!

chkal:
Awesome! Thank you!

PR #187, opened 2017-01-29
Add `build()` method that has required fields.

checketts:
The builder pattern is ideal when you have optional parameters you would like to leave out of a constructor. However, when there are values that are required to be set, it is good to include them as part of the builder's constructor, making it obvious that those values are needed.

On several occasions I've created a metric and forgotten the `help()` part (or even the name part a couple of times). Luckily it blows up when registering the metric, but that means it is a runtime check instead of a compile time check.


brian-brazil:
Thanks!

PR #188, opened 2017-01-29
Add metrics registration and 'child' explanation to readme

checketts:


README.md
```diff
@@ -215,6 +214,29 @@ class YourClass {
 }
 ```
 
+## Registering Metrics
+
+The best way to register a metric is via a `static final` class variable as is common with loggers.
+
+```java
+static final Counter requests = Counter.build()
+   .name("requests_total").help("Total requests.").labelNames("path").register();
```
brian-brazil: I'd namespace this as my_library_... or similar

README.md
```diff
@@ -215,6 +214,29 @@ class YourClass {
...
+```
+ 
+Using the default registry with variables that are `static` is ideal since registering a metric with the same name 
+is not allowed and the default registry is also itself static. You can think of registering a metric, more like 
+registering a definition (as in the `TYPE` and `HELP` sections). The actual samples are tracked via `Child` objects.
```
brian-brazil: I'm a little wary of mentioning Childs in too much detail, these docs would leave you confused as to what happens for a metric with no labels for example (and most metrics have no labels).

README.md
```diff
@@ -215,6 +214,29 @@ class YourClass {
...
+
+`Child` metrics aren't children in the inheritance sense of the word. They are the family members of the given metric.
+Metric definition now behave as mini-registries of the children that are defined for it. 
+
+Given the `requests` counter defined above, the `label()` method looks up or creates the corresponding `Child` metric.
+For that reason it can make sense to store the `Child` metrics when the labels are well known as in the 
+[Logback](https://github.com/prometheus/client_java/blob/master/simpleclient_logback/src/main/java/io/prometheus/client/logback/InstrumentedAppender.java#L13)
```
brian-brazil: It's actually a bit unusual to have the Childs as static, normally they're instance fields.

README.md
checketts: I know what you mean. The name `Child` also adds to the confusion. I initially had trouble trying to register metrics since I would register 2 metrics with the same name so I could add labels later. (Yes, I realize I was approaching it the wrong way).

Since opening this PR I've had to teach several devs to add metrics and the biggest hurdle for each is actually in the registration. 

Should I try removing the references to the `Child` construct and replace it with 'labelled metrics'.

brian-brazil: That sounds like a good approach.

checketts:
OK I've removed the references to `Child`

README.md
```diff
@@ -271,9 +271,9 @@ class YourClass {
 }
 ```
 
-Metrics with labels include `labelNames()` when building the metric and , the `label()` method looks up or creates 
+To create metrics with labels include `labelNames()` when building the metric. The `label()` method looks up or creates 
 the corresponding labelled metric. You might also consider storing the labelled metric as an instance variable if it is
```
brian-brazil: s/metric/timeseries/

README.md
```diff
@@ -271,9 +271,9 @@ class YourClass {
 }
 ```
 
-Metrics with labels include `labelNames()` when building the metric and , the `label()` method looks up or creates
```
brian-brazil: labels()

README.md
```diff
@@ -215,7 +239,57 @@ class YourClass {
...
+ 
+Using the default registry with variables that are `static` is ideal since registering a metric with the same name 
+is not allowed and the default registry is also itself static. You can think of registering a metric, more like 
+registering a definition (as in the `TYPE` and `HELP` sections). The metric 'definition' internally holds the samples 
+that are reported and pulled out by Prometheus. Here is an example of registering a metric that has no labels.
+
+```java
+class YourClass {
+  static final Gauge activeTransactions = Gauge.build()
+     .name("my_library_active_transaction")
```
brian-brazil: Plural is usual, and transactions_active as there may be other transaction-related stats.

README.md
brian-brazil: This is still outstanding

README.md
```diff
@@ -215,7 +239,57 @@ class YourClass {
...
+    } finally{
+        activeTransactions.dec();
+    }
+  }
+}
+```
+
+To create timeseries with labels, include `labelNames()` with the builder. The `labels()` method looks up or creates 
+the corresponding labelled timeseries. You might also consider storing the labelled timeseries as an instance variable if it is
+appropriate for performance, it is thread safe and can be used multiple times.
```
brian-brazil: appropriate. It is thread safe and can be used multiple times, which can help performance.

PR #192, opened 2017-01-31
Add Gauva metrics collector

checketts:
Add support for Guava metrics

I would love feedback regarding the metrics names that are built and labels. I'm still getting the hang of the naming convention and labels.

```
myapp_mycache_cache_request_total{found="hit",} 10.0
myapp_mycache_cache_request_total{found="miss",} 3.0
myapp_mycache_cache_eviction_total 1.0
myapp_mycache_cache_load_total{success="success",} 5.0
myapp_mycache_cache_load_total{success="exception",} 1.0
myapp_mycache_cache_load_sum_seconds 0.0034
```

### DIFF
--- a/simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
+++ b/simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
@@ -1,0 +1,83 @@
+package io.prometheus.client.cache.guava;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheStats;
+import com.google.common.cache.LoadingCache;
+import io.prometheus.client.Collector;
+import io.prometheus.client.CounterMetricFamily;
+import io.prometheus.client.Summary;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+
+/**
+ * Collect metrics from Guava's com.google.common.cache.Cache.
+ * <p>
+ * <pre>{@code
+ *
+ * // Note that `recordStats()` is required to gather non-zero statistics
+ * Cache<String, String> cache = CacheBuilder.newBuilder().recordStats().build();
+ * new CacheMetricsCollector(cache, "myapp_mycache", "Names cache").register();
+ *
+ * }</pre>
+ *
+ * Exposed metrics are prefixed with the provided prefix.
+ *
+ * With the examle above sample metric names would be:
+ * <pre>
+ *     myapp_mycache_cache_request_total{found="hit",} 10.0
+ *     myapp_mycache_cache_request_total{found="miss",} 3.0
+ *     myapp_mycache_cache_eviction_total 1.0
+ * </pre>
+ *
+ * Additionally if the cache includes a loader additional metrics would be provided:
+ * <pre>
+ *     myapp_mycache_cache_load_total{success="success",} 5.0
+ *     myapp_mycache_cache_load_total{success="exception",} 1.0
+ *     myapp_mycache_cache_load_sum_seconds 0.0034
+ * </pre>
+ *
+ */
+public class CacheMetricsCollector extends Collector {
+
+    private final Cache cache;
+    private final String cacheName;
+    private final String help;
+
+    public CacheMetricsCollector(Cache cache, String cacheName, String help) {
+        this.cache = cache;
+        this.cacheName = cacheName;
+        this.help = help;
+    }
+
+    @Override
+    public List<MetricFamilySamples> collect() {
+        List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
+
+        CacheStats stats = cache.stats();
+
+        CounterMetricFamily cacheRequestTotal = new CounterMetricFamily(cacheName+"_cache_request_total",
+                help + " cache request totals", Arrays.asList("found"));
+        mfs.add(cacheRequestTotal);
+        cacheRequestTotal.addMetric(Arrays.asList("hit"), stats.hitCount());
+        cacheRequestTotal.addMetric(Arrays.asList("miss"), stats.missCount());
+
+        mfs.add(new CounterMetricFamily(cacheName+"_cache_eviction_total",
+                help + " cache eviction totals", stats.evictionCount()));
+
+        if(cache instanceof LoadingCache) {
+            CounterMetricFamily cacheLoadTotal = new CounterMetricFamily(cacheName+"_cache_load_total",
+                    help + " cache load totals", Arrays.asList("success"));
+            mfs.add(cacheLoadTotal);
+            cacheLoadTotal.addMetric(Arrays.asList("success"), stats.loadSuccessCount());
+            cacheLoadTotal.addMetric(Arrays.asList("exception"), stats.loadExceptionCount());
+
+            mfs.add(new CounterMetricFamily(cacheName+"_cache_load_sum_seconds",
+                    help + " cache load total time in seconds", stats.totalLoadTime()/ Summary.NANOSECONDS_PER_SECOND));
+        }
+
+        return mfs;
+    }
+}
--- a/simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
+++ b/simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
@@ -56,22 +56,24 @@ public class CacheMetricsCollector extends Collector {
 
         CacheStats stats = cache.stats();
 
-        mfs.add(new CounterMetricFamily(cacheName+"_cache_hit_total",
-                " cache hit totals", stats.hitCount()));
-        mfs.add(new CounterMetricFamily(cacheName+"_cache_miss_total",
-                " cache miss totals", stats.missCount()));
-        mfs.add(new CounterMetricFamily(cacheName+"_cache_eviction_total",
-                " cache eviction totals", stats.evictionCount()));
+        mfs.add(new CounterMetricFamily(cacheName + "_cache_hit_total",
+                cacheName + " cache hit totals", stats.hitCount()));
+        mfs.add(new CounterMetricFamily(cacheName + "_cache_miss_total",
+                cacheName + " cache miss totals", stats.missCount()));
+        mfs.add(new CounterMetricFamily(cacheName + "_cache_requests_total",
+                cacheName + " cache totals (hits and misses)", stats.requestCount()));
+        mfs.add(new CounterMetricFamily(cacheName + "_cache_eviction_total",
+                cacheName + " cache eviction totals", stats.evictionCount()));
 
         if(cache instanceof LoadingCache) {
-            mfs.add(new CounterMetricFamily(cacheName+"_cache_load_success_total",
-                    " cache load totals", stats.loadSuccessCount()));
-            mfs.add(new CounterMetricFamily(cacheName+"_cache_load_error_total",
-                    " cache load totals", stats.loadExceptionCount()));
+            mfs.add(new CounterMetricFamily(cacheName + "_cache_load_success_total",
+                    cacheName + " cache load successes", stats.loadSuccessCount()));
+            mfs.add(new CounterMetricFamily(cacheName + "_cache_load_error_total",
+                    cacheName + " cache load failures", stats.loadExceptionCount()));
 
-            mfs.add(new SummaryMetricFamily(cacheName+"_cache_load_duration_seconds",
-                    " cache load total time in seconds", stats.loadCount(),
-                    stats.totalLoadTime()/ Summary.NANOSECONDS_PER_SECOND));
+            mfs.add(new SummaryMetricFamily(cacheName + "_cache_load_duration_seconds",
+                    cacheName + " cache load total time in seconds", stats.loadCount(),
+                    stats.totalLoadTime() / Collector.NANOSECONDS_PER_SECOND));
         }
 
         return mfs;
--- a/simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
+++ b/simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
@@ -6,9 +6,9 @@ import com.google.common.cache.LoadingCache;
 import io.prometheus.client.Collector;
 import io.prometheus.client.CounterMetricFamily;
 import io.prometheus.client.Summary;
+import io.prometheus.client.SummaryMetricFamily;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 
@@ -19,7 +19,7 @@ import java.util.List;
  *
  * // Note that `recordStats()` is required to gather non-zero statistics
  * Cache<String, String> cache = CacheBuilder.newBuilder().recordStats().build();
- * new CacheMetricsCollector(cache, "myapp_mycache", "Names cache").register();
+ * new CacheMetricsCollector(cache, "myapp_mycache").register();
  *
  * }</pre>
  *
@@ -27,15 +27,15 @@ import java.util.List;
  *
  * With the examle above sample metric names would be:
  * <pre>
- *     myapp_mycache_cache_request_total{found="hit",} 10.0
- *     myapp_mycache_cache_request_total{found="miss",} 3.0
+ *     myapp_mycache_cache_hit_total 10.0
+ *     myapp_mycache_cache_miss_total 3.0
  *     myapp_mycache_cache_eviction_total 1.0
  * </pre>
  *
  * Additionally if the cache includes a loader additional metrics would be provided:
  * <pre>
- *     myapp_mycache_cache_load_total{success="success",} 5.0
- *     myapp_mycache_cache_load_total{success="exception",} 1.0
+ *     myapp_mycache_cache_load_success_total 5.0
+ *     myapp_mycache_cache_load_error_total 1.0
  *     myapp_mycache_cache_load_sum_seconds 0.0034
  * </pre>
  *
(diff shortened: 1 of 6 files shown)


simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+        this.help = help;
+    }
+
+    @Override
+    public List<MetricFamilySamples> collect() {
+        List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
+
+        CacheStats stats = cache.stats();
+
+        CounterMetricFamily cacheRequestTotal = new CounterMetricFamily(cacheName+"_cache_request_total",
```
brian-brazil: requests

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+                help + " cache request totals", Arrays.asList("found"));
```
brian-brazil: Usually the help is static in cases like this

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+        mfs.add(cacheRequestTotal);
+        cacheRequestTotal.addMetric(Arrays.asList("hit"), stats.hitCount());
```
brian-brazil: Separate hit/miss/total metrics are easiest to work with, as you can simply divide to get the number you want.

With labels it's harder as you have to use group_left.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+
+        mfs.add(new CounterMetricFamily(cacheName+"_cache_eviction_total",
+                help + " cache eviction totals", stats.evictionCount()));
+
+        if(cache instanceof LoadingCache) {
+            CounterMetricFamily cacheLoadTotal = new CounterMetricFamily(cacheName+"_cache_load_total",
+                    help + " cache load totals", Arrays.asList("success"));
+            mfs.add(cacheLoadTotal);
+            cacheLoadTotal.addMetric(Arrays.asList("success"), stats.loadSuccessCount());
+            cacheLoadTotal.addMetric(Arrays.asList("exception"), stats.loadExceptionCount());
```
brian-brazil: Total and errors as separate metrics is best.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+
+            mfs.add(new CounterMetricFamily(cacheName+"_cache_load_sum_seconds",
```
brian-brazil: This sounds like we might be able to make this into a summary. Does it include failures?

simpleclient_guava/src/test/java/io/prometheus/client/cache/guava/CacheMetricsCollectorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    public void cacheExposesMetricsForHitMissAndEviction() throws Exception {
+        Cache<String, String> cache = CacheBuilder.newBuilder().maximumSize(2).recordStats().build();
+        CacheMetricsCollector collector = new CacheMetricsCollector(cache, "myapp_users", "Names cache");
+
+        cache.getIfPresent("user1");
+        cache.getIfPresent("user1");
+        cache.put("user1", "First User");
+        cache.getIfPresent("user1");
+
+        //Add to cache to trigger eviction
```
brian-brazil: Space after //, full stop at the end.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+public class CacheMetricsCollector extends Collector {
+
+    private final Cache cache;
+    private final String cacheName;
+    private final String help;
+
+    public CacheMetricsCollector(Cache cache, String cacheName, String help) {
+        this.cache = cache;
+        this.cacheName = cacheName;
+        this.help = help;
```
brian-brazil: There's no need to require the user to provide help

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
checketts: It does include failures. It is incremented both when a load is successful and when it errors out (http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/cache/CacheStats.html)

Great point about the `SummaryMetricFamily`. Perfect fit.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,79 @@
+package io.prometheus.client.cache.guava;
```
brian-brazil: guava is a lot more than a cache, we might want to do guave.cache to allow for future stuff

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public List<MetricFamilySamples> collect() {
+        List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
+
+        CacheStats stats = cache.stats();
+
+        mfs.add(new CounterMetricFamily(cacheName+"_cache_hit_total",
+                " cache hit totals", stats.hitCount()));
```
brian-brazil: Create a total from hits+misses too

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,79 @@
...
+        mfs.add(new CounterMetricFamily(cacheName+"_cache_miss_total",
+                " cache miss totals", stats.missCount()));
+        mfs.add(new CounterMetricFamily(cacheName+"_cache_eviction_total",
+                " cache eviction totals", stats.evictionCount()));
+
+        if(cache instanceof LoadingCache) {
+            mfs.add(new CounterMetricFamily(cacheName+"_cache_load_success_total",
```
brian-brazil: This should be a total

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,79 @@
...
+                    " cache load totals", stats.loadSuccessCount()));
+            mfs.add(new CounterMetricFamily(cacheName+"_cache_load_error_total",
+                    " cache load totals", stats.loadExceptionCount()));
+
+            mfs.add(new SummaryMetricFamily(cacheName+"_cache_load_duration_seconds",
+                    " cache load total time in seconds", stats.loadCount(),
+                    stats.totalLoadTime()/ Summary.NANOSECONDS_PER_SECOND));
```
brian-brazil: space before /

brian-brazil: NANOSECONDS_PER_SECOND is from Collector

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
```diff
@@ -0,0 +1,79 @@
...
+        this.cacheName = cacheName;
+    }
+
+    @Override
+    public List<MetricFamilySamples> collect() {
+        List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
+
+        CacheStats stats = cache.stats();
+
+        mfs.add(new CounterMetricFamily(cacheName+"_cache_hit_total",
```
brian-brazil: space around +

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
checketts: I considered that. I can't find any additional Guava items that might be added (but I'm happy to be proven wrong). 

Also I have `cache.caffeine` ready (I'll submit the PR after this one is done) and `cache.ehcache` soon after that. So it ay be a little weird to have a `guava.cache` that doesn't parallel those other 2.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
checketts: Sorry I'm not sure what you are saying here. The metric name ends with `_total` already

checketts:
Added updates from PR feedback

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
brian-brazil: _total indicates it's a counter. This is currently a measure of successful calls, all calls would be more useful.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
brian-brazil: EventBus is one that stands out, and I can imagine utility classes around the others.

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
checketts: All load calls is already tracked in the `_cache_load_duration_seconds` summary. Are you asking that I remove the `_cache_load_success_total` counter and duplicate the 'total load attempts' metric?

I thought the load success and failure metrics are needed to determine load success/failure ratios.

brian-brazil: Yes. While you can calculate it from the Summary, it's not obvious to everyone that _count can be used for this. There's not many metrics here, so the extra shouldn't be an issue.

Failure ratio is failures/total. With success and failure that is the more complex `f/(s+f)`

checketts: In my last comment I asked if I should remove the success counter, then you mentioned having an 'extra', So I'm understanding that to mean it is OK to keep the success counter and add in a `load total` counter?

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
ben-manes: @checketts For Caffeine, you can either pull metrics from CacheStats (like Guava) or push into metrics using a custom `StatsCollector`. See [stats-metrics](https://github.com/ben-manes/caffeine/tree/master/examples/stats-metrics) for a Dropwizard example of the push approach. Both are equivalent and the choice depends on preferences.

checketts:
Hi @ben-manes ! I stuck with the CacheStats approach. It was
straightforward. The only catch was in my unit test eviction seemed to
happen slightly asyncronously. I'll mention you once this Guava PR is
merged in.

ben-manes:
You can use `Caffeine.exeuctor(Runnable::run)` in tests. By default the maintenance work is deferred to `ForkJoinPool.commonPool()` to minimize user-facing latencies. That setting makes it act like Guava, where we amortized it on the callers, with very little performance impact. That should make tests more predictable. Alternative you can call `cleanUp` which does the work on the caller, forcing it to wait if a task is in progress.

checketts:
OK I think I've addressed all points in this PR. Let me know if I should revisit any.

checketts:
Would a better naming scheme for the metrics be:
```
guava_cache_hit_total{cache="mycache",} 10.0
guava_cache_miss_total{cache="mycache",} 3.0
guava_cache_requests_total{cache="mycache",} 13.0
```

It feels like using the library as a prefix makes sense and the label for grouping will make queries that include all caches would be better

simpleclient_guava/src/main/java/io/prometheus/client/cache/guava/CacheMetricsCollector.java
brian-brazil: Drop the success, have only a load total and a failure counter.

brian-brazil:
That naming scheme sounds good to me

simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+ *     myapp_mycache_cache_hit_total 10.0
+ *     myapp_mycache_cache_miss_total 3.0
+ *     myapp_mycache_cache_eviction_total 1.0
+ * </pre>
+ *
+ * Additionally if the cache includes a loader additional metrics would be provided:
+ * <pre>
+ *     myapp_mycache_cache_load_success_total 5.0
+ *     myapp_mycache_cache_load_error_total 1.0
+ *     myapp_mycache_cache_load_sum_seconds 0.0034
```
brian-brazil: This produces _count and _sum timeseries

simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * // Note that `recordStats()` is required to gather non-zero statistics
+ * Cache<String, String> cache = CacheBuilder.newBuilder().recordStats().build();
+ * new CacheMetricsCollector(cache, "myapp_mycache").register();
+ *
+ * }</pre>
+ *
+ * Exposed metrics are prefixed with the provided prefix.
+ *
+ * With the examle above sample metric names would be:
```
brian-brazil: example

These examples are out of sync with the code.

simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java
```diff
@@ -0,0 +1,136 @@
...
+ *     guava_cache_hit_total{cache="mycache"} 10.0
+ *     guava_cache_miss_total{cache="mycache"} 3.0
+ *     guava_cache_requests_total{cache="mycache"} 3.0
+ *     guava_cache_eviction_total{cache="mycache"} 1.0
+ * </pre>
+ *
+ * Additionally if the cache includes a loader, the following metrics would be provided:
+ * <pre>
+ *     guava_cache_load_failure_total{cache="mycache"} 5.0
+ *     guava_cache_load_duration_seconds_count{cache="mycache"} 1.0
```
brian-brazil: the value of this should be higher than the failure value

simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Collect metrics from Guava's com.google.common.cache.Cache.
+ * <p>
+ * <pre>{@code
+ *
+ * // Note that `recordStats()` is required to gather non-zero statistics
+ * Cache<String, String> cache = CacheBuilder.newBuilder().recordStats().build();
+ * CacheMetricsCollector cacheMetrics = new CacheMetricsCollector(cache, "myapp_mycache").register();
```
brian-brazil: You're missing the constructor to allow this usage

simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java
```diff
@@ -0,0 +1,136 @@
...
+
+        CounterMetricFamily cacheEvictionTotal = new CounterMetricFamily("guava_cache_eviction_total",
+                "Cache eviction totals, doesn't include manually removed entries", labelNames);
+        mfs.add(cacheEvictionTotal);
+
+        CounterMetricFamily cacheLoadFailure = new CounterMetricFamily("guava_cache_load_failure_total",
+                "Cache load failures", labelNames);
+        mfs.add(cacheLoadFailure);
+
+        CounterMetricFamily cacheLoadTotal = new CounterMetricFamily("guava_cache_loads_total",
```
brian-brazil: This isn't in the javadoc

simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * }</pre>
+ *
+ * Exposed metrics are labeled with the provided cache name.
+ *
+ * With the example above, sample metric names would be:
+ * <pre>
+ *     guava_cache_hit_total{cache="mycache"} 10.0
+ *     guava_cache_miss_total{cache="mycache"} 3.0
+ *     guava_cache_requests_total{cache="mycache"} 3.0
```
brian-brazil: 13

brian-brazil:
Thanks!

PR #193, opened 2017-02-01
Add convenience 'get'-methods to nolabel-Counter, Gauge and Summary

pskiwi:
Child classes of `Counter`, `Gauge `and `Summary` have `get()` methods for the current value.
This adds this capability to the base classes to make nolabel-objects able to get the current value.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Summary.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Summary.java
@@ -1,7 +1,6 @@
 package io.prometheus.client;
 
 import io.prometheus.client.CKMSQuantiles.Quantile;
-import io.prometheus.client.Summary.Child.Value;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -314,7 +313,7 @@ public class Summary extends SimpleCollector<Summary.Child> implements Counter.D
    * <p>
    * <em>Warning:</em> The definition of {@link Value} is subject to change.
    */
-  public Value get() {
+  public Child.Value get() {
     return noLabelsChild.get();
   }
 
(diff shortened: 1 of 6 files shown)


brian-brazil:
This is mosttly only used for unittests, and even then unittests of metrics are rare. Is this a common use case for you?

pskiwi:
In some cases I use metric values collected by Prometheus for internal calculations and statistics. It is very handy to directly use the existing instrumentation, otherwise I would have to duplicate the code.
I was wondering because the child-classes already had the getters but not the nolabel versions.

simpleclient/src/main/java/io/prometheus/client/Summary.java
```diff
@@ -1,6 +1,7 @@
 package io.prometheus.client;
 
 import io.prometheus.client.CKMSQuantiles.Quantile;
+import io.prometheus.client.Summary.Child.Value;
```
brian-brazil: Why the import?

pskiwi:
It seems to be useless. I will fix this. But I have no access to my development machine until Monday.

pskiwi:
The import wasn't useless. It doesn't compile this way.  But after qualifying the return value with `Child.Value` it works without import.

pskiwi:
@brian-brazil 
The build fails with the goal org.apache.maven.plugins:maven-javadoc-plugin:2.8:aggregate.
Do you have any idea what is going wrong here?

brian-brazil:
Offhand you're missing a </p>

pskiwi:
Thanks for the response. The cause was the missing qualification of the return type.

brian-brazil:
Thanks!

PR #196, opened 2017-02-05
Add JVM and OS version info to export

rswart:
This PR would allow us to report on JVM and OS versions, e.g. for vulnerability management.

@brian-brazil / @beorn7 Please review of this is PR can be accepted. Happy to make changes if and where needed.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
@@ -26,14 +26,13 @@ public class VersionInfoExports extends Collector {
 
 
     public List<MetricFamilySamples> collect() {
-        String UNKNOWN_LABEL_VALUE = "unknown";
         List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
 
         GaugeMetricFamily jvmInfo = new GaugeMetricFamily(
                 "jvm_info",
-                "jvm version info",
+                "JVM version info",
                 Arrays.asList("version", "vendor"));
-        jvmInfo.addMetric(Arrays.asList(System.getProperty("java.runtime.version", UNKNOWN_LABEL_VALUE), System.getProperty("java.vm.vendor", UNKNOWN_LABEL_VALUE)), 1L);
+        jvmInfo.addMetric(Arrays.asList(System.getProperty("java.runtime.version", "unknown"), System.getProperty("java.vm.vendor", "unknown")), 1L);
         mfs.add(jvmInfo);
 
         return mfs;
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
@@ -8,7 +8,7 @@ import java.util.Arrays;
 import java.util.List;
 
 /**
- * Exports JVM and OS version info.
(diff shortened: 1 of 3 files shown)


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
```diff
@@ -0,0 +1,49 @@
...
+
+        GaugeMetricFamily jvmInfo = new GaugeMetricFamily(
+                "jvm_info",
+                "jvm version info",
+                Arrays.asList("version", "vendor"));
+        jvmInfo.addMetric(Arrays.asList(System.getProperty("java.runtime.version", UNKNOWN_LABEL_VALUE), System.getProperty("java.vm.vendor", UNKNOWN_LABEL_VALUE)), 1L);
+        mfs.add(jvmInfo);
+
+        GaugeMetricFamily osInfo = new GaugeMetricFamily(
+                "os_info",
```
brian-brazil: Operating system info is not the responsibility of an application, that should come from the node exporter.

rswart: I agree. I put it in because I have clients where OS is managed by different department than application/middleware and the OS department is unfortunately still using less modern tools. Using node exporter is not option there.

Nevertheless, if you think it is better to remove it out I'll do that.

brian-brazil: Please remove it.

rswart: done

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
```diff
@@ -0,0 +1,41 @@
...
+ * <pre>
+ *   jvm_info{version="1.8.0_45-b14",vendor="Oracle Corporation"} 1.0
+ * </pre>
+ */
+
+public class VersionInfoExports extends Collector {
+
+
+    public List<MetricFamilySamples> collect() {
+        String UNKNOWN_LABEL_VALUE = "unknown";
```
brian-brazil: Just inline this.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/VersionInfoExports.java
```diff
@@ -0,0 +1,41 @@
...
+        List<MetricFamilySamples> mfs = new ArrayList<MetricFamilySamples>();
+
+        GaugeMetricFamily jvmInfo = new GaugeMetricFamily(
+                "jvm_info",
+                "jvm version info",
```
rswart: Done. Made JVM upper case in help text and added new line at end of file per complaint below.

brian-brazil:
Thanks!

SimenB:
I'm currently trying to create a table with this information (using grafana). Where I'm at right now is `query:sum(jvm_info) by (app, version)`, `legend:{{app}} {{version}}` in a table, resulting in:

image

What I'd like to do is have the label value in the right column though. Any easy way to do that? Or should I just go with a custom template directly in Prometheus?
https://prometheus.io/docs/visualization/consoles/

brian-brazil:
Please don't pollute closed issues with support requests, https://groups.google.com/forum/#!forum/prometheus-users is the appropriate place for that.

PR #198, opened 2017-02-09
Add caffeine metric collector

checketts:
This mimics the Guava cache usage for collecting metrics for https://github.com/ben-manes/caffeine

FYI @ben-manes

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ cacheMetrics.addCache("myCacheLabel", cache);
 ```
 
 The Caffeine equivalent is nearly identical. Again, be certain to call `recordStats()`
- when building the cache sot hat metrics are collected.
+ when building the cache so that metrics are collected.
 
 ```java
 CacheMetricsCollector cacheMetrics = new CacheMetricsCollector().register();
--- a/README.md
+++ b/README.md
@@ -34,6 +34,18 @@ version can be found on in the maven repository for
   <artifactId>simpleclient_pushgateway</artifactId>
   <version>0.0.20</version>
 </dependency>
+<!-- Guava Cache metrics-->
+<dependency>
+  <groupId>io.prometheus</groupId>
+  <artifactId>simpleclient_guava</artifactId>
+  <version>0.0.20</version>
+</dependency>
+<!-- Caffeine Cache metrics-->
+<dependency>
+  <groupId>io.prometheus</groupId>
+  <artifactId>simpleclient_pushgateway</artifactId>
+  <version>0.0.20</version>
+</dependency>
 ```
 
 ### Javadocs
--- a/README.md
+++ b/README.md
@@ -34,18 +34,6 @@ version can be found on in the maven repository for
   <artifactId>simpleclient_pushgateway</artifactId>
   <version>0.0.20</version>
 </dependency>
-<!-- Guava Cache metrics-->
-<dependency>
-  <groupId>io.prometheus</groupId>
-  <artifactId>simpleclient_guava</artifactId>
-  <version>0.0.20</version>
-</dependency>
-<!-- Caffeine Cache metrics-->
-<dependency>
-  <groupId>io.prometheus</groupId>
-  <artifactId>simpleclient_pushgateway</artifactId>
-  <version>0.0.20</version>
-</dependency>
 ```
 
 ### Javadocs
@@ -284,6 +272,28 @@ To register the log4j2 collector at root level:
         </Root>
     </Loggers>
 </Configuration>
+```
+
+#### Caches
+
+To register the Guava cache collector, be certain to add `recordStats()` when building
+the cache and adding it to the registered collector. 
+
+```java
+CacheMetricsCollector cacheMetrics = new CacheMetricsCollector().register();
+
+Cache<String, String> cache = CacheBuilder.newBuilder().recordStats().build();
+cacheMetrics.addCache("myCacheLabel", cache);
+```
+
+The Caffeine equivalent is nearly identical. Again, be certain to call `recordStats()`
+ when building the cache sot hat metrics are collected.
+
+```java
+CacheMetricsCollector cacheMetrics = new CacheMetricsCollector().register();
+
+Cache<String, String> cache = Caffeine.newBuilder().recordStats().build();
+cacheMetrics.addCache("myCacheLabel", cache);
 ```
 
 #### Hibernate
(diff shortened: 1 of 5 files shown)


simpleclient_caffeine/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient</artifactId>
+            <version>0.0.21-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>2.3.5</version>
```
brian-brazil: Use the oldest version you can so we can ensure we're as backwards compatible as possible

checketts:
I was going to add it to the readme after a release has occurred. Since it wouldn't actually be available for version `0.0.20`

```
<dependency>
  <artifactId>simpleclient_pushgateway</artifactId>
  <version>0.0.20</version>
</dependency>
```

I also want to revisit the giant XML and replace it with a single XML snippet as an example and a text description of the various modules available and usages.

brian-brazil:
There'll be a release fairly soon, there's been quite a few modules added.

ben-manes:
Thanks! I don't think you need to support 1.0 anymore. There is also an evictionWeight metric in the 2.x line. As an FYI, the AsycLoadingCache doesn't use the Cache interface due to subtle differences, but has a synchronous() view instead. You might want to mention that in your JavaDoc.

checketts:
@ben-manes I switched to 2.0. Thanks for the heads up on the `AsyncLoadingCache`. Instead of calling it out in the JavaDoc, I added a method that unwraps it for the user. Think that is sufficient?

I also couldn't find the `evictionWeight ` in the `CacheStats` class. Where would I find that? Also does it behave like a gauge to indicate how likely eviction is?

ben-manes:
That does look better.

It looks like `evictionWeight` was added in `2.3.0`. Note that since I'm strictly following semver, you should be okay using latest releases.

checketts:
@ben-manes The problem we will hit if we add in the `evictionWeight` metric is that people using < `2.3` will get a method missing error if they want metrics. 

I'll see if I can still include it.

ben-manes:
Wouldn't the transitive dependency cause their version to upgrade?

checketts:
Yes, if they choose to upgrade. It is common to use `DependencyManagement` to lock in a specific version and minimize changes when adding in libraries.

For example I wouldn't want adding in a new dep like the prometheus client to upgrade my version of Jetty.

checketts:
I've added the CacheEviction weight and it should be backwards compatible too.

@brian-brazil  I didn't like the empty catch. Is it acceptable to add `slf4j` as a dependency for logging?

checketts:
I disabled the AsyncLoading test because it required Java 8 classes. I know the `prometheus_simpleclient` is intended to be compatible with Java 6, but this module is Java 8 specific (since Caffeine is). Would it be acceptable to use Java 8 for the build, and compile for Java 6?

ben-manes:
You may want to use AnimalSniffer if using mixed compilation targets.

README.md
```diff
@@ -34,6 +34,18 @@ version can be found on in the maven repository for
   <artifactId>simpleclient_pushgateway</artifactId>
   <version>0.0.20</version>
 </dependency>
+<!-- Guava Cache metrics-->
```
brian-brazil: Don't expand this list, it's already annoying to maintain and I don't want to scare off users.

simpleclient_caffeine/src/test/java/io/prometheus/client/cache/caffeine/CacheMetricsCollectorTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+        assertMetric(registry, "caffeine_cache_load_failure_total", "loadingusers", 1.0);
+        assertMetric(registry, "caffeine_cache_loads_total", "loadingusers", 3.0);
+
+        assertMetric(registry, "caffeine_cache_load_duration_seconds_count", "loadingusers", 3.0);
+        assertMetricGreatThan(registry, "caffeine_cache_load_duration_seconds_sum", "loadingusers", 0.0);
+    }
+
+
+    // Test disabled since it requires Java 8 features
```
brian-brazil: Remove rather than commenting out.

simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
```diff
@@ -0,0 +1,159 @@
...
+        CounterMetricFamily cacheRequestsTotal = new CounterMetricFamily("caffeine_cache_requests_total",
+                "Cache request totals, hits + misses", labelNames);
+        mfs.add(cacheRequestsTotal);
+
+        CounterMetricFamily cacheEvictionTotal = new CounterMetricFamily("caffeine_cache_eviction_total",
+                "Cache eviction totals, doesn't include manually removed entries", labelNames);
+        mfs.add(cacheEvictionTotal);
+
+        GaugeMetricFamily cacheEvictionWeight = new GaugeMetricFamily("caffeine_cache_eviction_weight",
+                "Cache eviction weight", labelNames);
```
brian-brazil: Some more flavour would be good

simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
```diff
@@ -0,0 +1,159 @@
...
+        mfs.add(cacheEvictionTotal);
```
brian-brazil: cacheEvictionWeight

simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
```diff
@@ -0,0 +1,159 @@
...
+        mfs.add(cacheLoadSummary);
+
+        for(Map.Entry<String, Cache> c: children.entrySet()) {
+            List<String> cacheName = Arrays.asList(c.getKey());
+            CacheStats stats = c.getValue().stats();
+
+            try{
+                cacheEvictionWeight.addMetric(cacheName, stats.evictionWeight());
+            } catch (Exception e) {
+                // EvictionWeight metric is unavailable, newer version of Caffeine is needed
```
brian-brazil: fullstop at the end of comment.

checketts:
I'm glad I misunderstood when you said 'it should be in the readme'. Since the comment was on the POM xaml I assumed you meant adding to the maven block. I agree that XML doesn't need to grow any more.

README.md
```diff
@@ -274,6 +274,28 @@ To register the log4j2 collector at root level:
...
+
+```java
+CacheMetricsCollector cacheMetrics = new CacheMetricsCollector().register();
+
+Cache<String, String> cache = CacheBuilder.newBuilder().recordStats().build();
+cacheMetrics.addCache("myCacheLabel", cache);
+```
+
+The Caffeine equivalent is nearly identical. Again, be certain to call `recordStats()`
+ when building the cache sot hat metrics are collected.
```
brian-brazil: so that

brian-brazil:
Thanks!

PR #199, opened 2017-02-14
Remove unnecessary allocation in TextFormat

pkubowicz:
Further improve work from 5760fe1 - avoid allocating
StringBuilder with double size and allocating a string
from it.
Avoid reading the same character twice.
Add accented characters the the test.

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -9,6 +9,11 @@ import io.prometheus.client.Collector;
 
 public class TextFormat {
   /**
+   * Content-type for text version 0.0.4.
+   */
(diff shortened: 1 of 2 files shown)


pjfanning:
While the changes look like they should be faster, they appear to be slower in simple performance test: https://github.com/pjfanning/prometheus-client-test
I'm not really sure why.

pkubowicz:
@pjfanning I like your test, I only think it was too simple. See https://github.com/pkubowicz/prometheus-client-test/commit/2b771005d6a30e4532e3fb0c04e86a4021f7a6f3

I changed data set, changed how the test warms up and it shows that my change allocates 21% less memory and is 8% faster. I tried Parallel, MarkSweep and G1 garbage collectors and memory results are very consistent. There is more variation in execution time, but it's in favour of avoiding StringBuffer.

pjfanning:
@pkubowicz Your performance test looks good. I agree my variant was overly simplistic - I was just surprised when it was returning very consistent results.

brian-brazil:
Thanks!

PR #200, opened 2017-02-15
Improve array allocation

pkubowicz:
- precompute array size when it's obvious - avoids re-allocating
- use singletonList() when the array can be immutable
- replace some empty arrays with emptyList()

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
@@ -1,5 +1,6 @@
 package io.prometheus.client;
 
+import java.util.ArrayList;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.Arrays;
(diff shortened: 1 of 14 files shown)


simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -267,7 +267,7 @@ public Item(double value, int lower_delta, int delta) {
 
         @Override
         public String toString() {
-            return String.format("%d, %d, %d", value, g, delta);
+            return String.format("%f, %d, %d", value, g, delta);
```
brian-brazil: We should only be pulling down upstream changes here.

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -142,6 +143,13 @@ protected void initializeNoLabelsChild() {
    */
   protected abstract Child newChild();
 
+  List<MetricFamilySamples> familySamplesList(Collector.Type type, List<MetricFamilySamples.Sample> samples) {
```
brian-brazil: This should be protected

pkubowicz: Is it ok now?

brian-brazil:
Thanks!

PR #202, opened 2017-02-17
Added support for Prometheus labels in DropwizardExports

tengvig:
Added support for Prometheus labels in DropwizardExports.
All metrics with names like 
`amount_currency="USD"
`
Will be exported the metric with name=amount and currency as a label, e.g.
`amount{currency="USD"}
`


brian-brazil:
Thanks for the PR, this is not the convention for labels in Dropwizard. See the statsd/graphite exporters for what's required.

tengvig:
I wasn't aware of the convention for labels. I've looked into it now. Unfortunately, creating a configuration language is beyond the scope for me at the moment.

Issue #203, opened 2017-02-19
Roadmap for 1.0


checketts:
I'm finding the Prometheus Java Client to be very useful and adding many custom collectors as needed.  Is there a roadmap or list of ideas that need attention?

The client already feels pretty robust and it is a little surprising that the version number is `0.0.20` right now. 

Similar to Prometheus' overall roadmap: https://prometheus.io/docs/introduction/roadmap/ are there any particular parts of the Prometheus Java Client that I could work on, review, or submit PRs?

Items that have my interest and that I am aware of:

- Improving Spring boot integration (making it integrate fully with Spring boot Actuator, likely by working with upstream)
- Instrumentation for Threadpools/Executors
- Ehcache/Redis/Memcached client instrumentation
- RabbitMQ instrumentation (likely with upstream)

These are all usecases that apply to applications I'm using. 

Are there any higher priority items or overall roadmap efforts that I could contribute to?

brian-brazil:
The main thing from my standpoint is finishing off the ?name[]= support, and possibly adding a parser.

Integrations are handy, but don't affect the core.

checketts:
By `?name[]=` support, do you mean for the endpoint `\metrics?name[]=` the ability to filter requested metrics as a parameter for that endpoint? I would be happy to start an issue and matching PR for the servlet/spring/vert.x endpoints.

I suppose a ProtoBuf output option would also be needed?

I don't know much about parsers though so I won't be too much help there.

brian-brazil:
> the ability to filter requested metrics as a parameter for that endpoint?

Yes. See how it's done in Python.

> I suppose a ProtoBuf output option would also be needed? (As a separate issue)

That's not required.

> I don't know much about parsers though so I won't be too much help there.

It's a matter of porting the Python one.

brian-brazil:
At this point I think there's largely just some cleanup to be done, as post 1.0 we shouldn't be doing breaking changes. For example DropwizardExports should be DropwizardCollector.

brian-brazil:
I'm not explicitly tracking a 1.0 now, and there's still OpenMetrics at least to be sorted.

PR #204, opened 2017-02-20
Add support for `name[]` parameter filtering in CollectorRegistry and MetricsServlet

checketts:
As mentioned in #203


simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -125,31 +125,53 @@ public void clear() {
...
-        next = metricFamilySamples.next();
+        next = filter(metricFamilySamples.next());
       } else {
         while (collectorIter.hasNext()) {
           metricFamilySamples = collectorIter.next().collect().iterator();
           if (metricFamilySamples.hasNext()) {
-            next = metricFamilySamples.next();
+            next = filter(metricFamilySamples.next());
+            if(next == null) {
+              continue;
```
brian-brazil: This will skip the rest of the collector the first time it comes to a non-matching sample.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -125,31 +125,53 @@ public void clear() {
...
+      this(Collections.<String>emptySet());
+    }
+
     private void findNextElement() {
       if (metricFamilySamples != null && metricFamilySamples.hasNext()) {
-        next = metricFamilySamples.next();
+        next = filter(metricFamilySamples.next());
       } else {
         while (collectorIter.hasNext()) {
           metricFamilySamples = collectorIter.next().collect().iterator();
```
brian-brazil: This is invoking all collectors. An important part of this is the performance, it should only call the collectors it has to.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -125,31 +125,53 @@ public void clear() {
   public Enumeration<Collector.MetricFamilySamples> metricFamilySamples() {
     return new MetricFamilySamplesEnumeration();
   }
+
+  public Enumeration<Collector.MetricFamilySamples> metricFamilySamples(Set<String> includedNames) {
```
brian-brazil: This is going to be experimental pending further development of the HTTP API (e.g. will we need regexes?), so should probably get a different name for now.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
checketts: So I guess we could leverage the `Describable` interface to determine if we can skip the collector? If the collector doesn't implement `Describable` we can just assume it need to be called and filter it after that.

Or can we add to the collector signature so collectors can be aware of filtering? This seems like a potentially more performant approach, but then it puts the weight on individual collectors and that seems like it is leaking the logic.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
checketts: Good point. I'll update the unit test to expose that usecase.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
brian-brazil: The data needed is already built up in the collector registry.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
checketts: I'm not sure what you mean. Are you referring to Prometheus server development?

Are you saying the we might need regex matching for the `names[]` parameters instead of just the simple `contains` check I went with, or should I hold off on that? For now I was just mimicking the Python implementation (though I confess I'm not very familiar with Python)

Do you want me to rename the method to `filteredMetricFamilySamples`?

brian-brazil: > Are you referring to Prometheus server development?

No, this is all in the client libraries. This is a very new feature, and it's not fully clear how it'll develop.

> Are you saying the we might need regex matching for the names[] parameters instead of just the simple contains check I went with, or should I hold off on that?

We don't know yet, it'll be a different parameter if it happens.

> Do you want me to rename the method to filteredMetricFamilySamples?

Sounds good.

checketts:
Should I add to the readme discussing the `name[]` parameter and how it can be used to proactively filter metrics by setting a corresponding `metrics_path` and that it currently only supports full metric name matches?

Or should I keep iterating on it until it is equivalent to the match options used for federation? https://prometheus.io/docs/operating/federation/#configuring-federation 

Perhaps that is what you meant by adding a parser?

brian-brazil:
The docs for this will go on prometheus.io once it's more common among client libraries. You'd use params for this.

> Perhaps that is what you meant by adding a parser?

https://github.com/prometheus/client_python#parser

brian-brazil:
Are you still planning on working on this?

checketts:
Sorry I had assumed it was completed. I'll rebase against master.

brian-brazil:
To be honest I've lost track of where this review is.

checketts:
No problem. I'll let you know if it is complete after I rebase.

simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
```diff
@@ -37,13 +41,22 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
...
-      TextFormat.write004(writer, registry.metricFamilySamples());
+      TextFormat.write004(writer, registry.filteredMetricFamilySamples(parse(req)));
       writer.flush();
     } finally {
       writer.close();
     }
   }
 
+  private Set<String> parse(HttpServletRequest req) {
+    String includedParam = req.getParameter("names[]");
```
brian-brazil: You should use getParameterValues, as multiple params can be provided.

simpleclient_servlet/src/test/java/io/prometheus/client/exporter/MetricsServletTest.java
```diff
@@ -8,21 +8,50 @@
...
+
   @Test
   public void testWriterIsClosedNormally() throws IOException, ServletException {
     HttpServletRequest req = mock(HttpServletRequest.class);
     HttpServletResponse resp = mock(HttpServletResponse.class);
     PrintWriter writer = mock(PrintWriter.class);
     when(resp.getWriter()).thenReturn(writer);
-    new MetricsServlet().doGet(req, resp);
+    CollectorRegistry registry = new CollectorRegistry();
+    Gauge a = Gauge.build("a","a help").register(registry);
```
brian-brazil: space after ,

simpleclient_servlet/src/test/java/io/prometheus/client/exporter/MetricsServletTest.java
```diff
@@ -8,21 +8,50 @@
...
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.StringWriter;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import io.prometheus.client.Gauge;
+import static org.assertj.core.api.Assertions.*;
```
brian-brazil: Explicit imports please

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusEndpointConfiguration.java
```diff
@@ -1,14 +1,28 @@
 package io.prometheus.client.spring.boot;
 
 import io.prometheus.client.CollectorRegistry;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
```
brian-brazil: This is from a different change I think

simpleclient_spring_boot/src/test/java/io/prometheus/client/spring/boot/PrometheusEndpointTest.java
```diff
@@ -36,15 +36,17 @@
   @Test
   public void testMetricsExportedThroughPrometheusEndpoint() {
     // given:
-    final Counter promCounter = Counter.build()
-        .name("foo_bar")
-        .help("a simple prometheus counter")
+    final Counter promCounter = Counter.build("foo_bar","test counter" )
```
brian-brazil: spacing is off here

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -125,28 +125,73 @@ public void clear() {
...
+      this.includedNames = includedNames;
+      collectorIter = includedCollectorIterator(includedNames);
       findNextElement();
     }
 
-    private void findNextElement() {
-      if (metricFamilySamples != null && metricFamilySamples.hasNext()) {
-        next = metricFamilySamples.next();
+    private Iterator<Collector> includedCollectorIterator(Set<String> includedNames) {
+      if(includedNames.isEmpty()) {
```
brian-brazil: space after if

checketts:
Changes have been incorporated.

simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
```diff
@@ -49,12 +49,14 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
...
-    String includedParam = req.getParameter("names[]");
-    if(includedParam == null) {
-      return Collections.emptySet();
-    } else {
-      return new HashSet<String>(Arrays.asList(includedParam.split(",")));
+    String[] includedParam = req.getParameterValues("names[]");
+    HashSet<String> names = new HashSet<String>();
+    if(includedParam != null) {
+      for(String nameList : includedParam) {
+        names.addAll(Arrays.asList(nameList.split(",")));
```
brian-brazil: There's no need for the split, it's a name per param

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -125,28 +125,73 @@ public void clear() {
...
-    private void findNextElement() {
-      if (metricFamilySamples != null && metricFamilySamples.hasNext()) {
-        next = metricFamilySamples.next();
+    private Iterator<Collector> includedCollectorIterator(Set<String> includedNames) {
+      if (includedNames.isEmpty()) {
+        return collectors().iterator();
       } else {
+        HashSet<Collector> collectors = new HashSet<Collector>();
+        synchronized (namesToCollectors) {
+          for(Map.Entry<String,Collector> entry :namesToCollectors.entrySet()){
```
brian-brazil: Formatting is still off, I'd suggest running something against all the files touched.

simpleclient_vertx/src/test/java/io/prometheus/client/exporter/ExampleExporter.java
```diff
@@ -1,33 +0,0 @@
-package io.prometheus.client.exporter;
```
brian-brazil: This shouldn't be deleted.

checketts:
Wow I had thought I was mimicking the Python client. I see I had been mistaken. The is fixed now.  I had held off on the code reformat because it would indent incorrectly. That is fixed now and all the indentation should be clean now.

simpleclient_servlet/src/test/java/io/prometheus/client/filter/MetricsFilterTest.java
```diff
@@ -148,14 +148,14 @@ public Void answer(InvocationOnMock invocationOnMock) throws Throwable {
         f.doFilter(req, res, c);
 
         final Double sum = CollectorRegistry.defaultRegistry.getSampleValue("foo_sum", new String[]{"path", "method"}, new String[]{"/foo", HttpMethods.POST});
-        assertEquals(0.1, sum, 0.001);
+        assertEquals(0.1, sum, 0.01);
```
checketts: This test was flakey on my machine so I let a larger delta

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
```diff
@@ -1,25 +0,0 @@
-package io.prometheus.client.spring.boot;
```
brian-brazil: Why is this being deleted?

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMvcEndpoint.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import java.util.Set;
+
+import static org.springframework.http.HttpHeaders.CONTENT_TYPE;
+
+@ConfigurationProperties("endpoints.prometheus")
+public class PrometheusMvcEndpoint extends EndpointMvcAdapter {
```
brian-brazil: What's the impact of renaming this?

checketts: The impact should be minimal since users don't rely on this class explicitly; it should only be referenced by the `PrometheusEndpointConfiguration` class in the client.

brian-brazil:
Thanks!

PR #212, opened 2017-03-15
Last metric label has redundant comma

tomasd:
Prometheus was unable to parse format because of redundant comma (",") in the end of the labels list


brian-brazil:
The comma is valid.

PR #213, opened 2017-03-16
Add support for RedHat OpenShift HOSA

nicolinux72:
My extention aim to address the integration between prometheus metrics and RedHat Openshift platform. See this article for reference:

https://www.hawkular.org/blog/2017/01/17/obst-hosa.html

The short story is that HOSA, in the request where it collects application metrics, adds an HTTP header like this:
```
accept: application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.7,text/plain;version=0.0.4;q=0.3
```
This prevents your endpoint to answer correctly and HOSA receive instead a error: HTTP 404.

My modification adds a second endpoint at /prometheus/metrics witch produce content compatible with the HOSA request.

Bye
Nicola Santi

### DIFF
--- a/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
+++ b/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
@@ -1,0 +1,33 @@
+package io.prometheus.client.spring.boot;
+
+import org.springframework.boot.actuate.endpoint.mvc.EndpointMvcAdapter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@ConfigurationProperties("endpoints.prometheus")
+public class PrometheusMVCEndpoint extends EndpointMvcAdapter {
+
+	private final PrometheusEndpoint delegate;
+	
+	public PrometheusMVCEndpoint(PrometheusEndpoint delegate) {
+		super(delegate);
+		this.delegate = delegate;
+	}
+	
+	@ResponseBody
+	@RequestMapping(value="/metrics", produces={"text/plain;version=0.0.4"})
+	public String metrics() {
+		if (!this.delegate.isEnabled()) {
+			// Shouldn't happen - MVC endpoint shouldn't be registered when delegate's
+			// disabled
+			return getDisabledResponse().toString();
+		}
+		
+		ResponseEntity<String> result = this.delegate.invoke();
+		
+		return result.getBody();
+	}
+
+}
--- a/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
+++ b/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
@@ -6,6 +6,8 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.ResponseBody;
 
+import io.prometheus.client.exporter.common.TextFormat;
+
 @ConfigurationProperties("endpoints.prometheus")
 public class PrometheusMVCEndpoint extends EndpointMvcAdapter {
 
@@ -16,18 +18,19 @@ public class PrometheusMVCEndpoint extends EndpointMvcAdapter {
 		this.delegate = delegate;
 	}
 	
+	@SuppressWarnings("unchecked")
 	@ResponseBody
-	@RequestMapping(value="/metrics", produces={"text/plain;version=0.0.4"})
-	public String metrics() {
+	@RequestMapping(value="/metrics", produces={TextFormat.CONTENT_TYPE_004})
+	public ResponseEntity<String> metrics() {
 		if (!this.delegate.isEnabled()) {
 			// Shouldn't happen - MVC endpoint shouldn't be registered when delegate's
 			// disabled
-			return getDisabledResponse().toString();
+			return (ResponseEntity<String>) getDisabledResponse();
 		}
 		
 		ResponseEntity<String> result = this.delegate.invoke();
 		
-		return result.getBody();
+		return result;
 	}
 
 }
--- a/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
+++ b/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
@@ -20,7 +20,7 @@ public class PrometheusMVCEndpoint extends EndpointMvcAdapter {
 	
 	@SuppressWarnings("unchecked")
 	@ResponseBody
-	@RequestMapping(value="/metrics", produces={TextFormat.CONTENT_TYPE_004})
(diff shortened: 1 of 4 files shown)


brian-brazil:
That's not the standard accept header. What makes this a bug in the java client?

nicolinux72:
I don't think this is a bug of your java client but I can't assert that firmly: I don't know promethus protocol too much. Perhaps RedHat guy goes right.

For sure, actually, your library can't be integrate inside Red Had Openshift v3.

You could consider this pull merge a facility to integrate with that common paas. Or reject it, obviously :-)

See you and excuse me for my poor english
N.

marcorosi:
+1 for this

brian-brazil:
The java client works fine with Prometheus. If it doesn't work with HOSA, that indicates a bug in HOSA to me.

nicolinux72:
It appears that HOSA is not the only Prometheus client with that  Accept header, as you could discovery simply googling:

https://www.google.it/search?q=application%2Fvnd.google.protobuf%3Bproto%3Dio.prometheus.client.MetricFamily&oq=application%2Fvnd.google.protobuf%3Bproto%3Dio.prometheus.client.MetricFamily&aqs=chrome..69i57j69i58&sourceid=chrome&ie=UTF-8

Please note the first results are important, in my opinion: for example the prometheus/client_golang projects.

c u

brian-brazil:
I don't see what unittests for one particular implementation have to do with this. That is not what Prometheus sends.

Please indicate exactly where you think the bug is.

nicolinux72:
Hi Brian,

the bug is simply that your library reply with a 404 http error in response to a  **correct** request that specify the Accept header in this way:
```
accept: application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.7,text/plain;version=0.0.4;q=0.3
```
please note that official references from Prometheus web site confirm the   correctness of the above header.

https://prometheus.io/docs/instrumenting/exposition_formats/

Read the table that specifies _Format variants comparison_ expecially in relation to topic **Content-type** 

bye

marcorosi:
hi, any news about this? could be very useful for me

brian-brazil:
I haven't had time to investigate what's actually going on here. If there is a bug here, this that this PR requires a new endpoint doesn't feel right.

checketts:
It isn't creating a new endpoint but rather finishing the Spring Actuator support. The existing `PrometheusEndpoint` happens to have a default serialization. The MVC one just makes it an explicit web endpoint.

I actually have an equivalent change on my own implementation that I'd been meaning to submit upstream.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+		this.delegate = delegate;
+	}
+	
+	@ResponseBody
+	@RequestMapping(value="/metrics", produces={"text/plain;version=0.0.4"})
+	public String metrics() {
+		if (!this.delegate.isEnabled()) {
+			// Shouldn't happen - MVC endpoint shouldn't be registered when delegate's
+			// disabled
+			return getDisabledResponse().toString();
```
checketts: Change the return value to `Object` and remove the `toString` (otherwise you lose the 404 status code)

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+		}
+		
+		ResponseEntity<String> result = this.delegate.invoke();
+		
+		return result.getBody();
```
checketts: Return the `ResponseBody` instead of the `String`

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+
+	private final PrometheusEndpoint delegate;
+	
+	public PrometheusMVCEndpoint(PrometheusEndpoint delegate) {
+		super(delegate);
+		this.delegate = delegate;
+	}
+	
+	@ResponseBody
+	@RequestMapping(value="/metrics", produces={"text/plain;version=0.0.4"})
```
hsdmiranda: I would use TextFormat.CONTENT_TYPE_004 on the producer and I preferable the URL should be just /. Why create a new endpoint?

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
nicolinux72: Done. Thank you

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
nicolinux72: Done. I've added a new endpoint beacuse I wasn't able to add produce=TextFormat.CONTENT_TYPE_004 information to the AbstractEndpoint, presented yet in the project.

bye and thanks

nicolinux72: Ok, I got it: now I've extend the original endpoint to accept TextFormat.CONTENT_TYPE_004 also.

Thank you for your help.

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMVCEndpoint.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.boot.actuate.endpoint.mvc.EndpointMvcAdapter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import io.prometheus.client.exporter.common.TextFormat;
+
+@ConfigurationProperties("endpoints.prometheus")
+public class PrometheusMVCEndpoint extends EndpointMvcAdapter {
```
checketts: @nicolinux72 You may have noticed that this PR is addressing the same thing as #225 

I like how #225  implemented the `PrometheusMvcEndpoint` better by `extends AbstractEndpointMvcAdapter<PrometheusEndpoint> ` and the `invoke` implmentation (since it mirrors the HealthMvcEndpoint class in Spring).

simpleclient_spring_boot/src/test/java/io/prometheus/client/spring/boot/PrometheusEndpointTest.java
```diff
@@ -54,6 +58,35 @@ public void testMetricsExportedThroughPrometheusEndpoint() {
...
+    final Counter promCounter = Counter.build()
+        .name("foo_bar2")
+        .help("a simple prometheus counter")
+        .labelNames("label1", "label2")
+        .register();
+
+    // when:
+    promCounter.labels("val1", "val2").inc(3);
+   
+    final String xx = "application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.7,text/plain;version=0.0.4;q=0.3";
```
checketts: Please name variable better, perhaps explaining why it is needed or used for.

brian-brazil:
#225 handled this.

PR #214, opened 2017-03-17
TextFormat - iteration improvement

iobestar:
@brian-brazil - changed metricFamilySamples iteration to use enumeration. This is better because metrics can be streamed directly from source.

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -2,7 +2,6 @@ package io.prometheus.client.exporter.common;
 
 import java.io.IOException;
 import java.io.Writer;
-import java.util.Collections;
 import java.util.Enumeration;
 
 import io.prometheus.client.Collector;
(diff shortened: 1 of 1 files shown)


brian-brazil:
Do you have evidence that this is better?

iobestar:
Yes I have, the java.util.Collections#list implementation speaks for itself:
```
public static <T> ArrayList<T> list(Enumeration<T> e) {
    ArrayList<T> l = new ArrayList<>();
    while (e.hasMoreElements())
        l.add(e.nextElement());
    return l;
}
```
This code creates new ArrayList (with default capacity!) and adds all samples to that list. This is bad because we load all at once + adding to array list is slow because of expanding. Using Enumeration is great because we can implement streaming (for example if we read metrics form file).

brian-brazil:
Do you have a benchmark?

iobestar:
The quick one:
```
public class EnumerationBenchmark {

        final int size = 10000000;
        System.out.println("Collections[ms]: " + doWithCollections(new TestEnumeration(size)));
        System.out.println("Enumeration[ms]: " + doWithEnumerations(new TestEnumeration(size)));
    }

    private static long doWithEnumerations(Enumeration<Integer> enumeration) {

        StopWatch stopWatch = new StopWatch();
        stopWatch.start();
        while (enumeration.hasMoreElements()) {
            enumeration.nextElement();
            // DO SOME WORK
        }
        stopWatch.stop();
        return stopWatch.getTime();
    }

    private static long doWithCollections(Enumeration<Integer> enumeration) {

        StopWatch stopWatch = new StopWatch();
        stopWatch.start();
        for (Integer i : Collections.list(enumeration)) {
            // DO SOME WORK
        }
        stopWatch.stop();
        return stopWatch.getTime();
    }

    private static class TestEnumeration implements Enumeration<Integer> {

        private final int size;
        private int current;

        private TestEnumeration(int size) {
            this.size = size;
        }

        @Override
        public boolean hasMoreElements() {
            return current < size;
        }

        @Override
        public Integer nextElement() {
            return current++;
        }
    }
}
```

Results(2,2 GHz Intel Core i7, 16 GB 1600 MHz DDR3, Java 8):
Collections[ms]: 2314
Enumeration[ms]: 31

brian-brazil:
Thanks!

pjfanning:
https://github.com/pkubowicz/prometheus-client-test is a useful benchmarking tool for the TextFormat code

Issue #211, opened 2017-03-10
guava cache metric collector should also export cache size


fedork:
cache size is useful metric for cache, so it would make sense to add it to Guava cache stats collector

cemo:
This can be closed.

PR #217, opened 2017-03-20
Add cache size metrics to guave and caffeine

checketts:
Addresses #211

### DIFF
--- a/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
+++ b/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
@@ -37,6 +37,7 @@ import java.util.concurrent.ConcurrentMap;
  *     caffeine_cache_miss_total{cache="mycache"} 3.0
  *     caffeine_cache_requests_total{cache="mycache"} 13.0
  *     caffeine_cache_eviction_total{cache="mycache"} 1.0
+ *     caffeine_cache_estimated_size{cache="mycache"} 5.0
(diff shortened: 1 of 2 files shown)


checketts:
@stuartnelson3 I had to increase the MetricsFilter delta to get the tests to pass:

`java.lang.AssertionError: expected:<0.1> but was:<0.104399255>`

It was only allowing 0.001 difference. I let it allow 0.01. Let me know if that is too lenient and if there is a different, preferred approach.

simpleclient_servlet/src/test/java/io/prometheus/client/filter/MetricsFilterTest.java
```diff
@@ -117,7 +117,7 @@ public Void answer(InvocationOnMock invocationOnMock) throws Throwable {
 
         final Double sum = CollectorRegistry.defaultRegistry.getSampleValue("foobar_baz_filter_duration_seconds_sum", new String[]{"path", "method"}, new String[]{path, HttpMethods.POST});
         assertNotNull(sum);
-        assertEquals(0.1, sum, 0.001);
+        assertEquals(0.1, sum, 0.01);
```
brian-brazil: Can you remove the unrelated change?

checketts: Without this change the test was flaky and failed for me locally and failed in TravisCI based on system load.

I'll remove it from the PR, but I suspect it is needed in the long term.

checketts:
OK it is removed now.

brian-brazil:
Thanks!

Issue #222, opened 2017-04-18
Java client incorrectly flushes metric labels to pushgateway when -Dfile.encoding != UTF-8


fedotxxl:
1. Start application with -Dfile.encoding=cp1251
2. Created counter with label 'Привет мир!'
3. Check :9091/metrics

Looks like the problem is https://github.com/prometheus/client_java/blob/master/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java#L257 :

```
new BufferedWriter(new OutputStreamWriter(connection.getOutputStream()))
```

I think encoding should be forced to UTF-8

brian-brazil:
This should probably be fixed down in TextFormat. Would you like to send a PR?

PR #223, opened 2017-04-18
#222

fedotxxl:
fixes https://github.com/prometheus/client_java/issues/222 (checked manually)

### DIFF
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -254,7 +254,7 @@ public class PushGateway {
 
     try {
       if (!method.equals("DELETE")) {
-        BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(connection.getOutputStream()));
+        BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(connection.getOutputStream(), "UTF-8"));
         TextFormat.write004(writer, registry.metricFamilySamples());
         writer.flush();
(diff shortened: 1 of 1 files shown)


brian-brazil:
This is one place of several where this code is called, are all the others correct?

fedotxxl:
how can I check it?

brian-brazil:
Look at everywhere TextFormat.write004 is called.

fedotxxl:
I didn't find any other place to fix

brian-brazil:
Even https://github.com/prometheus/client_java/blob/master/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java#L40 ?

It seems odd to me that this would only be broken in one place.

fedotxxl:
Sorry, I can't help here - I'm not a project maintainer. I can help to fix a problem after it will be reproduced. So I can try to check something. Please tell me what should be checked and I'll try to check it manually

brian-brazil:
If you pass that flag and bring up a simple servlet to serve metrics, does it reproduce the problem you say?

pjfanning:
http://docs.oracle.com/javaee/6/api/javax/servlet/ServletResponse.html#setCharacterEncoding(java.lang.String) seems to indicate that MetricsServlet should be ok- since the Prometheus content type has charset=utf-8.

pjfanning:
@fedotxxl you might be able to workaround the issues by setting `-Dfile.encoding=utf-8` in your batch script (that runs the java command).

fedotxxl:
@pjfanning yep, already fixed this way

brian-brazil:
Thanks!

PR #229, opened 2017-04-28
Use foreach loop where possible

Jubobs:
Addresses #228.


brian-brazil:
This code is copied from upstream, we don't change it.

Jubobs:
@brian-brazil What do you mean by "upstream" in this context?

brian-brazil:
Actually TimeWindowQuantiles is ours so we can change that, I was thinking of its dependency.

See the file themselves for where we got the other two.

Jubobs:
Gotcha. I've amended my PR accordingly.

brian-brazil:
Thanks!

PR #230, opened 2017-04-29
Add dokumentation on how to use `MetricServlet` with Google Guice DI.

goettl79:
Adds documentation on how to use the prom java client in an Google Guice Dependency Injection Environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,7 +501,13 @@ server.setHandler(context);
 context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
 ```
 
+To expose the metrics of MetricsServlet in Google Guice you would need 
+to add the following lines to your ServletModule: 
 
+```java
+bind(MetricsServlet.class).in(Scopes.SINGLETON);
+serve("/metrics").with(MetricsServlet.class);
+```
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
There's many many frameworks out there that each have their own way of dealing with a standard Servlet. I think it's best to presume that a user knows how this would work in their favourite framework, so there's no need to clutter the docs with multiple fragments of solutions.

If you want to share this information, I'd suggest a blog post or similar that can go into more depth than reference documentation like this.

goettl79:
Hi Brian,

Understood. Just thought it might help.

Probably something that you could think about is to annotate the
MetricsServlet with the jsr-330 annotation "Singleton". This would make
integration simpler with guice and spring an might be reasonable too.

Regards, Georg


Brian Brazil <notifications@github.com> schrieb am Sa., 29. Apr. 2017,
16:22:

brian-brazil:
The MetricsServlet is not a singleton, you can have as many of them as you like (though usually only one).

checketts:
Another quick way to share your info would be to post a question on StackOverflow that you answer yourself.

PR #231, opened 2017-05-03
Add simpleclient_executor

cenkakin:
Support for instrumented executor and scheduled executor services

### DIFF
--- a/simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
+++ b/simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
@@ -1,0 +1,204 @@
+package io.prometheus.client.executor;
+
+import io.prometheus.client.Counter;
+import io.prometheus.client.Gauge;
+import io.prometheus.client.Histogram;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class InstrumentedExecutorService implements ExecutorService {
+
+  private static final AtomicLong NAME_COUNTER = new AtomicLong();
+  private static final Counter ST_COUNTER;
+  private static final Counter CT_COUNTER;
+  private static final Gauge R_GAUGE;
+  private static final Histogram D_HISTOGRAM;
+
+  private final ExecutorService delegate;
+  private final Counter.Child submittedCounter;
+  private final Counter.Child completedCounter;
+  private final Gauge.Child runningGauge;
+  private final Histogram.Child durationHistogram;
+
+  static {
+    ST_COUNTER = Counter.build()
+        .name("instrumented_executor_service_submitted_count")
+        .labelNames("serviceName")
+        .help("Instrumented executor service submitted total count")
+        .register();
+    CT_COUNTER = Counter.build()
+        .name("instrumented_executor_service_completed_count")
+        .labelNames("serviceName")
+        .help("Instrumented executor service completed total count")
+        .register();
+
+    R_GAUGE = Gauge.build()
+        .name("instrumented_executor_service_running_count")
+        .labelNames("serviceName")
+        .help("Instrumented executor service running total count")
+        .register();
+
+    D_HISTOGRAM = Histogram.build()
+        .name("instrumented_executor_service_duration_seconds")
+        .labelNames("serviceName")
+        .help("Instrumented executor service duration (s)")
+        .register();
+  }
+
+
+  public InstrumentedExecutorService(ExecutorService delegate) {
+    this(delegate, "instrumented_delegate_" + NAME_COUNTER.incrementAndGet());
+  }
+
+  public InstrumentedExecutorService(ExecutorService delegate, String name) {
+    this.delegate = delegate;
+    this.submittedCounter = ST_COUNTER.labels(name);
+    this.completedCounter = CT_COUNTER.labels(name);
+    this.runningGauge = R_GAUGE.labels(name);
+    this.durationHistogram = D_HISTOGRAM.labels(name);
+  }
+
+  @Override
+  public void shutdown() {
+    delegate.shutdown();
+  }
+
+  @Override
+  public List<Runnable> shutdownNow() {
+    return delegate.shutdownNow();
+  }
+
+  @Override
+  public boolean isShutdown() {
+    return delegate.isShutdown();
+  }
+
+  @Override
+  public boolean isTerminated() {
+    return delegate.isTerminated();
+  }
+
+  @Override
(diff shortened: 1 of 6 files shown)


simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+
+  private final ExecutorService delegate;
+  private final Counter.Child submittedCounter;
+  private final Counter.Child completedCounter;
+  private final Gauge.Child runningGauge;
+  private final Histogram.Child durationHistogram;
+
+  static {
+    ST_COUNTER = Counter.build()
+        .name("instrumented_executor_service_submitted_count")
```
brian-brazil: Counters end with _total by convention, _count is used by Summaries/Histograms

simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+        .help("Instrumented executor service submitted total count")
+        .register();
+    CT_COUNTER = Counter.build()
+        .name("instrumented_executor_service_completed_count")
+        .labelNames("serviceName")
+        .help("Instrumented executor service completed total count")
+        .register();
+
+    R_GAUGE = Gauge.build()
+        .name("instrumented_executor_service_running_count")
```
brian-brazil: Gauges shouldn't have a suffix

simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+  private final Histogram.Child durationHistogram;
+
+  static {
+    ST_COUNTER = Counter.build()
+        .name("instrumented_executor_service_submitted_count")
+        .labelNames("serviceName")
+        .help("Instrumented executor service submitted total count")
+        .register();
+    CT_COUNTER = Counter.build()
+        .name("instrumented_executor_service_completed_count")
```
brian-brazil: The instrumented and executor seem kinda redundant here

simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+        .help("Instrumented executor service completed total count")
+        .register();
+
+    R_GAUGE = Gauge.build()
+        .name("instrumented_executor_service_running_count")
+        .labelNames("serviceName")
+        .help("Instrumented executor service running total count")
+        .register();
+
+    D_HISTOGRAM = Histogram.build()
```
brian-brazil: A histogram has buckets to cover 1ms-10s by default, is that appropriate for this sort of usage?

A quantile-less summary is probably a better choice.

simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class InstrumentedExecutorService implements ExecutorService {
+
+  private static final AtomicLong NAME_COUNTER = new AtomicLong();
+  private static final Counter ST_COUNTER;
```
brian-brazil: You can do this all in one line, ST_ isn't very descriptive.

simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+  private final ExecutorService delegate;
+  private final Counter.Child submittedCounter;
+  private final Counter.Child completedCounter;
+  private final Gauge.Child runningGauge;
+  private final Histogram.Child durationHistogram;
+
+  static {
+    ST_COUNTER = Counter.build()
+        .name("instrumented_executor_service_submitted_count")
+        .labelNames("serviceName")
```
brian-brazil: camel_case is the convention for label names, `executor` would be a better name

simpleclient_executor/src/main/java/io/prometheus/client/executor/InstrumentedExecutorService.java
```diff
@@ -0,0 +1,204 @@
...
+    D_HISTOGRAM = Histogram.build()
+        .name("instrumented_executor_service_duration_seconds")
+        .labelNames("serviceName")
+        .help("Instrumented executor service duration (s)")
+        .register();
+  }
+
+
+  public InstrumentedExecutorService(ExecutorService delegate) {
+    this(delegate, "instrumented_delegate_" + NAME_COUNTER.incrementAndGet());
```
brian-brazil: unknown_...

would be better. No point in making it longer than it has to be.

cenkakin:
Can you review PR again?

brian-brazil:
My high level comments still apply, I don't think this belongs here.

cemo:
I can not follow your comment. I can only see an instrumented implementation of executor services. CodaHale's metric has also implementation of instrumented executor services. I think that providing parity between these two project would be nice. What do you think? 

Disclosure: I am Cenk's colleague.

brian-brazil:
This is only one executor, what if someone already has an executor? This doesn't help them.

I don't think it's sane or useful to implement every possible Java interface.

cemo:
This is an not ordinary java interface. `ExecutorService` is one of the fundamental class in java concurrency package.  It is used almost all mainstream libraries such as `Netty`, `Spring`, `Tomcat`, `Guava`, `Jetty`, `Logback`. Their usage is not limited with simple things. They have extensive support for implementing concurrency related things around them. I think that `ExecutorService` is good asset for the project.

brian-brazil:
You didn't address my point.

cemo:
We are addressing the most common usage of `Executor` as it was done in other libraries. Even in the `java.util.concurrent.Executors` package, methods are returning `ExecutorService`. Guava has helper class `ForwardingExecutorService` for `ExecutorService` but not `Executor`. 

If someone needs to use an `Executor`, it will be mostly based on `AbstractExecutorService` or another useful class which is implementing `ExecutorService`.

So for the people who are using `Executor` which is not a subclass of `ExecutorService`, we have not so much gems for them. If they want to contribute, here is the code. They can improve it. We are trying to cover as much as people, as `Guava` did.

checketts:
I can see the value of having this in the client as I know it has been very painful for me to monitor my current executors. That being said, I didn't have much success with the Drop Wizard equivalent back when I was using it since it would have forced me to use that specific implementation.

How far could we get with an a collector can take and existing executor and expose metrics for it?

Alternatively (if you are using Spring) would using `TaskDecorator` be a good extension point? (see ThreadPoolTaskExecutor)

cemo:
@brian-brazil we are considering to release a new library named `client_java_extra` and will add new modules there. What do you think?

brian-brazil:
That naming might be a bit confusing, but you're free to release what you like.

PR #234, opened 2017-05-13
add `simpleclient_spring_hibernate` module

abacaphiliac:
add `simpleclient_spring_hibernate` module, which integrates `simpleclient_hibernate` and `simpleclient_spring_boot`.

this module provides an annotated ApplicationListener to listen for the SpringBoot ApplicationReadyEvent and registers HibernateStatisticsCollector with an autowired SessionFactory and annotated `spring.application.name`.

this was how i was able to activate Hibernate stats in a SpringBoot app and thought it might save others some time. what do you think?

### DIFF
--- a/simpleclient_spring_hibernate/src/main/java/io/prometheus/client/spring/hibernate/SimpleClientHibernateApplicationListener.java
+++ b/simpleclient_spring_hibernate/src/main/java/io/prometheus/client/spring/hibernate/SimpleClientHibernateApplicationListener.java
@@ -1,0 +1,48 @@
+package io.prometheus.client.spring.hibernate;
+
+import io.prometheus.client.hibernate.HibernateStatisticsCollector;
+import org.hibernate.Session;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
+import org.springframework.context.ApplicationListener;
+import org.springframework.stereotype.Component;
+
+import javax.persistence.EntityManager;
+
+@Component
+public class SimpleClientHibernateApplicationListener implements ApplicationListener<ApplicationReadyEvent> {
+  @Autowired
+  private EntityManager entityManager;
+
+  @Value("${spring.application.name:bootstrap}")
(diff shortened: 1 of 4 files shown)


simpleclient_spring_hibernate/src/main/java/io/prometheus/client/spring/hibernate/SimpleClientHibernateApplicationListener.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public void onApplicationEvent(ApplicationReadyEvent event) {
+    if (entityManager == null) {
+      return;
+    }
+
+    Session session = (Session) entityManager.getDelegate();
+
+    collector.add(session.getSessionFactory(), appName);
+    collector.register();
```
brian-brazil: Can this be called more than once? Usually you register the collector when you create it.

abacaphiliac: called more than once by SpringBoot? not that i've seen, but i'm new to SpringBoot.

i moved collector initialization into the field declaration so that i could inject a mock for the sake of coverage. i'd be happy to replace the mock-heavy test with an application integration test which verifies collector configuration some other way.

checketts: I believe it can be called more than once. Specifically when using config client.

abacaphiliac: full disclosure, the application i've integrated is using SpringBoot `1.5.x`.

i'm working on an integration test right now to replace my mock-heavy test. i'm not so sure my 1.5 integration will work with a 1.3 app. i'll keep poking at it. ideally i'll come up with a solution for this repo that works with spring boot >= 1.3, instead of just >= 1.5.

abacaphiliac:
i'll re-open when i have a solution that works with older versions of SpringBoot, and an integration test.

PR #237, opened 2017-05-22
Adds support for one/five/fifteen and mean rate values for dropwizard…

dekstroza:
This PR adds support for one/five/fifteen/mean values exposed on dropwizard meter metrics, along with existing total value.
@brian-brazil

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -111,8 +111,27 @@ public class DropwizardExports extends io.prometheus.client.Collector implements
                         Arrays.asList(new MetricFamilySamples.Sample(name + "_total",
                                 new ArrayList<String>(),
                                 new ArrayList<String>(),
-                                meter.getCount())))
-
+                                meter.getCount()))),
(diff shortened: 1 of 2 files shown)


simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
```diff
@@ -111,8 +111,27 @@ private static String getHelpMessage(String metricName, Metric metric){
                         Arrays.asList(new MetricFamilySamples.Sample(name + "_total",
                                 new ArrayList<String>(),
                                 new ArrayList<String>(),
-                                meter.getCount())))
-
+                                meter.getCount()))),
+                new MetricFamilySamples(name + "_one_minute_rate", Type.COUNTER, getHelpMessage(dropwizardName, meter),
```
checketts: Counters only go up. So you would want a `Gauge` instead.

That being said, Prometheus calculates rates better based off the `_total` that are already reported. So I would recommend against adding this feature.

brian-brazil:
You can use the `rate` function in PromQL to calculate these. _mean_rate is not a total, it's an average since process start - and thus almost entirely useless.

dekstroza:
I see, that will do so, this is not needed then. Thanks for quick response.
Dejan

PR #238, opened 2017-05-23
Added to conditional register method to avoid throwing exception on r…

jmostella:
…egistering an existing name

We ran into an issue where multiple components needed to register collector and push metrics to the same collector, without coordination. 

Currently the registry method throws and exception if there is an existing collector (by name).

Here is an example of what we are trying to achieve on export:

```
server_activeRequests{handler="ShopManagementService",method="createAuthor",} 0.0
server_activeRequests{handler="ShopManagementService",method="deleteTargetingData",} 0.0
server_activeRequests{handler="ShopService",method="getAggregatedHome",} 0.0
server_activeRequests{handler="ShopService",method="getAggregatedShowcaseV4",} 0.0
server_activeRequests{handler="ShopService",method="getAuthorList",} 0.0
server_activeRequests{handler="ShopService",method="getAvailableBannerList",} 0.0
server_activeRequests{handler="ShopService",method="getCategories",} 0.0
```
@brian-brazil

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -11,6 +11,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.*;
+import java.util.concurrent.Callable;
 
 /**
  * A registry of Collectors.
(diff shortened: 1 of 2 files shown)


brian-brazil:
The existing behaviour is correct, multiple components should not be sharing a metric.

In this case I believe you should be instrumenting up a level in the call stack.

jmostella:
I see thanks, I will share this with the team.

Issue #239, opened 2017-05-25
MetricsFilter can't be instantiated with Spring


renannprado:
The below code is a common pattern to instantiate a filter in Spring. However because the `init` is being called inside of that constructor, we get an exception as the `init` will be called again by the container.

```java
@Bean
    @Autowired
    public FilterRegistrationBean metricsFilter(@Value("${spring.application.name}") final String appName) throws ServletException {
        // using the default values as per documentation
        MetricsFilter metricsFilter = new MetricsFilter(
                appName, "The time taken fulfilling servlet requests",
                1, new double[]{0.005,0.01,0.025,0.05,0.075,0.1,0.25,0.5,0.75,1,2.5,5,7.5,10});

        FilterRegistrationBean bean = new FilterRegistrationBean(metricsFilter);
        bean.setOrder(Ordered.HIGHEST_PRECEDENCE);
        bean.addUrlPatterns("/api/*");
        return bean;
    }
```

Exception: **Caused by: java.lang.IllegalArgumentException: Collector already registered that provides name: app_name_count**

A workaround is to register the filter with a `web.xml` file.

checketts:
Thanks for the report. Another work around is to use the no-arg constructor and set the init parameters via the registration:

`bean.addInitParameter("metric-name", appName);`

renannprado:
You guys are fast! Thanks a lot!
When are you going to release the next version?

brian-brazil:
I'll go for one next week I think.

PR #240, opened 2017-05-25
Solve #239 - Don't call init in Filter constructor since it tries to register the metric twice

checketts:
Solves #239


brian-brazil:
Thanks!

PR #241, opened 2017-05-26
Fix Timer#close signature to not throws IOException

bitterfox:
Please consider following method:
```
void run() { // Do not throw anything
    heavy();
}
```
Now I'd like to measure how long does `heavy` take using Timer with try-with-resources:
```
void run() { // Do not throw anything
    try (Timer t = FOO.startTimer()) {
        heavy();
    }
}
```
Unexpectedly, we got compiler error like: 
```
Unhandled exception from auto-closeable resource: IOException
```
Because the signature of `close` throws IOE, but it never gonna throw it.

So we need to write like following:
```
void run() { // Do not throw anything
    try (Timer t = FOO.startTimer()) {
        heavy();
    } catch (IOException e) {/* never gonna be happen */}
}
or
void run() throws IOException { // sometime it impossible by constraint of super method
    try (Timer t = FOO.startTimer()) {
        heavy();
    }
}
```
These will lose benefit of try-with-resources.

Let's fix the signature of `close`.

@brian-brazil

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Gauge.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Gauge.java
@@ -1,7 +1,6 @@
 package io.prometheus.client;
 
 import java.io.Closeable;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -120,10 +119,9 @@ public class Gauge extends SimpleCollector<Gauge.Child> implements Collector.Des
 
      /**
       * Equivalent to calling {@link #setDuration()}.
-      * @throws IOException
       */
      @Override
-     public void close() throws IOException {
+     public void close() {
        setDuration();
      }
    }
--- a/simpleclient/src/main/java/io/prometheus/client/Histogram.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Histogram.java
@@ -1,7 +1,6 @@
 package io.prometheus.client;
 
 import java.io.Closeable;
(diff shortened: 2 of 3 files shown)


checketts:
Looks good to me. Were you able to determine that the had been there in the first place? (via a git-blame perhaps?)

brian-brazil:
It's implementing the Closeable interface, which throws this exception.

bitterfox:
Oh, sorry for late reply.
As Brian mentioned, it derived from signature of Closeable. It might have been generated by IDE...?

I'm not sure that should I add test cases for this or what kind of test cases are suitable.

brian-brazil:
Try-with-resources requires Java 7, all our tests are Java 6. If you can confirm that this still works as expected in a try-with-resources on Java 7&8 then it can go in.

bitterfox:
I can confirmed it still works well in my small sample which mentioned above.
```
void run() { // Do not throw anything
    try (Timer t = FOO.startTimer()) {
        heavy();
    }
}
```

But unfortunately, I've noticed that if someone have been wrote following already for workaround:
```
void run() { // Do not throw anything
    try (Timer t = FOO.startTimer()) {
        notThrowAnything();
    } catch (IOException e) {}
}
```
he will get compile-time error such as `Exception 'java.io.IOException' is never thrown in the corresponding try block`.

brian-brazil:
I think that breakage is acceptable for the sake of simpler usage.

bitterfox:
Thank you for merging!!

PR #247, opened 2017-06-01
add instrumented QueuedThreadPool

### LINKED ISSUE #242: Instrumenting org.eclipse.jetty.util.thread.QueuedThreadPool
Hi,

I think it would be very beneficial if we can get metrics for org.eclipse.jetty.util.thread.QueuedThreadPool in addition to io.prometheus.client.jetty.JettyStatisticsCollector. What do you think about that? 

Thanks in advance.

cenkakin:
fixes #242. 

I wonder what you think about adding utilization ratio of threads in the pool as a metric ( (threads - idleThreads) / threads ). It would be useful, but it can be also derived by the metrics which I already added.

Also, I am aware you suggested me to take HibernateStatisticsCollector as an example implementation. But, if I extend Collector and get instance variable of QueuedThreadPool instead of extending it, I cannot access its jobs size metric since org.eclipse.jetty.util.thread.QueuedThreadPool#getQueue is protected method. Therefore, I do it in this way.

I look forward to your comments.


brian-brazil:
> It would be useful, but it can be also derived by the metrics which I already added.

It's best to do such things on the Prometheus end.

> I cannot access its jobs size metric since org.eclipse.jetty.util.thread.QueuedThreadPool#getQueue is protected method. Therefore, I do it in this way.

Why not use getQueueSize()?

cenkakin:
getQueueSize() does not exist in jetty v9.0.2.v20130417 which was released in 2013. I think the version should be increased to more recent one like 9.4.4.v20170414 which spring boot supports in default.

brian-brazil:
The main thing we need to worry about in this module in terms of compatibility is the servlet, as that's what most users are going to use. As I presume that'll remaining working with older versions of jetty, I think we're okay to bump the version.

cenkakin:
The CI build failed, however it seems it passed in the log. Can you check it also?

brian-brazil:
The test is in the same module, so the version bump could have broken something.

The collector needs to support multiple pools, like the hibernate collector.

cenkakin:
@brian-brazil can you review the PR?

brian-brazil:
It has been brought to my attention that Jetty 9.4 requires Java 8. This is a Java 6 library, so at the least this will need to be split out into a separate module so we can keep the servlet working with 1.6.

cenkakin:
Okay, you are right. I added a new module named 'simpleclient_jetty_jdk8'.

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/QueuedThreadPoolStatisticsCollector.java
```diff
@@ -0,0 +1,79 @@
...
+    return Arrays.<MetricFamilySamples>asList(
+        buildGauge("queued_thread_pool_threads", "Number of total threads",
+            new ValueProvider() {
+              @Override
+              public double getValue(QueuedThreadPool queuedThreadPool) {
+                return queuedThreadPool.getThreads();
+              }
+            }
+        ),
+        buildGauge("queued_thread_pool_idle_threads", "Number of idle threads",
```
brian-brazil: threads_idle would be more typical

simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
```diff
@@ -0,0 +1,119 @@
...
+ * context.setContextPath("/");
+ * server.setHandler(context);
+ *
+ * HandlerCollection handlers = new HandlerCollection();
+ *
+ * StatisticsHandler statisticsHandler = new StatisticsHandler();
+ * statisticsHandler.setServer(server);
+ * handlers.addHandler(statisticsHandler);
+ *
+ * // register collector
```
brian-brazil: Register collector.

simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.eclipse.jetty.server.handler.StatisticsHandler;
+
+/**
+ * Collect metrics from jetty's org.eclipse.jetty.server.handler.StatisticsHandler.
+ * <p>
+ * <pre>{@code
+ * Server server = new Server(0);
```
brian-brazil: Maybe use a working port number here?

simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
```diff
@@ -0,0 +1,119 @@
...
+        name,
+        Collections.singletonList("code"),
+        Collections.singletonList(status),
+        value);
+  }
+
+  @Override
+  public List<MetricFamilySamples> collect() {
+    return Arrays.asList(
+        buildGauge("jetty_requests_total", "Number of requests", statisticsHandler.getRequests()),
```
brian-brazil: new CounterMetricFamily("my_counter_total", "help", 42)

simpleclient_jetty_jdk8/src/test/java/io/prometheus/client/jetty/JettyStatisticsCollectorTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.net.URL;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.server.handler.HandlerCollection;
+import org.eclipse.jetty.server.handler.StatisticsHandler;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.junit.After;
+import org.junit.Test;
+
+public class JettyStatisticsCollectorTest {
```
brian-brazil: This should be removed from its existing module.

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
```diff
@@ -72,11 +71,19 @@ private static MetricFamilySamples buildGauge(String name, String help, double v
...
+    return new MetricFamilySamples(
+        name,
+        Type.COUNTER,
+        help,
+        Collections.singletonList(new MetricFamilySamples.Sample(name, EMPTY_LIST, EMPTY_LIST, value)));
+  }
+
   private MetricFamilySamples buildStatusGauge() {
-    String name = "jetty_responses";
+    String name = "jetty_responses_total";
```
cenkakin: this one will be breaking change

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
```diff
@@ -72,11 +71,19 @@ private static MetricFamilySamples buildGauge(String name, String help, double v
...
 
+  private static MetricFamilySamples buildCounter(String name, String help, double value) {
+    return new MetricFamilySamples(
+        name,
+        Type.COUNTER,
+        help,
+        Collections.singletonList(new MetricFamilySamples.Sample(name, EMPTY_LIST, EMPTY_LIST, value)));
+  }
+
   private MetricFamilySamples buildStatusGauge() {
```
brian-brazil: This name should be changed to Counter for clarity

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
brian-brazil: There's a few breaking changes in here, getting them all done together is best.

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/QueuedThreadPoolStatisticsCollector.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public QueuedThreadPoolStatisticsCollector add(QueuedThreadPool queuedThreadPool, String name) {
+    queuedThreadPoolMap.put(name, queuedThreadPool);
+    return this;
+  }
+
+  @Override
+  public List<MetricFamilySamples> collect() {
+    return Arrays.<MetricFamilySamples>asList(
+        buildGauge("queued_thread_pool_threads", "Number of total threads",
```
brian-brazil: I'd prefix these with jetty_

simpleclient_jetty_jdk8/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brian-brazil: This comment remains

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
```diff
@@ -1,18 +1,17 @@
 package io.prometheus.client.jetty;
```
brian-brazil: Why is this file in both the jetty and jetty_jdk8 modules? I'd expect it to be in only one of them.

simpleclient_jetty_jdk8/pom.xml
cenkakin: If you mean maven compiler source and target, I did in the subsequent lines. I don't actually understand what you are referring to?

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
cenkakin: I don't want to depend one module to another. In this way, we can get rid of the older one in the future.
Also, in the case I remove this class in jetty module, if someone wants to add jetty_jdk8 dependendency, QueuedThreadPoolStatisticsCollector will be seen as duplicate. It would be also confusing. 
What is your suggestion about this?

brian-brazil: We're not getting rid of the older one anytime soon, and I don't want to maintain two copies of identical classes as they're likely to drift.

simpleclient_jetty_jdk8/pom.xml
brian-brazil: I don't see it, did you forget to add it to the change?

cenkakin: Line [20 -23] in this file.

brian-brazil: Ah, once again confirming that I'm blind.

simpleclient_jetty/src/main/java/io/prometheus/client/jetty/JettyStatisticsCollector.java
cenkakin: What about adding third module which includes common metrics collectors, i.e. JettyStatisticsCollector?

brian-brazil: I don't think that helps much, we going to run into some classes which work on 1.6 and other than need 1.8 so we'd likely just end up with jetty_stats and jetty_stats_jdk8.

brian-brazil:
That all looks good. The final thing is to add a short bit to the README about the new collector.

README.md
```diff
@@ -399,6 +400,39 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
...
+StatisticsHandler stats = new StatisticsHandler();
+stats.setHandler(server.getHandler());
+server.setHandler(stats);
+// Register collector.
+new JettyStatisticsCollector(stats).register();
+
+```
+
+Also, you can collect `QueuedThreadPool` metrics. If there is a single `QueuedThreadPool`
+to keep track of, following lines work:
```
brian-brazil: , use the following:

README.md
```diff
@@ -399,6 +400,39 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
...
+If you want to collect multiple `QueuedThreadPool` metrics, also you can achieve it like this:
+
+```java
+new QueuedThreadPoolStatisticsCollector()
+    .add(queuedThreadPool1, "myapp1")
+    .add(queuedThreadPool2, "myapp2")
+    .register();
+```
+
+Note: If you are using 9.3.x or more recent Jetty version which requires Java 8,
```
brian-brazil: This isn't quite right. If you want certain features you'd need both. As we don't currently state which module you need for what (and a github search will tell you), I'd say leave it out.

cemo:
Is this OK to merge @brian-brazil ?

README.md
```diff
@@ -399,6 +400,35 @@ or `EntityManagerFactory`, you can use this code to access the underlying `Sessi
 ```java
 SessionFactory sessionFactory = entityManagerFactory.unwrap(SessionFactory.class);
 ```
+### Jetty
+
+There is a collector for recording various Jetty server metrics. You can do it by  registering the collector like this:
+
+```java
+// Configure StaticticsHandler.
```
brian-brazil: Statistics

cenkakin:
What do you think right now?

brian-brazil:
Thanks!

PR #248, opened 2017-06-02
Remove checked exception from constructor since it is not needed

kdombeck:
Since this exception is not thrown in the constructor it is not needed.

@brian-brazil


brian-brazil:
Just noticed this myself today, thanks!

PR #251, opened 2017-06-07
Convert spring boot actuator metrics to prometheus 

hsdmiranda:
Spring boot actuator already measures some HTTP metrics and expose them, but the spring boot integration with prometheus just convert them in a way that is not really useful. See example bellow:

Spring boot actuator format:
``` 
"gauge.response.GET.metrics": 76.0,
"gauge.response.GET.info": 7.0,
"gauge.response.info": 25.0,
"gauge.response.metrics": 76.0
```

How is current been converted, what is quite useless if you try to create some dashboards 
```
# HELP gauge_response_GET_metrics gauge_response_GET_metrics
# TYPE gauge_response_GET_metrics gauge
gauge_response_GET_metrics 6.0
# HELP gauge_response_metrics gauge_response_metrics
# TYPE gauge_response_metrics gauge
gauge_response_metrics 5.0
```

And this is how is supposed to be:
```
# HELP gauge_response gauge_response
# TYPE gauge_response gauge
gauge_response{endpoint="/metrics", method="GET"} 4534.0
gauge_response{endpoint="/metrics"} 22.0
```

In case this PR is accepted I can also create another converter for COUNTERs.

### DIFF
--- a/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/GaugeResponseActuatorMetricConverter.java
+++ b/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/GaugeResponseActuatorMetricConverter.java
@@ -1,0 +1,55 @@
+package io.prometheus.client.spring.boot;
+
+import io.prometheus.client.Collector.MetricFamilySamples;
+import org.springframework.boot.actuate.metrics.Metric;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class GaugeResponseActuatorMetricConverter {
+
+	private List<String> labelsMerged;
+	private List<String> labelsPerHttpMethod;
+
+	private final static String GAUGE_RESPONSE_NAME = "gauge_response";
+
+	public GaugeResponseActuatorMetricConverter() {
+		this.labelsMerged = new ArrayList<String>();
+		this.labelsMerged.add("endpoint");
+
+		this.labelsPerHttpMethod = new ArrayList<String>(labelsMerged);
+		this.labelsPerHttpMethod.add("method");
+	}
+
+	public MetricFamilySamples.Sample convert(Metric<?> metric) {
+		MetricFamilySamples.Sample sample;
+		if (containsHttpMethod(metric.getName())) {
+			List<String> labelValues  = retrieveLabelValues("gauge.response.", metric.getName());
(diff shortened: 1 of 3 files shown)


t0kieu:
Hi, Is possible add HTTP response code (example: 200,400,500) to labels?

hsdmiranda:
It is for counters. Spring actuator expose those as well, but not for gauge. 
`"counter.status.GET.200.prometheus": 4913` 
If this PR is accepter i will also create a converter for it. In order to make spring boot expose all those metrics you need to enable in the configuration like this:
```
endpoints:
  metrics:
    filter:
      gauge-submissions: per-http-method, merged
      counter-submissions: per-http-method, merged
```

brian-brazil:
This seems very fragile and depends on assumptions of metric naming that are not going to be true in general. I don't think this is something we can include out of the box, as it's going to be environment specific.

hsdmiranda:
I did feel the same way once I wrote it, but it is environment specific just for spring boot. Without this converter the metrics becomes quite useless.

brian-brazil:
It's environment specific to what other metrics you have lying around, and which endpoints you have.

The spring boot metrics integration is only meant to be a stop gap until you can get instrumentation with the client library in place.

checketts:
> The spring boot metrics integration is only meant to be a stop gap until you can get instrumentation with the client library in place

@hsdmiranda Spring has initial support to milti-dimensional metrics coming https://github.com/spring-projects/spring-metrics/blob/master/src/main/java/org/springframework/metrics/instrument/web/MetricsWebFilter.java

It has a few point releases, so you can pull it in via maven. It is pretty new, but coming together pretty quickly.

I agree that the current `SpringbootMetricsCollector` is really useless in its current form without your changes, I would even recommend we deprecate it.

hsdmiranda:
> The spring boot metrics integration is only meant to be a stop gap until you can get instrumentation with the client library in place

I don't really agree with that. But I strongly agree with the fragile reason.

brian-brazil:
> I agree that the current SpringbootMetricsCollector is really useless in its current form without your changes, I would even recommend we deprecate it.

The same could be said about the Dropwizard integration. It's better than nothing, and gives you something to work from when starting out.

hsdmiranda:
I agree that is better than nothing. But with the current solution for spring boot is just like nothing :)

I think the solution as proposed by @checketts will be a better one, once they release the multi-dimension metrics

PR #257, opened 2017-06-13
Accept URL to Pushgateway constructor. Closes #258

### LINKED ISSUE #258: Pushgateway module doesn't support SSL
Pushgateway URL is hard-coded to use 'http', which prevents it from working in our environment.

theneva:
The `address` parameter to the Pushgateway constructor confused me a bit as I tried to simply pass in an actual URL (protocol and slash after path).

I think accepting a `java.net.URI` is better for hiding internals. The behaviour changes slightly as an exception is thrown when the `Pushgateway` is constructed (typically on startup), rather than upon trying to publish metrics.

The URI is `normalize()`d in case the caller includes trailing slashes in the URI.

As a side note, this allows `https` support rather than just `http`, which might be helpful.

I believe all the code paths I've introduced are visited by the existing test suite—please correct me if I'm wrong.

cc @brian-brazil will you please have a look?

### DIFF
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -7,6 +7,7 @@ import io.prometheus.client.exporter.common.TextFormat;
 import java.net.HttpURLConnection;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.net.URI;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.Collections;
@@ -62,7 +63,18 @@ public class PushGateway {
    * @param address  host:port or ip:port of the Pushgateway.
    */
   public PushGateway(String address) {
-    this.address = address;
+    this(URI.create("http://" + address));
+  }
+
+  /**
+   * Construct a Pushgateway, with the given URI.
+   * <p>
+   * @param serverBaseURI the base URI and optional context path of the Pushgateway server.
+   */
+  public PushGateway(URI serverBaseURI) {
+    this.address = URI.create(serverBaseURI.toString() + "/metrics/job/")
+        .normalize()
+        .toString();
   }
 
   /**
@@ -235,12 +247,14 @@ public class PushGateway {
   }
 
   void doRequest(CollectorRegistry registry, String job, Map<String, String> groupingKey, String method) throws IOException {
-    String url = "http://" + address + "/metrics/job/" + URLEncoder.encode(job, "UTF-8");
+    String url = address + URLEncoder.encode(job, "UTF-8");
+
     if (groupingKey != null) {
       for (Map.Entry<String, String> entry: groupingKey.entrySet()) {
         url += "/" + entry.getKey() + "/" + URLEncoder.encode(entry.getValue(), "UTF-8");
       }
     }
+
     HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
     connection.setRequestProperty("Content-Type", TextFormat.CONTENT_TYPE_004);
     if (!method.equals("DELETE")) {
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -59,7 +59,6 @@ public class PushGateway {
   // Visible for testing
   final String gatewayBaseURL;
   private static final int MILLISECONDS_PER_SECOND = 1000;
-
   /**
    * Construct a Pushgateway, with the given address.
    * <p>
@@ -272,7 +271,6 @@ public class PushGateway {
         url += "/" + entry.getKey() + "/" + URLEncoder.encode(entry.getValue(), "UTF-8");
       }
     }
-
     HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
     connection.setRequestProperty("Content-Type", TextFormat.CONTENT_TYPE_004);
     if (!method.equals("DELETE")) {
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
@@ -7,6 +7,7 @@ import io.prometheus.client.exporter.common.TextFormat;
 import java.net.HttpURLConnection;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLEncoder;
@@ -55,26 +56,43 @@ import java.io.OutputStreamWriter;
  */
 public class PushGateway {
 
-  private final String address;
+  // Visible for testing
+  final String gatewayBaseURL;
   private static final int MILLISECONDS_PER_SECOND = 1000;
+
   /**
    * Construct a Pushgateway, with the given address.
    * <p>
    * @param address  host:port or ip:port of the Pushgateway.
    */
   public PushGateway(String address) {
-    this(URI.create("http://" + address));
+    this(createURLSneakily("http://" + address));
   }
 
   /**
    * Construct a Pushgateway, with the given URI.
    * <p>
-   * @param serverBaseURI the base URI and optional context path of the Pushgateway server.
+   * @param serverBaseURL the base URL and optional context path of the Pushgateway server.
    */
-  public PushGateway(URI serverBaseURI) {
-    this.address = URI.create(serverBaseURI.toString() + "/metrics/job/")
-        .normalize()
-        .toString();
+  public PushGateway(URL serverBaseURL) {
+    this.gatewayBaseURL = URI.create(serverBaseURL.toString() + "/metrics/job/")
+      .normalize()
+      .toString();
+  }
+
+  /**
+   * Creates a URL instance from a String representation of a URL without throwing a checked exception.
+   * Required because you can't wrap a call to another constructor in a try statement.
+   *
+   * @param urlString the String representation of the URL.
+   * @return The URL instance.
+   */
+  private static URL createURLSneakily(final String urlString) {
+    try {
+      return new URL(urlString);
+    } catch (MalformedURLException e) {
+      throw new RuntimeException(e);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


brian-brazil:
Catching errors earlier is always good.

My one question is might URL be a better choice than URI?

theneva:
Thanks for having a look!

I prefer URI over URL here because URI has `normalize()`. Also, I also very rarely see `URL`s in our business logic—we tend to always use `URI`s—but that might be a quirk of my workplace.

I could add a third constructor that accepts a URL if you think that's better, just let me know!

theneva:
Eh, after thinking through it I agree that the constructor shouldn't really accept any URI.

I looked into it a bit more, and all `URL` constructors throw (same for `URI#toURL`), so the approach in this PR would break the existing API if we used `URL` by throwing.

The alternative is just accepting a String representation of a URI (or just `protocol/scheme`, `host`, `port`, and `context path` ad individual parameters), but I'm not too keen on that either.

The simplest thing for our client code is if we could just use URI. What do you think?

Eloo:
@theneva maybe you wanna check my old PR
https://github.com/prometheus/client_java/pull/155

brian-brazil:
URL vs URI doesn't make a difference exception wise, so I think we have to live with a breaking change there for the sake of a better API.

Reading through docs, it seems like only a URL would be the correct semantics as it's a URL we need. There is a `URI.toURL` function.

theneva:
Alright, I agree. Could catch the MalformedURLException and throw an unchecked exception instead to avoid breaking. I'll revise my PR asap!

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -62,7 +63,18 @@
...
+    this(URI.create("http://" + address));
+  }
+
+  /**
+   * Construct a Pushgateway, with the given URI.
+   * <p>
+   * @param serverBaseURI the base URI and optional context path of the Pushgateway server.
+   */
+  public PushGateway(URI serverBaseURI) {
+    this.address = URI.create(serverBaseURI.toString() + "/metrics/job/")
```
SimenB: Instead of `normalize`, you can do something like `this.address = new URL(serverBaseURL, "/metrics/job/");` if you receive an URL. That deals with trailing slashes.

`new URL(new URL("http://sup.com/"), "/something").toString()` => `http://sup.com/something`

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,15 +56,42 @@
  */
 public class PushGateway {
 
-  private final String address;
+  // Visible for testing
```
theneva: I don't know if this comment is necessary. I wanted to test that the URL is actually normalised, so that passing a trailing slash into the constructor accepting a String is okay.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,15 +56,42 @@
...
+
+  /**
+   * Construct a Pushgateway, with the given URL.
+   * <p>
+   * @param serverBaseURL the base URL and optional context path of the Pushgateway server.
+   */
+  public PushGateway(URL serverBaseURL) {
+    this.gatewayBaseURL = URI.create(serverBaseURL.toString() + "/metrics/job/")
+      .normalize()
+      .toString();
```
theneva: After trying a lot of different approaches to getting a normalised URL as a String out of this constructor, I ended up here.

The constructor accepts a `URL`, so we know that it has the required URL properties (i.e., a scheme is present and `:` is followed by either a valid port number or `/`).

Once we've established those facts, there is no harm in turning it into a more general `URI` for normalisation. I'm more happy with this solution.

Side note: I'm surprised at how loose the `URL` class's validator is. For example, `http://http://` is considered to be a valid URL.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,15 +56,42 @@
...
+   *
+   * @param urlString the String representation of the URL.
+   * @return The URL instance.
+   */
+  private static URL createURLSneakily(final String urlString) {
+    try {
+      return new URL(urlString);
+    } catch (MalformedURLException e) {
+      throw new RuntimeException(e);
+    }
```
theneva: This is my hack to avoid breaking the existing API, which would have happened otherwise due to introducing a checked exception.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -235,7 +264,8 @@ public void delete(String job, String instance) throws IOException {
   }
 
   void doRequest(CollectorRegistry registry, String job, Map<String, String> groupingKey, String method) throws IOException {
-    String url = "http://" + address + "/metrics/job/" + URLEncoder.encode(job, "UTF-8");
+    String url = gatewayBaseURL + URLEncoder.encode(job, "UTF-8");
```
theneva: the `gatewayBaseURL` is now always `<scheme>://<host>[:<port>]/metrics/job/`, so we can just concatenate the job name and grouping keys.

simpleclient_pushgateway/src/test/java/io/prometheus/client/exporter/PushGatewayTest.java
```diff
@@ -36,6 +36,20 @@ public void setUp() {
     groupingKey.put("l", "v");
   }
 
+  @Test(expected = RuntimeException.class)
+  public void testInvalidURLThrowsRuntimeException() {
+    new PushGateway("::"); // ":" is interpreted as port number, so parsing fails
+  }
```
theneva: This checks that the `createURLSneakily` method does its job.

simpleclient_pushgateway/src/test/java/io/prometheus/client/exporter/PushGatewayTest.java
```diff
@@ -36,6 +36,20 @@ public void setUp() {
...
+
+  @Test
+  public void testMultipleSlashesAreStrippedFromURL() {
+    final PushGateway pushGateway = new PushGateway("example.com:1234/context///path//");
+    Assert.assertEquals(
+        "http://example.com:1234/context/path/metrics/job/",
+        pushGateway.gatewayBaseURL
+    );
```
theneva: Here's why I've exposed the `gatewayBaseURL` field: it's mostly there to check that the code allows passing an address with a trailing slash to the original constructor, which is new (and improved, in my opinion) behaviour.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
theneva: Thanks! I've bypassed this problem by constructing a `URI` instance based on the `URL` that's passed to this constructor.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
brian-brazil: The comment is good. Please add a full stop.

For clarity I'd have a blank line after the variable, and explicitly make it `protected`.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
brian-brazil: `http://http://` is a valid URL :)

It's the HTTP protocol talking to the host `http` on the default port with a path of `//`.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
brian-brazil: I'm okay with a bit of breakage here. What we might do is leave it in for now, but put a TODO to remove it along with the various deprecated methods before this library becomes 1.0.

brian-brazil:
Thanks for the detailed explanatory comments.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
theneva: Do you think `protected` is better? Package local is the strictest thing available (protected allows access from both package and inheritance), but I'll change it if you prefer the added access from heirs.

brian-brazil: There's no way to explicitly say package, so protected is the next best thing.

This avoids misconceptions that you forgot to put in an access level.

theneva: Fine by me, I'll make the changes when I get the chance!

theneva: Done, done, and done.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
theneva: That makes sense, I was too quick there.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
theneva: I've added a TODO!

theneva:
Sorry about the delay, I've been super busy. I've made the requested changes!

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/PushGateway.java
```diff
@@ -54,15 +55,45 @@
...
+    this.gatewayBaseURL = URI.create(serverBaseURL.toString() + "/metrics/job/")
+      .normalize()
+      .toString();
+  }
+
+  /**
+   * Creates a URL instance from a String representation of a URL without throwing a checked exception.
+   * Required because you can't wrap a call to another constructor in a try statement.
+   *
+   * TODO: Remove this along with other deprecated methods before version 1.0 is released.
```
brian-brazil: We'll change the signature, not remove it.

brian-brazil:
Thanks!

theneva:
Thank you for helping me through it!

PR #259, opened 2017-06-13
Seconds are 1000 milliseconds, not the other way around

theneva:
I'm just cleaning up a bit as I'm reading the code :)

cc @brian-brazil


brian-brazil:
Thanks!

PR #256, opened 2017-06-10
fix NPE in PrometheusMvcEndpoint.value()

sunghyuk:
it resolves #252

PR #261, opened 2017-06-14
Fix NullPointerException

JWThewes:
Fix NPE when invoking MVCEndpoint without a name parameter


brian-brazil:
Did #256 not fix this?

brian-brazil:
I'm presuming this was already fixed.

JWThewes:
@brian-brazil Oh I'm totally sorry. Forgot to add a note...
Of course #256 fixed it. Just another solution of fixing it but it works.
Sorry :(

brian-brazil:
No problem at all, PRs fixing things are always welcome.

PR #269, opened 2017-06-22
Add interceptor to collect spring rest template metrics

### LINKED ISSUE #268: Implementing interceptor for collecting data from Spring RestTemplate
I think it would be very useful if prometheus serves us Spring RestTemplate metrics. What do you think about adding that to simpleclient_spring_web module? 

Thanks in advance,

cenkakin:
closes #268


simpleclient_spring_web/src/main/java/io/prometheus/client/spring/resttemplate/MetricsClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+  private final List<MetricsRestTemplateLabelType> labelTypes;
+
+  public MetricsClientHttpRequestInterceptor() {
+    this(DEFAULT_LABELS);
+  }
+
+  public MetricsClientHttpRequestInterceptor(List<MetricsRestTemplateLabelType> labelTypes) {
+    this.labelTypes = labelTypes;
+    metricClientInterceptorSummary = Summary.build()
+        .name("rest_template_request_duration_seconds")
```
brian-brazil: It's not quite obvious from this name what this means, are these templates being served back to the user?

Something more like spring_rest_client_template_request_duration_seconds would be clearer.

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/resttemplate/MetricsRestTemplateLabelType.java
```diff
@@ -0,0 +1,29 @@
...
+    public String resolve(HttpRequest request) {
+      String host = request.getURI().getHost();
+      return host != null ? host : "unknown";
+    }
+  },
+
+  PATH {
+    @Override
+    public String resolve(HttpRequest request) {
+      String path = request.getURI().getPath();
```
brian-brazil: Is this the template or the actual path?

simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/RestTemplateMetricsConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+          this.interceptor = this.context.getBean(MetricsClientHttpRequestInterceptor.class);
+        }
+        interceptors.add(interceptor);
+        interceptors.addAll(currentInterceptors);
+        restTemplate.setInterceptors(interceptors);
+      }
+      return bean;
+    }
+
+    private boolean checkMetricsInterceptorPresent(
```
checketts: This reads better as `isMetricsInterceptroPresent`

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/resttemplate/MetricsClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+        .labelNames(getLabelNames())
+        .help("Rest template request duration (s)")
+        .register();
```
checketts: It would be nice to be able to specify the registry, also this will allow the unit test to not rely on the default registry.

cenkakin:
@brian-brazil Why do you prefer hardcoded labels rather than configurable ones and which labels should be added (path, host, path) according to you?

I don't understand what your hesitation is in load balancing case. Can you explain it with examples?

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/resttemplate/MetricsClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+    final String[] labelNames = new String[size];
+    for (int i = 0; i < size; i++) {
+      labelNames[i] = labelTypes.get(i).name().toLowerCase();
+    }
+    return labelNames;
+  }
+
+  @Override
+  public ClientHttpResponse intercept(HttpRequest request, byte[] body,
+      ClientHttpRequestExecution execution) throws IOException {
```
t0kieu: What do you think about counting errors per destination endpoint? When request timeouts or it is unsuccessful.

brian-brazil:
> Why do you prefer hardcoded labels rather than configurable ones and which labels should be added (path, host, path) according to you?

Configurable labels are simpler, and work better for horizontal monitoring.

> I don't understand what your hesitation is in load balancing case. Can you explain it with examples?

If the url is http://myapi/blah then it works, if it's http://10.0.0.1/blah, http://10.0.0.2/blah,  http://10.0.0.3/blah etc. then that's a problem.

cenkakin:
> If the url is http://myapi/blah then it works, if it's http://10.0.0.1/blah, http://10.0.0.2/blah, http://10.0.0.3/blah etc. then that's a problem.
Do you mean that client makes a request to the hosts instead of load balancer? If that is the case, the host would be '10.0.0.[1,2,3]'. What is the exact problem in here? Dots?

brian-brazil:
> If that is the case, the host would be '10.0.0.[1,2,3]'. What is the exact problem in here? Dots?

Cardinality. If there's hundreds of ips over time that's going to cause performance issues.

cenkakin:
I see. Actually, in the first place, I give users freedom to configure the labels by considering their possible cardinality issue. Also, path label would be problematic here due to same reason with host.

Then, I guess, method type of request is left as the only label which we can track, right?

brian-brazil:
Method alone wouldn't be particularly useful. If we can get path before it's templated, then it should be okay. If we put warnings in the docs about this then I think it'd be okay.

cenkakin:
Unfortunately, we cannot get path before it is templated.

brian-brazil:
That's quite problematic then, as that basically only leaves us with method and host which isn't of too much use. I think we might need to consider instrumentation at a level above this so we can get the templated path.

cenkakin:
Instrumenting rest template can be considered to solve the problem. But it would be completely different implementation than this. So, I am closing this PR.

PR #272, opened 2017-07-11
Add metric name prefix parameter for DropwizardExports

goldobin:
Hi @brian-brazil  

This PR adds a possibility to add a prefix to the exported Dropwizard metric names.

Some libraries are exposing Dropwizard metrics with short library specific names (for example [Cassandra Java driver](https://github.com/datastax/java-driver)). When exporting them as Prometheus metrics it is useful to add the prefix to distinguish them from other modules metrics.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -14,13 +14,32 @@ import java.util.logging.Logger;
  * Collect Dropwizard metrics from a MetricRegistry.
  */
 public class DropwizardExports extends io.prometheus.client.Collector implements io.prometheus.client.Collector.Describable {
-    private MetricRegistry registry;
+    private final String prefix;
+    private final MetricRegistry registry;
     private static final Logger LOGGER = Logger.getLogger(DropwizardExports.class.getName());
 
     /**
      * @param registry a metric registry to export in prometheus.
      */
     public DropwizardExports(MetricRegistry registry) {
+        this("", registry);
+    }
+
+    /**
+     * @param prefix a prefix to put in front of each metric name.
+     * @param registry a metric registry to export in prometheus.
+     */
+    public DropwizardExports(String prefix, MetricRegistry registry) {
+
+        if (prefix == null) {
+            throw new NullPointerException("prefix is null");
+        }
+
+        if (registry == null) {
+            throw new NullPointerException("registry is null");
+        }
+
+        this.prefix = prefix.isEmpty() ? "" : prefix + "_";
         this.registry = registry;
     }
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
Could you explain a bit more about why you want this?

If some metrics don't have great names, it'd seem better to get those improved rather than adding prefixes to everything in the registry.

goldobin:
For example Cassandra driver exposes Dropwizard metrics and when it is exported via DropwizzardExports they are looking like this:
```
# HELP blocking_executor_queue_depth Generated from Dropwizard metric import (metric=blocking-executor-queue-depth, type=com.datastax.driver.core.Metrics$5)
# TYPE blocking_executor_queue_depth gauge
blocking_executor_queue_depth 0.0
# HELP connected_to Generated from Dropwizard metric import (metric=connected-to, type=com.datastax.driver.core.Metrics$2)
# TYPE connected_to gauge
connected_to 3.0
# HELP executor_queue_depth Generated from Dropwizard metric import (metric=executor-queue-depth, type=com.datastax.driver.core.Metrics$5)
# TYPE executor_queue_depth gauge
executor_queue_depth 0.0
# HELP known_hosts Generated from Dropwizard metric import (metric=known-hosts, type=com.datastax.driver.core.Metrics$1)
# TYPE known_hosts gauge
known_hosts 3.0
# HELP open_connections Generated from Dropwizard metric import (metric=open-connections, type=com.datastax.driver.core.Metrics$3)
# TYPE open_connections gauge
open_connections 4.0
# HELP reconnection_scheduler_task_count Generated from Dropwizard metric import (metric=reconnection-scheduler-task-count, type=com.datastax.driver.core.Metrics$5)
# TYPE reconnection_scheduler_task_count gauge
reconnection_scheduler_task_count 0.0
```
I think it is very hard at this point to request changes in Cassandra driver to add a prefix to metrics. Also suspect that some other libraries can do like Cassandra driver does.

brian-brazil:
If this is a problem here, it's probably also a problem for other people using it. This is probably best resolved either with Datastax, or by having an adapter before the metrics get into the dropwizard registry.

I'm wary of having any such functionality in the client, as those used to graphite-style metrics would be tempted to use it to prefix metrics with application/instance information rather than using labels.

goldobin:
Yeah, labels are the coolest feature of Prometheus! I'm also convincing guys on my team to use labels instead of generating a numerous metrics with suffixes (or prefixes). But for DropwizardExports you have no choice to use labels.

I think it is common that you have no option to modify modules/libraries that are already exposing DropWizard metrics. But you still need to somehow integrate them with your code which is already exposing well prepared Prometheus metrics prefixed with namespaces, subsystems, which is using labels.

Dropwizard itself is solving the problem with the prefix on reporter level. For example:

JmxReporter:

```
final JmxReporter reporter = JmxReporter
	.forRegistry(registry)
	.inDomain(clusterName + "-metrics")
	.build();

reporter.start()
```

GraphiteReporter:

```
final GraphiteReporter reporter = GraphiteReporter
	.forRegistry(registry)
    .prefixedWith("web1.example.com")
    // ...
    .build();

reporter.start(1, TimeUnit.MINUTES)
```

So for me, it seems that DropwizardExports is a right place to add such prefix to solve naming issue with the already written code.

brian-brazil:
The examples you've provided are where you'd use target labels in Prometheus, and thus that's not something I want to support. Your examples are not solving the issue of badly named metrics.

goldobin:
I see your point. If this change will lead to the client misuse it should be avoided. Will try to hack my problem somehow differently. Thanks for your time.

brian-brazil:
No problem.

One thing that might help is an adapter DW registry that you could register the Cassandra metrics with, which would then adjust the names and get registered with the actual DW registry.

goldobin:
Unfortunately, cassandra driver does not allow to supply your instance of `MetricsRegistry`. By now I've hacked it by extending `DropwizardExports` class and overriding package private `from*` methods. No so fancy but works :)

```java
public class PrefixedDropwizardExports extends DropwizardExports {
    private final String prefix;
    // ...
    @Override
    List<MetricFamilySamples> fromCounter(String dropwizardName, Counter counter) {
        return super.fromCounter(prefix + dropwizardName, counter);
    }
    // ...
}
```

Issue #274, opened 2017-07-12
simpleclient_spring_boot depends on spring-boot-starter-web now?


atrepczik:
In the most recent release of simpleclient, Spring Boot dependencies have been updated. Along with that it seems spring-boot-starter-web has become a compile time dependency whereas it previously has been test-scoped. Is this on purpose? It builds just fine without.

https://github.com/prometheus/client_java/compare/parent-0.0.23...parent-0.0.24#diff-b809dadb2515a75f45862ca6818b38b5L106

brian-brazil:
That looks like a mistake, would you like to send a PR readding the scope?

atrepczik:
Sure, I will create a PR shortly.

PR #275, opened 2017-07-12
Add back test scope to spring-boot-starter-web

atrepczik:
Fixes mistaken scope, should be test instead of compile, see #274.


brian-brazil:
Thanks!

PR #276, opened 2017-07-12
Add a Httpserver

brian-brazil:
@leogomes

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -42,14 +42,22 @@ public class HTTPServer {
         return response;
     }
 
-    public HTTPServer(int port) throws IOException {
+    public HTTPServer(InetSocketAddress addr) throws IOException {
         HttpServer mServer = HttpServer.create();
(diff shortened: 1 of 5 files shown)


simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -0,0 +1,112 @@
...
+ * {@code
+ * HTTPServer server = new HTTPServer(1234);
+ * }
+ * </pre>
+ * */
+public class HTTPServer {
+    static class HTTPMetricHandler implements HttpHandler {
+        private CollectorRegistry registry;
+
+        HTTPMetricHandler(CollectorRegistry registry) {
```
leogomes: Nice to have it as a parameter 👍

leogomes:
LGTM. I will update my PR as soon as this is released.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
brian-brazil: Passing the registry is largely for unittesting, it's not something users should generally need.

PR #277, opened 2017-07-14
allow user to specify time unit to use when converting dropwizard timers

asmabri:
@brian-brazil 

This is a pretty simple improvement that allows a user to specify a time unit to use for Timer dropwizard metrics when instantiating the DropwizardExports object.


brian-brazil:
Prometheus by convention uses base units such as seconds, which is why this is hardcoded.

Why do you want a different value?

asmabri:
Ah, I didn't realize using seconds was a prometheus convention. The javadoc on FromTimer led me to believe that it was supposed to be configurable. 

My use case was for measuring the latency of endpoints. When viewing dashboards and creating alerts, engineers in my org are used to using millis as the base unit, so that would make the transition to prometheus a little easier for me.
I understand that we don't want to encourage going against prometheus conventions, so feel free to close this PR.

brian-brazil:
Hmm, the javadoc there is a bit all over the place.

PR #283, opened 2017-07-24
Follow maven conventions and best-practices wrt version handling

davidkarlsen:
This change is [Reviewable](https://reviewable.io/reviews/prometheus/client_java/283)


brian-brazil:
There are non-maven build tools that can't deal with variables like this, so I'm afraid I can't accept this.

Issue #285, opened 2017-07-30
missing HistogramMetricFamily


dams:
The HistogramMetricFamily is mentioned in the documentation (see https://github.com/prometheus/client_java/blame/master/README.md#L625 ) but it seems that it was never implemented, as I can't find the class anywhere. Am I missing something ? Is it not yet implemented ? Thanks

checketts:
Good catch. Looks like it was overlooked.

brian-brazil:
The histogram metric is sufficiently complicated that a helper function would do basically nothing to reduce code complexity, thus it was not included.

dams:
@brian-brazil I think it could still simplify things, like separate between normal labels and buckets, and would help make code more readable/balanced when dealing with various types of metrics

brian-brazil:
Do you have an example of how it'd make it simpler?

I note that the need for histograms in custom collectors is quite rare, very few other systems produce histograms in a way compatible with what Prometheus uses.

dams:
@brian-brazil I'm not an expert, but I thought that HistogramMetricFamily could generate the _bucket, _count, _sum timeseries, given a basename, buckets definitions, and values. But maybe it's not adding enough value compared with doing it manually.

I'm coding a kafka exporter to provide a prometheus endpoint from kafka messages. We forge them, so we make sure they are compatible with prometheus.

brian-brazil:
You probably want a normal histogram for that.

dams:
I'm building a custom collector (reads kafka messages, provides a prom endpoint), so I would need to go from having normal histograms to something that I can add to a MetricFamilySamples. How do I do that ?

brian-brazil:
In the usual fashion, register the histogram.

dams:
that's what my first implementation was doing, but I moved to using GaugeMetricFamily because of the documentation paragraph about proxying. Anyway I suggest that the doc at least removes the mention of HistogramMetricFamily, and at best is clearer on when to use MetricFamily directly

brian-brazil:
But you're not proxying here, you're creating a new metric iirc.

checketts:
The readme line mentioning the `HistogramMetricFamily` was a section I had copied from the Python client readme. Does Python have a `HistogramMetricFamily`? Or does it also need to remove that line from the readme?

brian-brazil:
The Python client does, maps are easier to use in Python so a HistogramMetricFamily is useful there.

PR #286, opened 2017-07-31
state that there is no HistogramMetricFamily

dams:
Make it clear that there is no HistogramMetricFamily and briefly explain why (see #285 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,7 +622,7 @@ class YourCustomCollector extends Collector {
 static final YourCustomCollector requests = new YourCustomCollector().register()
 ```
 
-`SummaryMetricFamily` and `HistogramMetricFamily` work similarly.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -622,7 +622,7 @@ class YourCustomCollector extends Collector {
 static final YourCustomCollector requests = new YourCustomCollector().register()
 ```
 
-`SummaryMetricFamily` and `HistogramMetricFamily` work similarly.
+`SummaryMetricFamily` work similarly. The histogram metric is complicated so a `HistogramMetricFamily` helper would do basically nothing to reduce code complexity, thus it was not included.
```
brian-brazil: I'd just mention SummaryMetricFamily here without going into design decisions. Also s/work/works/.

dams:
should be good now

brian-brazil:
Thanks!

PR #290, opened 2017-08-29
Export dropwizard meter minute rate metrics

cwins:
@brian-brazil - Found the minute rates available from Dropwizard Metered metric missing when such metrics are used. Expanded the meter translator with functionality for exposing all available minute rates as SUMMARY.


brian-brazil:
These can be calculated from the count on the Prometheus end, there's no reason to expose them.

cwins:
Yes - just came to the same conclusion after playing with rate() - closing

PR #292, opened 2017-09-10
Undertow Web handler for simple client

kislitsyn:
This is implementation of Undertow web handler.
Require Java 1.7, cause Undertow use this version for compilation


brian-brazil:
Thanks for the PR. I'd suggest that you maintain this in your own repository, as having everything in one place is straining maintainability a bit. We can list it and any other java integrations on the exporters and integrations page down the bottom.

PR #300, opened 2017-10-10
 Adding ability to control whether the httpserver is a daemon thread

ilucker:
In some cases it is desirable to run HttpServer in daemon threads. E.g. look into the issue when [Java agent preventing cassandra startup #170](https://github.com/prometheus/jmx_exporter/issues/170)

This pull request add ability to control whether the httpserver will use daemon thread or not (default behavior for backward capability).

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -13,7 +13,10 @@ import java.util.List;
 import java.util.Set;
 import java.util.HashSet;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.ThreadFactory;
 import java.util.zip.GZIPOutputStream;
 
 import com.sun.net.httpserver.HttpHandler;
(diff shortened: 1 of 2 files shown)


simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -107,34 +110,109 @@ protected static boolean shouldUseCompression(HttpExchange exchange) {
...
+
+        CustomThreadFactory(ThreadFactory delegate, boolean daemon) {
+            this.delegate = delegate;
+            this.daemon = daemon;
+        }
+
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = delegate.newThread(r);
+            if (daemon ^ t.isDaemon()) {
```
brian-brazil: why not always setDaemon?

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -107,34 +110,109 @@ protected static boolean shouldUseCompression(HttpExchange exchange) {
...
+                @Override
+                public void run() {
+                    server.start();
+                }
+            }, null);
+            CustomThreadFactory.defaultThreadFactory(daemon).newThread(startTask).start();
+            try {
+                startTask.get();
+            } catch (InterruptedException e) {
+                //should not happen. but propagate interrupted flag just in case.
```
brian-brazil: Comments should be full sentences.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -107,34 +110,109 @@ protected static boolean shouldUseCompression(HttpExchange exchange) {
...
+                Thread.currentThread().interrupt();
+            } catch (ExecutionException e) {
+                Throwable cause = e.getCause();
+                if (cause instanceof RuntimeException) {
+                    throw (RuntimeException) cause;
+                } else if (cause instanceof Error) {
+                    throw (Error)cause;
+                } else throw new RuntimeException("Unexpected exception on starting HttpSever", cause);
```
brian-brazil: The class is HTTPServer

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -107,34 +110,109 @@ protected static boolean shouldUseCompression(HttpExchange exchange) {
...
+            server.start();
+        } else {
+            FutureTask<Void> startTask = new FutureTask<Void>(new Runnable() {
+                @Override
+                public void run() {
+                    server.start();
+                }
+            }, null);
+            CustomThreadFactory.defaultThreadFactory(daemon).newThread(startTask).start();
+            try {
```
brian-brazil: Is this really necessary? We've just re-throwing the exception.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
ilucker: It does not matter in this case. I just looked through the code of DefaultThreadFactory and they use similar code: 
```java
  if (t.isDaemon())
    t.setDaemon(false);
```
Will change to setDaemon since it looks  simpler.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
ilucker: Otherwise i have to implement internal class which should implement Runnable and store exceptions in the instance fields to re-throw them later in the main thread:
```java
class StartTask implements Runnable {
    private volatile RuntimeException exception;
    private volatile Error error;

    @Override
    public void run() {
        try {
            server.start();
        } catch (RuntimeException e) {
            exception = e;
        } catch (Error e) {
            error = e;
        }
    }

    void rethrow() {
        if (exception != null) throw exception;
        if (error != null) throw error;
    }
}
StartTask task = new StartTask();
Thread thread = CustomThreadFactory.defaultThreadFactory(daemon).newThread(task);
thread.start();
try {
    thread.join();
    //should not happen. but propagate interrupted flag just in case.
    Thread.currentThread().interrupt();
}
task.rethrow();
``` 

i decided to use tool that JDK provides for this – FutureTask

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
ilucker: This message related to exception on starting com.sun.net.httpserver.HttpServer, but if you insist i will change.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
brian-brazil: I don't see why the current implementation has to do so much exception handling, I would expect it to have to do none.

This also all runs in the main thread.

ilucker:
@brian-brazil Thank you for reviewing it. I've rebased PR with the latest changes according to your review.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
ilucker: In the current implementation you just call server.start() in main thread. It can complete successfully or with Error or RuntimeException. In the latest case the HTTPServer will not be created and exception will be propagated to HTTPServer constructor caller. Now put server.start() into a separate thread and i want to propagate potential Error/RuntimeException to the HTTPServer again.

brian-brazil: I must be missing something here. All I see is rethrowing exceptions in the same thread. Thus this seems quite a bit more complicated than it needs to be.

ilucker: Ok. Will try to explain. RuntimeException or Error may happen in a thread that i launch in 
```java
DaemonThreadFactory.defaultThreadFactory(daemon).newThread(startTask).start();
```
The call `startTask.get()` will block until task is completed. Since it is blocking call it can be interrupted and throw InterruptedException. InterruptedException is a checked exception and should be handled explicitly. The same with an ExecutionException which is in signature of a `Future.get` and I have to handle it as well. Since I can not extend signature of the HTTPServer constructors via new checked exception, I have handle it here and the best way to do not swallow an error is to re-throw as  RuntimeException/Error.

ilucker: @brian-brazil any thoughts?

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -107,34 +110,109 @@ protected static boolean shouldUseCompression(HttpExchange exchange) {
...
+                @Override
+                public void run() {
+                    server.start();
+                }
+            }, null);
+            DaemonThreadFactory.defaultThreadFactory(daemon).newThread(startTask).start();
+            try {
+                startTask.get();
+            } catch (ExecutionException e) {
+                Throwable cause = e.getCause();
```
brian-brazil: You can simplify by just rethrowing this. It's probably a port already in use.

ilucker: Do you mean rethrowing ExecutionException? In JDK it is declared as 
```Java
public class ExecutionException extends Exception
```
It is a checked exception and therefore to rethrow it I have to change signature of `start` to 
```Java
private void start(boolean daemon) throws ExecutionException {
```
and then I have to change signature of every constructor of HTTPServer and add the same ExecutionException in it. It will break backward compatibility with all existing clients of  HTTPServer class. 
We know, that `startTask.get();` executes server.start(); which can throw RuntimeException and Errors only, and the best place to convert ExecutionException into the cause RuntimeException or Error is in this place rather then let clients of HTTPServer deal with it.

brian-brazil: You could rethrow it as a RuntimeException

ilucker: That is exactly what I’m doing here, with a small optimization. This is a standard practice in Java world to analyze cause of execution exception and dethrone it. But if you prefer to wrap a wrapper of runtime exception into yet another runtime exception I can do that to pith this fix forward.

ilucker: @brian-brazil changed per review.

brian-brazil:
Thanks!

I'll put a java release on my todo list.

PR #303, opened 2017-10-16
Shutdown HTTPServer executor service to avoid non-daemon threads potentially blocking exiting JVM

marcelmay:
The HTTPServer creates a Java ExecutorService but does not shut it down when stopping the HTTPServer. 

As the default executor service creates non-daemon threads, this will cause the JVM to hang if the JVM does not exist using a forced System.exit(...).

@brian-brazil  : This happens currently when using the latest jmx_exporter (disable the System.exit(...) in TestApplication to reproduce/verify).


marcelmay:
Just saw related PR #300

brian-brazil:
Thanks!

PR #304, opened 2017-10-24
Add preferred numbers for histogram buckets

bharrisau:
This change adds a helper method for creating buckets following the preferred number style. It is similar to exponential buckets, but limits the significant figures in the output. Using 6 buckets per decade limits the relative error to 20% (https://en.wikipedia.org/wiki/E-series_of_preferred_numbers)

https://en.wikipedia.org/wiki/Preferred_number

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Histogram.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Histogram.java
@@ -55,8 +55,9 @@ import java.util.Map;
  *         .name("requests_latency_seconds").help("Request latency in seconds.").register();
  * }
  * </pre>
- * {@link Histogram.Builder#linearBuckets(double, double, int) linearBuckets} and
- * {@link Histogram.Builder#exponentialBuckets(double, double, int) exponentialBuckets}
+ * {@link Histogram.Builder#linearBuckets(double, double, int) linearBuckets},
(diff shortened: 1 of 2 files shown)


brian-brazil:
How does this help a user compared to using exponential? I'd expect exponential to have lower error for the same number of buckets.

The name is also rather presumptive.

bharrisau:
Hmm, a couple of differences to come mind:
- User doesn't need to enter the fifth root of 10 as the exponential factor
- User can reason about how wide a range the histogram covers - they could get a calculator and do `start * exp ^ count`, but this way if you set count to some factor of buckets, then you know how many decades it covers (e.g. buckets = 5, count = 15, start = 0.1 -> then end is going to be up to 100)
- If User previously was starting from 100ms but then wanted to do 10ms, this way keeps the larger buckets static so they don't lose all the data
- Potential performance improvement on prometheus due to less `le_X` labels. The exponential creates longer labels, and if any variable are different then there is no overlap with other series's labels.

Potential questions to improve this would be:
- Should there be an `end` instead of a `count`? Then the range is even more explicit.

I could find a different term than `preferred` if needed. It's just the term for using this style of number.

brian-brazil:
I don't think this makes sense, bucket values are not really meant for human consumption. There's already an API to provide your own buckets if you want to use these.

PR #305, opened 2017-10-26
Add BufferPool metrics

giannello:
Add metrics for BufferPool (https://docs.oracle.com/javase/8/docs/api/java/lang/management/BufferPoolMXBean.html)

`jvm_buffer_pool_bytes_used`: Used bytes of a given JVM buffer pool.
`jvm_buffer_pool_bytes_capacity`: Bytes capacity of a given JVM buffer pool.
`jvm_buffer_pool_buffers_used`: Used buffers of a given JVM buffer pool.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryPoolsExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryPoolsExports.java
@@ -3,6 +3,7 @@ package io.prometheus.client.hotspot;
 import io.prometheus.client.Collector;
 import io.prometheus.client.GaugeMetricFamily;
 
+import java.lang.management.BufferPoolMXBean;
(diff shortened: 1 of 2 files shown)


brian-brazil:
This client is targeted at Java 6, and I'm afraid this bean was only added in Java 7.

giannello:
Indeed, that bean was introduced in Java7.
If you can share your preferred way to make sure this feature won't interfere with the java6 compatibility, I would be happy to rework my PR.

brian-brazil:
I'm not aware of a good way to do that. Do you have any suggestions?

pjfanning:
@giannello could you use java reflection to access BufferPoolMXBean? If the class is not found (eg when you run with Java 6 Runtime), then return zeroes for the new metrics and if the class is found then use reflection to access the values.

giannello:
@pjfanning thanks for the hint and for reminding me about this PR.

PR #306, opened 2017-11-02
Extension and customization of TextFormat exporter

iddol:
Enable extension and customization by changing private methods to protected and breaking down logic of main method to smaller methods.
@brian-brazil

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -20,38 +20,45 @@ public class TextFormat {
      * for the output format specification. */
     while(mfs.hasMoreElements()) {
       Collector.MetricFamilySamples metricFamilySamples = mfs.nextElement();
-      writer.write("# HELP ");
-      writer.write(metricFamilySamples.name);
-      writer.write(' ');
-      writeEscapedHelp(writer, metricFamilySamples.help);
-      writer.write('\n');
-
-      writer.write("# TYPE ");
-      writer.write(metricFamilySamples.name);
-      writer.write(' ');
-      writer.write(typeString(metricFamilySamples.type));
-      writer.write('\n');
-
+      writeHeader(writer, metricFamilySamples);
       for (Collector.MetricFamilySamples.Sample sample: metricFamilySamples.samples) {
-        writer.write(sample.name);
-        if (sample.labelNames.size() > 0) {
-          writer.write('{');
-          for (int i = 0; i < sample.labelNames.size(); ++i) {
-            writer.write(sample.labelNames.get(i));
-            writer.write("=\"");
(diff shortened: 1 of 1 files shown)


brian-brazil:
Why do you want to extend this?

iddol:
We use multiple collector registries, and our servlet exports the samples from all of them.  The same collector can appear in more than one registry.
If we simply concatenate the output of each registry, the result contains the TYPE and HELP headers more than once, which is not valid.  To avoid this, the servlet maintains a set of the names of all the metrics that were already collected, and when it encounters one of the collected metrics it skips the headers and reports just the samples.
It seems to be working OK, but currently I have to duplicate all the code your TextFormat class.  We would like to make it more extensible, so we can reuse your code.

brian-brazil:
Changing the exposition code is the wrong way to deal with that, the problem here is the way you're using the registry. You should have only one registry in standard usage.

iddol:
We have standard metrics that are collected by all or our services.  This works OK when each service is deployed independently.  But we have few services deployed in the same web app and thus running in the same JVM.  When they try to register the collector, they fail because the default registry is static and doesn't support multiple definitions of collectors with the same name. We overcome this by using a separate registry for each service, which is the same as what we have when each service is deployed independently.

brian-brazil:
It sounds like you might want to have a different endpoint per application, as you're basically a clustering system at that point.

iddol:
If I understand correctly, you're suggesting to continue working with multiple registries, but to expose a different URL for scraping the samples from each registry?  That is, instead of http://<webapp>/metrics that collects samples from all services, have http:<webapp>/<serviceName> for each service?

brian-brazil:
Yes.

iddol:
But we have a generic monitor that sends the http://webapp/metrics request to all our servers, and doesn't know what's deployed on each server.  Per your suggestion, the monitor would somehow need to know which service is deployed on each server, in order to construct all the proper URLs.

brian-brazil:
> Per your suggestion, the monitor would somehow need to know which service is deployed on each server, in order to construct all the proper URLs.

Yes, that's standard when dealing with a cluster scheduler. Multiple independent apps inside on JVM is the same as multiple independent processes on a machine.

brian-brazil:
It sounds like what you've built is a monitoring system. The java client is not is intended to be used in that fashion, rather it feeds data into the Prometheus monitoring system.

I think you need to write a exporter to convert between your custom monitoring system and Prometheus.

archmisha:
What I described is our exported for prometheus. We don't want it to know every service and be required to uodate it for every new deployed service. This is also different in different deployment environments

brian-brazil:
> What I described is our exported for prometheus.

You're attempting to implement a monitoring systems exporter as direct instrumentation, as you've discovered this does not work.

Fundamentally, you need to handle this before it hits a registry as this is something to be handled in a custom collector. There's no changes required in the exposition code.

PR #307, opened 2017-11-06
SimpleCollecor.labels() initialization performance

mcoolive:
- Now, we do 3 lookups in the children Map when we create a new tupple of
labels. From my understanding, it's to manage concurrent initializations.

It can be improved. I suggest to check the result of 'putIfAbsent' to avoid the third
lookup.

Because we do not want to create a new Child for each call to #labels, I
don't find a solution with only 1 lookup. BUT with Java 8, we can consider
the use of 'computeIfAbsent' with a lambda.


mcoolive:
An opinion would be welcome.

brian-brazil:
Thanks!

PR #317, opened 2017-11-21
add initial memory metrics

pjfanning:
https://github.com/prometheus/client_java/issues/312

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryPoolsExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryPoolsExports.java
@@ -29,6 +29,10 @@ import java.util.List;
  * </pre>
  */
 public class MemoryPoolsExports extends Collector {
+  private static final List<String> AREA_LABELS = Collections.singletonList("area");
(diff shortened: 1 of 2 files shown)


brian-brazil:
Can you limit the PR to the issue at hand?

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryPoolsExports.java
```diff
@@ -71,6 +71,14 @@ void addMemoryAreaMetrics(List<MetricFamilySamples> sampleFamilies) {
     max.addMetric(Collections.singletonList("heap"), heapUsage.getMax());
     max.addMetric(Collections.singletonList("nonheap"), nonHeapUsage.getMax());
     sampleFamilies.add(max);
+
+    GaugeMetricFamily init = new GaugeMetricFamily(
+        "jvm_memory_bytes_init",
+        "Init bytes of a given JVM memory area.",
```
brian-brazil: Initial would be clearer. This reads as an instruction.

brian-brazil:
Thanks!

PR #319, opened 2017-12-01
Produce metric for max thread pool size

tommy-bo:
To be able to measure and alert on capacity

### DIFF
--- a/simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/QueuedThreadPoolStatisticsCollector.java
+++ b/simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/QueuedThreadPoolStatisticsCollector.java
@@ -37,6 +37,8 @@ public class QueuedThreadPoolStatisticsCollector extends Collector {
         buildGauge("jetty_queued_thread_pool_threads_idle", "Number of idle threads",
             QueuedThreadPool::getIdleThreads),
         buildGauge("jetty_queued_thread_pool_jobs", "Number of total jobs",
+            QueuedThreadPool::getQueueSize),
+        buildGauge("jetty_queued_thread_pool_max", "Max size of thread pool",
             QueuedThreadPool::getQueueSize));
   }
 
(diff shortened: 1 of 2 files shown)


simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/QueuedThreadPoolStatisticsCollector.java
```diff
@@ -37,6 +37,8 @@ public QueuedThreadPoolStatisticsCollector add(QueuedThreadPool queuedThreadPool
         buildGauge("jetty_queued_thread_pool_threads_idle", "Number of idle threads",
             QueuedThreadPool::getIdleThreads),
         buildGauge("jetty_queued_thread_pool_jobs", "Number of total jobs",
+            QueuedThreadPool::getQueueSize),
+        buildGauge("jetty_queued_thread_pool_max", "Max size of thread pool",
```
brian-brazil: I don't think you've got the right function.

tommy-bo: I think you're right :man_facepalming:

simpleclient_jetty_jdk8/src/main/java/io/prometheus/client/jetty/QueuedThreadPoolStatisticsCollector.java
```diff
@@ -37,7 +37,9 @@ public QueuedThreadPoolStatisticsCollector add(QueuedThreadPool queuedThreadPool
         buildGauge("jetty_queued_thread_pool_threads_idle", "Number of idle threads",
             QueuedThreadPool::getIdleThreads),
         buildGauge("jetty_queued_thread_pool_jobs", "Number of total jobs",
-            QueuedThreadPool::getQueueSize));
+            QueuedThreadPool::getQueueSize),
+        buildGauge("jetty_queued_thread_pool_max", "Max size of thread pool",
```
brian-brazil: threads_max would seem more in line with the above metrics

tommy-bo: Perhaps. Or would size be more appropriate?

brian-brazil: I'd say that it's implicit that it's the maximum size of the threadpool, there's nothing else it can really be the max of.

brian-brazil:
Thanks!

tommy-bo:
Thanks for your time, much abliged :-)

MouadBenchchaoui0000:
Sorry if this the wrong channel to ask, but I am wondering why this is done for jdk8 only and `simpleclient_jetty`? we are more interested in the thread pool metrics but I am not sure which to use.

PR #322, opened 2017-12-13
MetricsFilter package changed

fschlag:
While integrating the MetricsFilter into our project I found out that the package of the MetricsFilter changed, so I updated the documentation.


brian-brazil:
Thanks!

PR #325, opened 2018-01-07
Fix bug in CollectorRegistry#collectorNames method

savulchik:
The switch statement was missing breaks for SUMMARY and HISTOGRAM cases.
Due to this bug wrong and duplicate collector names are produced for summaries and histograms.
cc @brian-brazil


brian-brazil:
Thanks!

PR #331, opened 2018-01-21
Adding Histogram as an optional type of Collector for Spring Web PrometheusTimeMethod annotated methods

mmattozzi:
The `PrometheusTimeMethod` annotation is a very convenient way to collect timing metrics for a method -- I think it makes sense to let a user choose between `Summary` and `Histogram` collectors since they might want to calculate percentiles and aggregate data across hosts. 

```
@PrometheusTimeMethod(name = "foobar", help = "help info", collectorClass = Histogram.class)
```

For backwards compatibility, collectorClass defaults to `Summary`.

This code would be a little cleaner if Summary and Histogram inherited startTimer from a common interface (and used the same Timer object) but I didn't want to make any changes to the core classes. Open to suggestions if this implementation doesn't sit well.

### DIFF
--- a/simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
+++ b/simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
@@ -1,5 +1,7 @@
 package io.prometheus.client.spring.web;
 
+import io.prometheus.client.Histogram;
+import io.prometheus.client.SimpleCollector;
 import io.prometheus.client.Summary;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
@@ -10,6 +12,7 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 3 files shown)


simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
```diff
@@ -25,8 +28,8 @@
 @Scope("prototype")
 @ControllerAdvice
 public class MethodTimer {
-    private final ReadWriteLock summaryLock = new ReentrantReadWriteLock();
-    private final HashMap<String, Summary> summaries = new HashMap<String, Summary>();
+    private final ReadWriteLock collectorLock = new ReentrantReadWriteLock();
+    private final HashMap<String, SimpleCollector> collectors = new HashMap<String, SimpleCollector>();
```
brian-brazil: SimpleCollector is intended as an implementation detail, it shouldn't be used directly.

mmattozzi:
My thought was that using a pre-made annotation eliminates a client from having to pepper their code with: 
```
startTimer
try {
  // do stuff
} finally {
  observeDuration
}
```

To me, even adding a `histogramBuckets` parameter to this annotation (taking an array of bucket values) would result in more minimal code additions for Prometheus instrumenting than just using the client library normally. I do think that the default histogram buckets are fairly well suited to people shooting for web API timing measurements and I'm sure a lot of people do just use the defaults. I would argue that a Summary also doesn't have a ton of value without being able to specify an interesting quantile since a lot of people are more interested in 95th or 99th percentiles than average. However, that would be harder to condense into an annotation parameter than a list of bucket values...

simpleclient_spring_web/src/main/java/io/prometheus/client/spring/web/MethodTimer.java
mmattozzi: Can definitely change this to `Collector` if we move forward with this PR. Thanks!

brian-brazil:
Quantiles might be useful for measuring end-user experience, but due to the inability to do math on them their uses elsewhere are limited. An average is useful everywhere, and when you're timing functions you usually want averages - particularly when cardinality is considered.
I'd also expect that most functions would take quite a bit less time than the default histogram buckets envision.
Overall I'm not sure the complexity is worth it. I see histograms more as something to use at the start and end of your system, not the middle. The servlet filters are more useful for histograms.

PR #334, opened 2018-01-24
Fix javadocs link in README

matthavener:


brian-brazil:
Thanks! I guess the new website broke this.

PR #335, opened 2018-01-27
Improve CollectorRegistry#unregister()

mcoolive:
- In 'unregister', remove the collector before its names, so we do operations
in the reverse order compare to 'register'. It is consistent with 'clear'.
- So we do 1 lookup in collectors hash map instead of 2.

Signed-off-by: Cyril Martin <c.martin@criteo.com>


brian-brazil:
Thanks!

PR #337, opened 2018-02-01
Added optional "timestamp" value as part of metric sample

efelcon:
I'm using an exporter based on this code and it's working fine.

To be honest I'm not using `equals()` and `hashCode()` methods, so I haven't test it.

One more thing, about `hashCode()`: I'm not totally sure if code is totally agree with the current structure, so please check.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -77,12 +77,18 @@ public abstract class Collector {
       public final List<String> labelNames;
       public final List<String> labelValues;  // Must have same length as labelNames.
       public final double value;
+      public final Long timestamp;
 
-      public Sample(String name, List<String> labelNames, List<String> labelValues, double value) {
+      public Sample(String name, List<String> labelNames, List<String> labelValues, double value, Long timestamp) {
         this.name = name;
         this.labelNames = labelNames;
         this.labelValues = labelValues;
         this.value = value;
+        this.timestamp = timestamp;
+      }
+
+      public Sample(String name, List<String> labelNames, List<String> labelValues, double value) {
+    	  this(name, labelNames, labelValues, value, null);
       }
 
       @Override
@@ -92,7 +98,8 @@ public abstract class Collector {
         }
         Sample other = (Sample) obj;
         return other.name.equals(name) && other.labelNames.equals(labelNames)
-          && other.labelValues.equals(labelValues) && other.value == value;
+          && other.labelValues.equals(labelValues) && other.value == value
+		  && other.timestamp.equals(timestamp);
       }
 
       @Override
@@ -103,13 +110,14 @@ public abstract class Collector {
         hash = 37 * hash + labelValues.hashCode();
         long d = Double.doubleToLongBits(value);
         hash = 37 * hash + (int)(d ^ (d >>> 32));
+        hash = 37 * hash + timestamp.hashCode();
         return hash;
       }
 
       @Override
       public String toString() {
         return "Name: " + name + " LabelNames: " + labelNames + " labelValues: " + labelValues +
-          " Value: " + value;
+          " Value: " + value + " Timestamp: " + timestamp;
       }
     }
   }
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -117,7 +117,8 @@ public abstract class Collector {
         hash = 37 * hash + labelValues.hashCode();
         long d = Double.doubleToLongBits(value);
         hash = 37 * hash + (int)(d ^ (d >>> 32));
-        hash = 37 * hash + timestamp.hashCode();
+        if (timestamp != null)
+        	hash = 37 * hash + timestamp.hashCode();
         return hash;
       }
 
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -77,14 +77,14 @@ public abstract class Collector {
       public final List<String> labelNames;
       public final List<String> labelValues;  // Must have same length as labelNames.
       public final double value;
-      public final Long timestamp;
+      public final Long timestampMs;  // It's an epoch format with milliseconds value included (this field is subject to change).
 
-      public Sample(String name, List<String> labelNames, List<String> labelValues, double value, Long timestamp) {
+      public Sample(String name, List<String> labelNames, List<String> labelValues, double value, Long timestampMs) {
         this.name = name;
         this.labelNames = labelNames;
         this.labelValues = labelValues;
         this.value = value;
-        this.timestamp = timestamp;
+        this.timestampMs = timestampMs;
       }
 
       public Sample(String name, List<String> labelNames, List<String> labelValues, double value) {
@@ -98,17 +98,9 @@ public abstract class Collector {
         }
         Sample other = (Sample) obj;
 
-        boolean timestampEquals = true;
-        if ((timestamp != null && other.timestamp == null) || (timestamp == null && other.timestamp != null)) {
-          timestampEquals = false;
-        }
-        if (timestamp != null && other.timestamp != null) {
-          timestampEquals = other.timestamp.equals(timestamp);
-        }
-        
         return other.name.equals(name) && other.labelNames.equals(labelNames)
           && other.labelValues.equals(labelValues) && other.value == value
-          && timestampEquals;
+          && ((timestampMs == null && other.timestampMs == null) || (other.timestampMs != null) && (other.timestampMs.equals(timestampMs)));
       }
 
       @Override
@@ -119,8 +111,8 @@ public abstract class Collector {
         hash = 37 * hash + labelValues.hashCode();
         long d = Double.doubleToLongBits(value);
         hash = 37 * hash + (int)(d ^ (d >>> 32));
-        if (timestamp != null) {
-          hash = 37 * hash + timestamp.hashCode();
+        if (timestampMs != null) {
+          hash = 37 * hash + timestampMs.hashCode();
         }
         return hash;
       }
@@ -128,7 +120,7 @@ public abstract class Collector {
       @Override
       public String toString() {
         return "Name: " + name + " LabelNames: " + labelNames + " labelValues: " + labelValues +
-          " Value: " + value + " Timestamp: " + timestamp;
+          " Value: " + value + " TimestampMs: " + timestampMs;
       }
     }
   }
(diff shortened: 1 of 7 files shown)


brian-brazil:
This is breaking tests, and there should be unittests added for this.

It's be better to take a unboxed value.

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -91,8 +97,16 @@ public boolean equals(Object obj) {
           return false;
         }
         Sample other = (Sample) obj;
+        boolean timestampEquals = true;
```
pjfanning: This is the Java 7 java.utils.Objects.equals. This project maintains Java 6 compatibility but it might be tidier to paste in this method and use it compare all the member objects using this method. 

`static boolean equals(Object a, Object b) {
        return (a == b) || (a != null && a.equals(b));
    }`

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -103,13 +117,14 @@ public int hashCode() {
         hash = 37 * hash + labelValues.hashCode();
         long d = Double.doubleToLongBits(value);
         hash = 37 * hash + (int)(d ^ (d >>> 32));
+        hash = 37 * hash + timestamp.hashCode();
```
pjfanning: this needs to be wrapped in an `if (timestamp != null)` check

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -46,6 +46,10 @@ public static void write004(Writer writer, Enumeration<Collector.MetricFamilySam
         }
         writer.write(' ');
         writer.write(Collector.doubleToGoString(sample.value));
+        if (sample.timestamp != null){
+        	writer.write(' ');
```
pjfanning: could you remove the tab and use spaces for indents?

simpleclient/src/main/java/io/prometheus/client/Collector.java
efelcon: I don't totally understand what you say:
```
 boolean timestampEquals = true;
        if ((timestamp != null && other.timestamp == null) || (timestamp == null && other.timestamp != null)) {
        	timestampEquals = false;
        }
        if (timestamp != null && other.timestamp != null) {
        	timestampEquals = other.timestamp.equals(timestamp);
        }
```
The `equals()` used here is the same used in other parts of previously existing code, such as:
```
return other.name.equals(name) && other.labelNames.equals(labelNames)
          && other.labelValues.equals(labelValues) && other.value == value;
```

simpleclient/src/main/java/io/prometheus/client/Collector.java
efelcon: Thank you, I have changed it, but still not pushed.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
efelcon: Sorry, I didn't realize about it, I have changed it, but still not pushed.

simpleclient/src/main/java/io/prometheus/client/Collector.java
pjfanning: My suggestion is to based on readability - but it's just a suggestion.

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -103,13 +117,15 @@ public int hashCode() {
         hash = 37 * hash + labelValues.hashCode();
         long d = Double.doubleToLongBits(value);
         hash = 37 * hash + (int)(d ^ (d >>> 32));
+        if (timestamp != null)
+        	hash = 37 * hash + timestamp.hashCode();
```
pjfanning: could you remove the tab here?

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -91,8 +97,16 @@ public boolean equals(Object obj) {
...
+        if ((timestamp != null && other.timestamp == null) || (timestamp == null && other.timestamp != null)) {
+        	timestampEquals = false;
+        }
+        if (timestamp != null && other.timestamp != null) {
+        	timestampEquals = timestamp.equals(other.timestamp);
+        }
         return other.name.equals(name) && other.labelNames.equals(labelNames)
-          && other.labelValues.equals(labelValues) && other.value == value;
+          && other.labelValues.equals(labelValues) && other.value == value
+		  && timestampEquals;
```
pjfanning: also appears to be a tab here

pjfanning:
@efelcon would you have time to add the unit tests that Brian asked for?

efelcon:
I have the unit tests creation pending, now I have no time to complete it, but I expect to complete before weekend.

brian-brazil:
I'm at FOSDEM and unlikely to give this a proper look until Tuesday.

simpleclient/src/test/java/io/prometheus/client/CounterTest.java
```diff
@@ -82,4 +82,25 @@ public void testCollect() {
     assertEquals(mfsFixture, mfs.get(0));
   }
 
+  @Test
+  public void testCollectWithTimestamp() {
+  	Long now = System.currentTimeMillis();
```
pjfanning: another tab

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -77,12 +77,18 @@ public String toString() {
       public final List<String> labelNames;
       public final List<String> labelValues;  // Must have same length as labelNames.
       public final double value;
+      public final Long timestamp;
```
brian-brazil: timestampMs would be clearer.

Can you also add a note that this field is subject to change?

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -91,8 +97,18 @@ public boolean equals(Object obj) {
           return false;
         }
         Sample other = (Sample) obj;
+
+        boolean timestampEquals = true;
```
brian-brazil: This is a bit long, I think you can simplify a bit.

simpleclient/src/main/java/io/prometheus/client/Counter.java
```diff
@@ -108,6 +108,16 @@ protected Child newChild() {
...
   public static class Child {
     private final DoubleAdder value = new DoubleAdder();
+    /**
+     *  Milliseconds since epoch. Optional value.
+     */
+    private Long timestamp;
+    /**
+     * Set the optional value for time stamp in epoch format.
+     */
+    public void setTimestamp(Long timestamp) {
```
brian-brazil: Direct instrumentation has no need for timestamps, these need to be removed.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -46,6 +46,10 @@ public static void write004(Writer writer, Enumeration<Collector.MetricFamilySam
         }
         writer.write(' ');
         writer.write(Collector.doubleToGoString(sample.value));
+        if (sample.timestamp != null){
+          writer.write(' ');
+          writer.write(sample.timestamp.toString());
```
brian-brazil: This needs unittests

simpleclient/src/main/java/io/prometheus/client/Counter.java
efelcon: If I remove this, the _Unit Tests_ for `Counter` called `testCollectWithTimestamp` must be removed, because there will be no way for setting the _timestamp_ property (also happen the same for _Gauge_)). For me there is no problem so, are you agree to remove also the both _Unit Tests_ called `testCollectWithTimestamp`?

simpleclient/src/main/java/io/prometheus/client/Collector.java
efelcon: I'm agree, this has been changed.

simpleclient/src/main/java/io/prometheus/client/Collector.java
efelcon: Done. Now has been reduced to one line.

simpleclient/src/main/java/io/prometheus/client/Counter.java
brian-brazil: Timestamps only work via custom collectors, so tests need to implement one of those.

efelcon: Ok, now I think I understand what you mean. But I'm afraid I have commited several changes. Tomorrow I will work on this.

efelcon:
I have added a _Unit Test_ for TextFormat java class.

Please confirm me if you want me to remove the methods and _Unit Tests_ added in `Counter`, `CounteTest`, `Gauge` and `GaugeTest` classes.

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/TextFormatTest.java
```diff
@@ -53,6 +68,42 @@ public void testCounterOutput() throws IOException {
...
+    noLabels.inc();
+    noLabels.setTimestamp(now);
+    TextFormat.write004(writer, registry.metricFamilySamples());
+    assertEquals("# HELP nolabels help\n"
+                 + "# TYPE nolabels counter\n"
+                 + "nolabels 1.0 "+ now +"\n", writer.toString());
+  }
+
+  @Test
+  public void testMetricOutputWithTimetamp() throws IOException {
```
brian-brazil: stamp

efelcon:
I have removed all code you request to remove, so I think that all is ready for your check.

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/TextFormatTest.java
```diff
@@ -54,6 +57,30 @@ public void testCounterOutput() throws IOException {
...
+        mfs.add(new MetricFamilySamples("nolabels", Collector.Type.UNTYPED, "help", samples));
+        return mfs;
+      }
+    }
+    
+    new CustomCollector().register(registry);
+    TextFormat.write004(writer, registry.metricFamilySamples());
+    assertEquals("# HELP nolabels help\n"
+                 + "# TYPE nolabels untyped\n"
+                 + "nolabels 1.0 "+ now +"\n", writer.toString());
```
brian-brazil: I'd prefer this was a literal string rather than dynamic, as the output here depends on Java's implicit conversion to string.

brian-brazil:
Thanks!

PR #339, opened 2018-02-07
Make gitignore file understand VS Code

matzew:
For Visusal Studio Code, the `.gitignore` file needs some help


matzew:
@brian-brazil do you mind taking a look at this minor fix ?

brian-brazil:
The general stance of the project is that you should add these to your global .gitignore, rather than having to do it for every project that you touch.

matzew:
fair point

PR #343, opened 2018-02-16
Add host label to the MetricFilter

fjavierm:
Add a new label `host` to the metrics generated by the `MetricFilter` to be able to create more compressive report with the generated metrics.

@brian-brazil


brian-brazil:
This is a breaking change, and I'd expect most users would not be serving different content on different hosts.

I'd suggest handling this via direct instrumentation in your own code.

PR #346, opened 2018-02-22
Update dropwizard metrics

ChristopherDavenport:


brian-brazil:
I'd like to make sure that we don't accidentally break compatibility with older versions. Is there some particular reason you want to upgrade?

ChristopherDavenport:
I have a diamond dependency personally with http4s which uses the upgraded and this which wants the older. If it has to wait till a next release that's fine - I'm not very familiar with your library except as a dependency. It's blocking some folks I know from upgrading to the newer version so I figured I would open a PR to open the conversation.

checketts:
@ChristopherDavenport Glancing at the release notes, it appears Dropwizard metrics 4 targets Java 8. Could you verify that?

brian-brazil:
This has gone stale, so closing for now.

PR #348, opened 2018-02-26
Annotations for interfaces

auduntorp:
Add metrics to classes implementing an interface by decorating them with a PrometheusMonitor proxy


brian-brazil:
This is interesting, however I think it'd be better served by being in another repo rather than the core client as it's making some strong assumptions about usage.

Your metrics naming also a bit off, duration summary's should have a unit suffix of _seconds and no _total suffix. The example uses of the namespace field in tests also are not in line usual naming practices, which is a tad worrying.

auduntorp:
Thank you for the feedback. I will probably move this to a standalone repo then.

PR #352, opened 2018-03-14
Handle null values gracefully in DropwizardExports

henri-tremblay:


simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
```diff
@@ -5,6 +5,7 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
```
brian-brazil: This isn't used.

henri-tremblay: Right

henri-tremblay:
Unit test added

brian-brazil:
Thanks!

PR #354, opened 2018-03-15
Add version-range to the guava import

### LINKED ISSUE #353: Increase Guava version range
Since Guava 21 Guava promised to no longer make breaking changes. 
I think the simpleclient_guava is compatible with Guava 18.0 onward - so the version import should reflect that. At least an OSGi version range from 18 to 24 would be good. Ideally the version range is extended into the future to accommodate future releases. Not sure what version makes sense but probably at least 26 could be done (or directly 30)

svogt:
Fixes #353

Signed-off-by: Sascha Vogt <s.vogt@seeburger.de>

### DIFF
--- a/simpleclient_guava/pom.xml
+++ b/simpleclient_guava/pom.xml
@@ -33,6 +33,20 @@
         </developer>
     </developers>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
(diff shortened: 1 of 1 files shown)


brian-brazil:
As I said in the issue, I'd prefer not to add packaging system specific tweaks. I'm open to removing the felix stuff wholesale though.

PR #356, opened 2018-03-21
Fix jvm_gc_collection metrics naming

guillaumebort:
- GC collections count has nothing to do with "seconds".
- As GC MBean metrics are already aggregated I don't think we can use a `SummaryMetric`.


brian-brazil:
The naming is correct, this is how Summary metrics work.

guillaumebort:
So what does `jvm_gc_collection_seconds_count` mean? Because this is currently the number of garbage collections and it has nothing to do with time. So `seconds` does not mean anything in this name.

brian-brazil:
The summary metric as a whole is about how long the GCs take.

Issue #351, opened 2018-03-08
Extra comma for metrics with labels


fraisse:
https://github.com/prometheus/client_java/blob/0deffbc4cf06513ca1c75b00936db3956ca1fb1c/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java#L43

It seems there will systematically be a comma at then of the labels, no?

brian-brazil:
Dupe of #210

Issue #321, opened 2017-12-13
TextFormat write004


nicopoirier:
Hello

In the method write004 of class TextFormat, prometheus client has a small bug. The method generate an extra comma in the labels after the last attribute.

https://github.com/prometheus/client_java/blob/1a8bf7da57402f444861c32abc9e0c7cd313748e/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java#L43

Regards

brian-brazil:
This behaviour is correct.

Issue #262, opened 2017-06-14
Line 43 of TextFormat is wrong - always appends a comma to labelNames


alexp700:
As I see it, it always adds a comma to the list of labelNames, so this isn't parsed correctly by prometheus, or am I doing it wrong?

Thanks,

Alex

checketts:
That is intentional.

brian-brazil:
Make sure you're using the latest version of the python parser or prometheus, whichever it is you're using.

alexp700:
I am using 1.7 - so almost the latest (when I started).

yuanman:
my prometheus version is 2.0.0, can not scrape the metric. the log is "no token found". 
Default lablename is this format? have comma in the end?  like this  --> info_size{type="o",} 679.0

checketts:
@yuanman The trailing comma is expected and not . I suspect you have a metric name with a hyphen in it. That is not allowed since it can be mistaken for a subtraction sign.

Issue #210, opened 2017-03-08
TextFormat add extra comma when sample has tags


ezraroi:
When sample in `MetricFamilySamples` has tags... the writer adds extra comma at the end of the tags output which break prometheus collection.

Here is the problematic code:
```
for (Collector.MetricFamilySamples.Sample sample: metricFamilySamples.samples) {
        writer.write(sample.name);
        if (sample.labelNames.size() > 0) {
          writer.write('{');
          for (int i = 0; i < sample.labelNames.size(); ++i) {
            writer.write(sample.labelNames.get(i));
            writer.write("=\"");
            writeEscapedLabelValue(writer, sample.labelValues.get(i));
            writer.write("\",");
          }
          writer.write('}');
        }
        writer.write(' ');
        writer.write(Collector.doubleToGoString(sample.value));
        writer.write('\n');
      }
```

This is how it looks:
```
# HELP rest rest
# TYPE rest gauge
rest{method="PUT",statistic="min",status="404",uri="api_jobs_-fullId-_-iteration-",} 2.0
```

brian-brazil:
This is valid output.

ezraroi:
ok.. thanks.. found the problem.. i have few entries with the same help

ezraroi:
looks like this is the problem i have:
```
# HELP rest rest
# TYPE rest gauge
rest{method="PUT",statistic="min",status="404",uri="api_jobs_-fullId-_-iteration-",} 2.0
# HELP rest rest
# TYPE rest gauge
rest{method="POST",statistic="95thPercentile",status="200",uri="api_jobs_-fullId-_-iteration-",} 126.0
```

ezraroi:
2 help blocks are the same.. right?

brian-brazil:
That's not permitted by the format, and will not be accepted by Prometheus. You need at most one help per metric.

ezraroi:
yes... thanks... just read it

PR #357, opened 2018-03-22
Remove trailing comma from label values

marcust:
Looks untidy. I was just wondering if I broke something because of that, so I decided to fix that

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -40,7 +40,10 @@ public class TextFormat {
             writer.write(sample.labelNames.get(i));
             writer.write("=\"");
             writeEscapedLabelValue(writer, sample.labelValues.get(i));
-            writer.write("\",");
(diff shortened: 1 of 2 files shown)


brian-brazil:
This is permitted by the spec.

checketts:
@marcust See #210  #351 #321 #262

marcust:
Well if it is permitted is not the point, no? The point is that I apparently was the fifth person that got confused by that at looked for a bug that wasn't there. Kudos to Prometheus to accepting that as input, but why would you write something that looks odd, even if it is permitted by the spec?

himanshub16:
I get that trailing comma in last label is not an error, and is permitted by spec.
However, I have tested `python`, and `golang` clients, and they don't have such behaviour.
Also, it looks weird to have something of a JSON-like structure and have a trailing comma to the last item.

Thoughts?

adigupta13:
Completely agree with @marcust . The fact that everyone is confused with this behaviour and is not consistent with other platform clients like python and golang should be enough reason to fix it even if it is supported in the spec. And no acknowledgment by @brian-brazil since 3 years doesn't look good.

PR #358, opened 2018-03-22
Allow extra labels in DropwizardExporter

marcust:
I have the problem that I'm exporting multiple metric registries and I don't know which output belongs to what, and I can't filter on that in prometheus, so I added static labels.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -4,6 +4,7 @@ import com.codahale.metrics.*;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.SortedMap;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


brian-brazil:
This doesn't make sense at all semantically. This is something for metric names not labels, and I don't want to add either as it would be heavily abused.

I'd suggest a) fixing your metrics to have distinct names b) switching to using this client for instrumentation.

marcust:
The metrics have distinct names, it is just that the Cassandra driver decided to export something that does not have the name Cassandra in it as Dropwizard metrics. And as this is the compatibility layer I think it belongs here for now, obviously I would prefer the Cassandra guys to have a better interface point for metrics. I thought the alternative of adding a prefix is even uglier.

brian-brazil:
You need to be able to aggregate across labels, so this is not an appropriate use of instrumentation labels. 

How is this usually handled for those using only dropwizard metrics?

marcust:
Not sure how that is usually used. I guess it is equally confusing. But I guess changing the metrics names is not really an option, because I guess a lot of people depend on them and changing names is usually something one would not expect.

brian-brazil:
Metric names change all the time. It seems here that the real problem is elsewhere and I don't think it's a good idea to try to paper over the problem at this layer. This is a bug report for Cassandra.

PR #360, opened 2018-03-26
Fix null pointer in unregister

marcust:
If there is no such collector it nullpointers. As the intention of remove is to
remove the given collector from the registry and if there is no such collector
this intention is fulfilled, this should not nullpointer.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -67,6 +67,9 @@ public class CollectorRegistry {
   public void unregister(Collector m) {
     synchronized (collectorsToNames) {
       List<String> names = collectorsToNames.remove(m);
+      if (names == null) {
+        return;
+      }
       for (String name : names) {
         namesToCollectors.remove(name);
       }


brian-brazil:
Why are you removing a collector that has already been removed?

marcust:
Maybe it has never been registered, I don't know yet. But also there is no way to look it up if it is registered. And there is no thread safe way to do so anyways. 
In any case it should not NullPointer

brian-brazil:
This is not the way that you are meant to use a Prometheus client library. You're meant to register at initialisation and that's it.

marcust:
Initialisation of what?

In my case I have something that only becomes available at a certain point in the application life cycle. So I can't do that, as much as I'd love to. And because of other complexities I currently have no way to ensure that it becomes only available once during the application lifecycle. At least that is currently too much work for what I try to achieve and will take a couple of month to be rolled out over all our services. 

What is the remove method for if that is not a use-case?

brian-brazil:
The application.

> What is the remove method for if that is not a use-case?

It comes up sometimes in unittesting. Once registered a metric should not be removed otherwise.

brian-brazil:
It's be design that you can't look it up, see https://www.robustperception.io/label-lookups-and-the-child/

This sounds like something you need to manage in your own code.

PR #363, opened 2018-03-29
Add DBCP DataSource collector

edumucelli:
Hi @brian-brazil, I needed this DBCP collector and thought it would be a good thing to contribute back, as I am currently using the Guava collector from this lib. As a first PR on this project, I may have missed something. Comments are welcome.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,7 @@
         <module>simpleclient_jetty</module>
         <module>simpleclient_jetty_jdk8</module>
         <module>simpleclient_vertx</module>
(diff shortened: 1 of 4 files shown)


brian-brazil:
I'm not planning on accepting more modules, so I'd suggest hosting this in your own repository.

I'd also suggest removing the _num from the metric names as it doesn't add anything, and moving the _min/_max to the end of the metric names.

edumucelli:
Alright. Although I think it makes no sense to have another repo just for one specific collector.

PR #365, opened 2018-04-07
Made Collector.Builder fields protected instead of package-private.

mpartel:
So client code can more easily define its own collector builders.

Ping @brian-brazil per contributing guidelines.


brian-brazil:
This class is explicitly internal and should not be depended on directly. Why do you want this?

mpartel:
Ah, I failed to read the comment on the surrounding class. Ok, closing.

PR #366, opened 2018-04-07
Added method Histogram.observeMany

mpartel:
This is occasionally useful when events are observed in bulk, and it enables (hackily) exporting existing histogram data to Prometheus.

Ping @brian-brazil per contributing guidelines.


brian-brazil:
If you want to do this you can call the existing function several times. For the expected direct instrumentation uses cases you would only ever have one event.

mpartel:
Seems wasteful to do it potentially thousands of times, but ok, I'll either do that or find some other workaround. I admit I don't have a "nice" use-case for this, just the hacky one I mentioned.

PR #367, opened 2018-04-07
added Gauge.Builder.fromCallback method

yarosla:
@brian-brazil 

It is quite common to create Gauges for externally calculated values. Eg. cache or queue sizes, etc.

I've added `Gauge.Builder.fromCallback()` method to enhance usability for this use case. Instead of using `setChild()` construct, one can write:
```java
    Gauge.build().name("queue_size").help("Current queue length.")
                   .fromCallback(() -> (double) queue.size())
                   .register();
```
I've also added corresponding unit tests.

I can also amend the documentation.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Gauge.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Gauge.java
@@ -1,10 +1,10 @@
 package io.prometheus.client;
 
 import java.io.Closeable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * Gauge metric, to report instantaneous values.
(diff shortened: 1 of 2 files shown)


brian-brazil:
This feature is only really a shortcut for those with simple requirements, and I don't believe what you're proposing saves any significant code. If you want to start having labels then you should really be using a custom collector as that's a non-trivial use case.

yarosla:
I am proposing this because we already faced with this issue in our project (which is huge). We want to monitor cache and buffer sizes (which could be simple `Map`s) and every time we have to write:

```java
    Gauge.build().name("queue_size").help("Current queue length.").create()
        .setChild(new Gauge.Child() {
           public double get() {
             return queue.size();
           }
         }).register();
```

The problem is that this cannot be reduced to simple lambda, so it makes our code bloated (+3 meaningless lines per gauge). Sometimes we have two or three metrics like this in a class.

We believe metrics code should be as unobtrusive as possible.

We have three way to solve this:
- write our own wrapper that creates gauges from lambdas and then import it everywhere (not desirable);
- migrate to Micrometer, which implements all the lambda stuff already (although it has other drawbacks);
- persuade you to improve usability of your library (I thought this is best).

brian-brazil:
That code is the same length as the example in what you're proposing. It doesn't make sense to have this as a first class thing in the builder, callbacks only really make sense on a child basis. setChild is not intended to be first-class citizen either as those coming from other monitoring systems with different designed would use them where they should really instead be doing direct instrumentation.

Have you considered a utility class that makes creating the child easier? Or a utility that takes a queue object and a name and produces a metric?

yarosla:
New code:
```java
    Gauge.build().name("queue_size").help("Current queue length.")
                   .fromCallback(() -> (double) queue.size())
                   .register();
```
is two times shorter than:
```java
    Gauge.build().name("queue_size").help("Current queue length.").create()
        .setChild(new Gauge.Child() {
           public double get() {
             return queue.size();
           }
         }).register();
```

What is even more important it is semantically readable. `Child` is something really internal to your library and should not be visible to user.

Utility class is the first way of solving this, as I described earlier. But that would mean I need to make wrappers for all your metric types so that they look consistent. Which leads to the second way: why not use Micrometer, which has already did the wrapping.

yarosla:
I think collecting result from lambda (`Supplier` or `Callable`) is very common use case for direct instrumentation. Perhaps this could be done in other class, not `Gauge`, but that piece is definitely missing in your toolkit.

I can help adding that but need some guidance from you.

yarosla:
I think client libraries should provide convenient tools for common use cases.

In many cases it is not reasonable to `inc` and `dec` state, while there exists a callback providing ready to use "snapshot of state". For example, size of cache, queue, buffer or pool.

The library can include generic collector for such use cases, which will make life easier for many developers.

I agree that existing `Gauge` class is not best fit. And suggest writing a new one for the library.

yarosla:
It could be something like this:
```java
    Meter.build().name("queue_size").help("Current queue length.")
                   .fromCallback(() -> (double) queue.size())
                   .register();
```

It can also include methods like `.forMapSize(map)`, `.forCollectionSize(list)`, etc.

Need to decide how to deal with labels. One way:
```java
    Meter.build().name("queue_size").help("Current queue length.")
                   .labelNames("stage")
                   .fromCallback(asList("a"), () -> (double) queue1.size())
                   .fromCallback(asList("b"), () -> (double) queue2.size())
                   .register();
```
What do you think?

brian-brazil:
Meter means something else in other instrumentation systems, it'd be a Counter in Prometheus so that's a bad choice of term.

I'm not sure it's a good idea to support labels with such a thing, that's getting into territory where you really need a custom collector. What happens when a new queue is added for example?

yarosla:
It would be up to user to decide whether to group several queues under same metric name (with different labels) or create different metric name for every queue (not using labels at all). I think such a flexibility would be nice to have.

brian-brazil:
Whether it belongs as one metric or not is an important semantic question. If they are to share a metric then that should probably be implemented within the queue class itself rather than be something the user has to setup. That also implies there's some form of queue naming going on. Put another way, two random queues probably shouldn't be sharing a metric name.

Also, queue.size is not threadsafe.

yarosla:
Currently I do not have real cases for using labels. So having such a collector without labels might be sufficient.

I think there are implementations of Queue that are thread-safe (eg. BlockingQueues or ConcurrentLinkedQueue). But that was just for example. My real cases deal with ConcurrentMaps or custom classes with `size()` method, which is thread safe.

It should be stated in documentation that user is responsible for thread safety of supplied callable.

brian-brazil:
This is stale, and as stated above doesn't belong in the core library.

PR #368, opened 2018-04-07
allow Histograms and Summaries to time Callables

yarosla:
Another usability enhancement allowing instead of:

```java
    String v;
    try (Summary.Timer timer = summary.startTimer()) {
      v = someMethod();
    }
```

to write:

```java
    String v = summary.time(this::someMethod);
```

This is especially useful when programming (reactive) streams.

Trying to make metric code as unobtrusive as possible.

BTW Micrometer already does that.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Histogram.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Histogram.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
 
 /**
  * Histogram metric, to track distributions of events.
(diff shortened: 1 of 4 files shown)


simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -207,6 +208,24 @@ public double time(Runnable timeable) {
       return elapsed;
     }
 
+    /**
+     * Executes callable code (i.e. a Java 8 Lambda) and observes a duration of how long it took to run.
```
brian-brazil: e.g.

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -207,6 +208,24 @@ public double time(Runnable timeable) {
...
+     *
+     * @param timeable Code that is being timed
+     * @return Result returned by callable.
+     */
+    public <E> E time(Callable<E> timeable) {
```
brian-brazil: Gauge also has a time() method

simpleclient/src/main/java/io/prometheus/client/Histogram.java
```diff
@@ -207,6 +208,24 @@ public double time(Runnable timeable) {
...
+      Timer timer = startTimer();
+
+      try {
+        return timeable.call();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
brian-brazil: Why are you wrapping this exception rather than just letting it propagate naturally?

simpleclient/src/main/java/io/prometheus/client/Histogram.java
yarosla: I see. Then on old method this should also be corrected, as I copied text from there.

brian-brazil: Yes, I must have missed that back when that was added.

yarosla:
I've fixed all the comments. Don't know how to deal with this DCO thing.

brian-brazil:
The easiest thing to do is to squash the commits with rebase, and make sure the remaining commit has the signoff.

yarosla:
I've created new PR #378 from squashed branch.

PR #369, opened 2018-04-09
Add Collector integration with Kafka client metrics

liff:
A small proposed addition of integration with Kafka clients library.

This integrates the Prometheus collector with Kafka metrics via kafka-clients `MetricsReporter`.

The kafka-clients version in this is somewhat conservatively 0.11 but could be bumped to 1.0 or 1.1 if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,27 @@ new QueuedThreadPoolStatisticsCollector()
     .register();
 ```
 
+### Kafka
+
+There is a collector for recording metrics from Kafka clients. Use the `metric.reporters` to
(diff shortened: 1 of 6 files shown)


brian-brazil:
I'm not accepting more modules into this repository, so I'd suggest keeping this in your own github org.

This approach looks fine generally. The one note I'd have is that you're presuming that everything is a Gauge - this is not the case as the Count and possibly Total types are Counters. If you don't want to worry about this you can use Untyped instead.

liff:
Ah ok. Thanks and sorry for the trouble.

PR #370, opened 2018-04-10
add spectator client

zhengyangyong:
Signed-off-by: zhengyangyong <yangyong.zheng@huawei.com>

add exporter for [netflix spectator](https://github.com/Netflix/spectator)


brian-brazil:
I'm afraid that I'm not accepting new modules in this repo, I'd suggest maintaining it in your own org.

Your code looks fine, one thing is that label names may also need validation.

PR #375, opened 2018-04-13
Upgrade Mockito to build on JDK10

moio:
@brian-brazil this is a trivial fix to fix building the simpleclient_hibernate in JDK 10.

Current version of Mockito relies on an older version of bytebuddy that does not work correctly with the latest Java release.


brian-brazil:
Thanks!

PR #377, opened 2018-04-19
Add Dropwizard metric reporter to pushgateway

shiroari:
Hi,

This PR adds scheduled reporter for Dropwizard metrics through push-gateway as an alternative to pull mechanism. It can be useful for projects like Spark to collect data from dynamic set of executors. 

Originally I didn't think about adding reporter to this projects, but since `DropwizardExports` class does not really open for extension, instead of just making its methods public and implementing reporter in another project, I decided to add it here. I think it fits good. It the same time it might be a good idea to make `fromGauge/fromHistogram/fromTimer/fromMeter/fromCounter` methods public static and move to utility class.

@brian-brazil please review and say if the feature fits this project.

### DIFF
--- a/simpleclient_dropwizard/pom.xml
+++ b/simpleclient_dropwizard/pom.xml
@@ -17,6 +17,10 @@
         Collector of data from Dropwizard metrics library.
     </description>
 
+    <properties>
+        <dropwizard.metrics.version>3.1.2</dropwizard.metrics.version>
+    </properties>
+
     <licenses>
         <license>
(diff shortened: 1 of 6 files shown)


brian-brazil:
This is a violation of the separation of concerns between collectors and bridges. I think you've made this quite a bit more complicated than this needs to be, it should only be a few lines of code on top of the interfaces that are already there.

At a high level this is not an appropriate use of the pushgateway, see https://prometheus.io/docs/practices/pushing/

PR #378, opened 2018-04-21
allow Histograms and Summaries to time Callables (2)

yarosla:
@brian-brazil Recreated #368 from squashed branch + signed-off

Another usability enhancement allowing instead of:

    String v;
    try (Summary.Timer timer = summary.startTimer()) {
      v = someMethod();
    }
to write:

    String v = summary.time(this::someMethod);
This is especially useful when programming (reactive) streams.

Trying to make metric code as unobtrusive as possible.

BTW Micrometer already does that.


brian-brazil:
Thanks!

PR #382, opened 2018-05-22
More sophisticated path to label mapping configuration in MetricsFilter

greenjoe:
Specifying number of path components that is taken into account may not be enough for some projects where number of components in different endpoints differs.
This commit allows passing a custom path to label mapper from the java code into the MetricsFilter constructor.


brian-brazil:
Thanks for sharing. This filter is meant to get you up and running quickly, if you have more complex needs you should probably use direct instrumentation.

greenjoe:
Ah, we don't use AOP in our code. Anyway, I can see your point and I will just create a completely new filter then.

PR #383, opened 2018-05-26
EhCache 2 implementation 

luan-cestari:
Hi,

I created the following implementation to extract some metrics for applications using EhCache 2 . I mentioned the issue #203 as I saw a mention about EhCache there

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,7 +622,7 @@ class YourCustomCollector extends Collector {
 static final YourCustomCollector requests = new YourCustomCollector().register()
 ```
 
-`SummaryMetricFamily` and `HistogramMetricFamily` work similarly.
+`SummaryMetricFamily` works similarly.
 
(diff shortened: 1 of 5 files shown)


brian-brazil:
Thanks for sharing this. I'm afraid we're not accepting new modules, so I'd suggest maintaining it in your own repo. We can add a link to it from the docs.

I would suggest a read of https://prometheus.io/docs/instrumenting/writing_exporters/ in order to improve your metrics. In particular counters and only counters should end in _total, prefer exposing raw numerators/denominators rather than ratios, metrics should have units, and don't expose per-second values - rather expose the counter and let it be calculated in Prometheus. The latency_avg are unfortunately as the library isn't providing great stats, but I'd drop the "avg" (it's implied) and mention in the help over what time frame it is calculated.

PR #387, opened 2018-06-13
Ability to add labels to drop wizard metrics

dbathgate:
This adds the ability to create global labels for Dropwizard metrics

```
DropwizardExports exports = new DropwizardExports(metricRegistry);
exports.addLabel("env", "prod");
exports.addLabel("version", "1.2");
CollectorRegistry.defaultRegistry.register(exports);
```

Signed-off-by: Darren Bathgate <dbathgate@kenzan.com>


brian-brazil:
This is an anti-pattern in Prometheus, you should be using target labels on the Prometheus end. In addition version numbers are usually best handled via an info-style metric rather than a label across an entire target.

dbathgate:
Thanks for the feedback @brian-brazil . I am curious to know more about this info-style metric.

What I am after is the ability to monitor 2 different versions of the application metric for a canary analysis. 

What would you recommend?

brian-brazil:
https://www.robustperception.io/exposing-the-software-version-to-prometheus/

dbathgate:
Thanks! This is helpful.

PR #388, opened 2018-06-15
Added per-query metrics in HibernateStatisticsCollector

thombergs:
This PR extends the `HibernateStatisticsCollector` so that it collects metrics per query from Hibernate's `QueryStatistics` object. 

Collecting metrics on a per-query basis is disabled by default and must be enabled by calling `enablePerQueryMetrics()`. 

We used the so created per-query metrics in a test system in combination with Grafana and successfully tracked down a database issue with long-lasting database queries.

I thought you might find it useful as well, so I created this PR :).

### DIFF
--- a/simpleclient_hibernate/pom.xml
+++ b/simpleclient_hibernate/pom.xml
@@ -45,7 +45,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
-            <version>5.3.1.Final</version>
+            <version>5.2.0.Final</version>
             <scope>provided</scope>
         </dependency>
 
--- a/simpleclient_hibernate/pom.xml
+++ b/simpleclient_hibernate/pom.xml
@@ -45,7 +45,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
-            <version>4.2.0.Final</version>
+            <version>5.3.1.Final</version>
             <scope>provided</scope>
         </dependency>
 
--- a/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
+++ b/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
@@ -557,6 +557,16 @@ public class HibernateStatisticsCollector extends Collector {
                         .getExecutionAvgTime();
               }
             }
+        ),
+        createGaugeForQuery("hibernate_per_query_execution_total_time",
+            "Accumulated execution time of query in milliseconds (getExecutionTotalTime)",
+            new ValueProviderPerQuery() {
+              @Override
+              public double getValue(Statistics statistics, String query) {
+                return statistics.getQueryStatistics(query)
+                    .getExecutionTotalTime();
+              }
+            }
         )
     ));
 
--- a/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
+++ b/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
@@ -505,7 +505,7 @@ public class HibernateStatisticsCollector extends Collector {
               }
             }
         ),
-        createCounterForQuery("hibernate_per_query_executions_total",
+        createCounterForQuery("hibernate_per_query_execution_total",
                 "Global number of executions for query (getExecutionCount)",
             new ValueProviderPerQuery() {
               @Override
--- a/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
+++ b/simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
@@ -528,33 +528,33 @@ public class HibernateStatisticsCollector extends Collector {
               }
             }
         ),
-        createGaugeForQuery("hibernate_per_query_execution_min_time_ms",
-                "Minimum execution time of query in milliseconds (getExecutionMinTime)",
+        createGaugeForQuery("hibernate_per_query_execution_min_time",
+                "Minimum execution time of query in seconds (based on getExecutionMinTime)",
             new ValueProviderPerQuery() {
               @Override
               public double getValue(Statistics statistics, String query) {
-                return statistics.getQueryStatistics(query)
-                        .getExecutionMinTime();
+                return toSeconds(statistics.getQueryStatistics(query)
+                        .getExecutionMinTime());
               }
             }
         ),
-        createGaugeForQuery("hibernate_per_query_execution_max_time_ms",
-                "Maximum execution time of query in milliseconds (getExecutionMaxTime)",
+        createGaugeForQuery("hibernate_per_query_execution_max_time",
+                "Maximum execution time of query in seconds (based on getExecutionMaxTime)",
             new ValueProviderPerQuery() {
               @Override
               public double getValue(Statistics statistics, String query) {
-                return statistics.getQueryStatistics(query)
-                        .getExecutionMaxTime();
+                return toSeconds(statistics.getQueryStatistics(query)
+                        .getExecutionMaxTime());
               }
             }
         ),
-        createGaugeForQuery("hibernate_per_query_execution_time_ms",
-            "Accumulated execution time of query in milliseconds (getExecutionTotalTime)",
+        createGaugeForQuery("hibernate_per_query_execution_time",
+            "Accumulated execution time of query in seconds (based on getExecutionTotalTime)",
             new ValueProviderPerQuery() {
               @Override
               public double getValue(Statistics statistics, String query) {
-                return statistics.getQueryStatistics(query)
-                    .getExecutionTotalTime();
+                return toSeconds(statistics.getQueryStatistics(query)
+                    .getExecutionTotalTime());
               }
             }
         )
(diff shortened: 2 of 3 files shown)


simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -448,25 +470,172 @@ public double getValue(Statistics statistics) {
...
+            new ValueProviderPerQuery() {
+              @Override
+              public double getValue(Statistics statistics, String query) {
+                return statistics.getQueryStatistics(query)
+                        .getExecutionRowCount();
+              }
+            }
+        ),
+        createGaugeForQuery("hibernate_per_query_execution_min_time",
+                "Minimum execution time of query (getExecutionMinTime)",
```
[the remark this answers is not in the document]
thombergs: I would have liked counters better, too, but Hibernate's [QueryStatistics](https://docs.jboss.org/hibernate/core/3.3/api/org/hibernate/stat/QueryStatistics.html) only provides the current average, max and min values for execution times of single queries. 

Sadly, Hibernate provides no hook to count things...

However, for the use case of making especially long-running database queries transparent, the "max" metric is still helpful. We had the queries listed in a table in Grafana, sorted by their "max" value, which was quite enlightening :)

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -74,6 +83,16 @@ public HibernateStatisticsCollector add(SessionFactory sessionFactory, String na
     return this;
   }
 
+  /**
+   * Enables collection of per-query metrics. Produces a lot of monitoring data, so use with caution.
```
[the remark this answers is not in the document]
thombergs: I modified the javadoc and added an example query and a sentence about the query parameters not being replaced (not sure if you meant the query parameters with "constants").

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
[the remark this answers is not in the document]
brian-brazil: In this situation we usually take only the average, and document what is actually is. How is this average implemented?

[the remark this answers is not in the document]
thombergs: The max, avg and min times are measured in milliseconds. Max and min are held as local variables over the lifetime of the application. The avg is calculated on access by dividing the accumulated time by the number of query executions (both also held as local variables).

[This](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/stat/internal/QueryStatisticsImpl.java) is the source.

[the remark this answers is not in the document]
brian-brazil: If you expose totalExecutionTime then you can calculate average on the Prometheus side, as with getExecutionCount it makes a Summary.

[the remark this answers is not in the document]
thombergs: I added the totalExecutionTime metric. I had to bump up the Hibernate dependency, though, because this metric has only been added with Hibernate 5 (which is three years old by now).

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -74,6 +83,19 @@ public HibernateStatisticsCollector add(SessionFactory sessionFactory, String na
...
+   * <p>
+   * Per-query metrics have a label "query" with the actual HQL query as value. Parameters within
+   * the query will not be replaced (example: {@code select u from User u where id=?}).
```
brian-brazil: s/not//

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -448,25 +473,182 @@ public double getValue(Statistics statistics) {
...
+            new ValueProviderPerQuery() {
+              @Override
+              public double getValue(Statistics statistics, String query) {
+                return statistics.getQueryStatistics(query)
+                        .getExecutionRowCount();
+              }
+            }
+        ),
+        createGaugeForQuery("hibernate_per_query_execution_min_time",
+                "Minimum execution time of query in milliseconds (getExecutionMinTime)",
```
brian-brazil: Seconds are the base unit, and the metric name should also include the unit

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -448,25 +473,182 @@ public double getValue(Statistics statistics) {
...
+            new ValueProviderPerQuery() {
+              @Override
+              public double getValue(Statistics statistics, String query) {
+                return statistics.getQueryStatistics(query)
+                        .getExecutionMaxTime();
+              }
+            }
+        ),
+        createGaugeForQuery("hibernate_per_query_execution_avg_time",
+                "Average execution time of query in milliseconds (getExecutionAvgTime)",
```
brian-brazil: Remove this, it is no longer needed

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -448,25 +473,182 @@ public double getValue(Statistics statistics) {
...
+            new ValueProviderPerQuery() {
+              @Override
+              public double getValue(Statistics statistics, String query) {
+                return statistics.getQueryStatistics(query)
+                        .getExecutionAvgTime();
+              }
+            }
+        ),
+        createGaugeForQuery("hibernate_per_query_execution_total_time",
```
brian-brazil: "total" doesn't add anything here

simpleclient_hibernate/pom.xml
```diff
@@ -45,7 +45,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
-            <version>4.2.0.Final</version>
+            <version>5.3.1.Final</version>
```
brian-brazil: 5.3.1 was only a bit over a month ago, I doubt many people are using it. What's the oldest version with this feature, and how many people are using at least that version?

thombergs: Changed the dependency to 5.2.0 where the new metric was added (2 years old). If we're to believe https://mvnrepository.com/artifact/org.hibernate/hibernate-core the 5.2.x versions are currently most widely used.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
thombergs: The sentence was ambiguous. Changed the comment to make it clear.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
thombergs: Added "_ms" to all three gauge metrics.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
thombergs: Removed the avg metric.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
thombergs: Removed the "total" from the metric name.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
brian-brazil: You should convert to seconds, as that's standard in Prometheus

thombergs: @brian-brazil milliseconds are being converted to seconds now. Also added "seconds" to the metric names.

simpleclient_hibernate/src/main/java/io/prometheus/client/hibernate/HibernateStatisticsCollector.java
```diff
@@ -448,25 +473,176 @@ public double getValue(Statistics statistics) {
...
+                "Maximum execution time of query in seconds (based on getExecutionMaxTime)",
+            new ValueProviderPerQuery() {
+              @Override
+              public double getValue(Statistics statistics, String query) {
+                return toSeconds(statistics.getQueryStatistics(query)
+                        .getExecutionMaxTime());
+              }
+            }
+        ),
+        createGaugeForQuery("hibernate_per_query_execution_seconds",
```
brian-brazil: This is a Counter, and should end in _total

thombergs: Ah yes, missed that. Thanks.

The metric now ends in "_total" and is a counter instead of a gauge.

brian-brazil:
Thanks!

thombergs:
@brian-brazil  thank you for the thorough review :)

PR #389, opened 2018-06-19
Exporter for EclipseLink library

VitaNuova:
This exporter follows the same structure as the Hibernate exporter, but exports the metrics available in EclipseLink. I had to write it for my project and thought I might try to contribute it. @brian-brazil I would be happy about any feedback.

### DIFF
--- a/simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
+++ b/simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
@@ -4,16 +4,13 @@ import io.prometheus.client.Collector;
 import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.CounterMetricFamily;
 import io.prometheus.client.GaugeMetricFamily;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.concurrent.ConcurrentHashMap;
 import org.eclipse.persistence.sessions.Session;
 import org.eclipse.persistence.sessions.SessionProfiler;
 import org.eclipse.persistence.tools.profiler.PerformanceMonitor;
+
+import java.util.*;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Collect metrics from one or more EclipseLink Session instances.
@@ -54,7 +51,7 @@ public class EclipseLinkStatisticsCollector extends Collector {
      * creating an empty collector and adding the session using {@link #add(Session, String)}.
      *
      * @param session The EclipseLink Session to collect metrics for
-     * @param name A unique name for this Session
+     * @param name    A unique name for this Session
      */
     public EclipseLinkStatisticsCollector(Session session, String name) {
         add(session, name);
@@ -64,7 +61,7 @@ public class EclipseLinkStatisticsCollector extends Collector {
      * Registers an EclipseLink Session with this collector.
      *
      * @param session The EclipseLink Session to collect metrics for
-     * @param name A unique name for this Session
+     * @param name    A unique name for this Session
      * @return Returns the collector
      */
     public EclipseLinkStatisticsCollector add(Session session, String name) {
@@ -96,9 +93,9 @@ public class EclipseLinkStatisticsCollector extends Collector {
                         metrics -> {
                             Object value = metrics.getOrDefault(SessionProfiler.UowCommits, 0);
                             if (value instanceof Long) {
-                                return ((Long)value).doubleValue();
+                                return ((Long) value).doubleValue();
                             }
-                            return ((Integer)value).doubleValue();
+                            return ((Integer) value).doubleValue();
                         }
                 ),
                 createCounter(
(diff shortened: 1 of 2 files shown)


simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
```diff
@@ -0,0 +1,744 @@
...
+      throw new IllegalStateException("You must register at least one Session.");
+    }
+    return super.register(registry);
+  }
+
+  private List<MetricFamilySamples> getCounters() {
+    return Arrays.<MetricFamilySamples>asList(
+        createCounter(
+            "eclipselink_uow_commits_total",
+            "Total number of unit of work commits",
```
brian-brazil: Is uow a common term?

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
```diff
@@ -0,0 +1,744 @@
...
+                if (value != null) {
+                  return ((Integer) value).doubleValue();
+                }
+                return 0;
+              }
+            }
+        ),
+        createCounter(
+            "eclipselink_cache_misses_total",
+            "Total number of cache misses",
```
brian-brazil: It's best to have cache requests as a metric too, it makes it easier to calculate the hit ratio.

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
```diff
@@ -0,0 +1,744 @@
...
+                if (value != null) {
+                  return ((Integer) value).doubleValue();
+                }
+                return 0;
+              }
+            }
+        ),
+        createCounter(
+            "eclipselink_change_sets_processed_total",
+            "Total number of processed change sets",
```
brian-brazil: Are this one and the next one successes and failures? Failures and total is a bit easier to work with

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
```diff
@@ -0,0 +1,744 @@
...
+                Object value = metrics.get(SessionProfiler.OptimisticLockException);
+                if (value != null) {
+                  return ((Integer) value).doubleValue();
+                }
+                return 0;
+              }
+            }
+        ),
+        createCounter(
+            "eclipselink_rcm_received_total",
```
brian-brazil: What's rcm?

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
```diff
@@ -0,0 +1,744 @@
...
+              }
+            }
+        )
+    );
+  }
+
+  private List<MetricFamilySamples> getTimers() {
+    return Arrays.<MetricFamilySamples>asList(
+        createGauge(
+            "eclipselink_uow_commits_seconds",
```
brian-brazil: These all sound like counters

VitaNuova: They all measure durations of particular operations in seconds, not sure how they can be counters. As durations, they can arbitrarily go up and down.

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
VitaNuova: Afaik, Unit of Work is a common term (see also https://stackoverflow.com/questions/3532258/unit-of-work-pattern-and-history). However, I should probably avoid the abbreviation in the metric name.

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
brian-brazil: In that case see if you can get to the underlying data, and create a Summary from that

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
VitaNuova: Agree, but the metric as such was not present in EclipseLink statistics. I can add cache requests as the sum of cache hits and cache misses.

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
VitaNuova: RemoteCommandManager (https://www.eclipse.org/eclipselink/api/2.2/org/eclipse/persistence/sessions/coordination/RemoteCommandManager.html). I can also spell out the whole thing in the metric name.

VitaNuova:
Thank you for the comments :)

simpleclient_eclipselink/src/main/java/io/prometheus/client/eclipselink/EclipseLinkStatisticsCollector.java
VitaNuova: I see your point, but EclipseLink API returns one value at a time which represents the cumulative duration of operation. Unfortunately, I don't see how I can get enough data for a summary or a histrogram in one scrape.

PR #391, opened 2018-06-21
Extend JVM metrics with thread states

Teots:
@brian-brazil 

Hi,

at work we found it quite useful to get an insight in the number of existing threads and their state. I created this PR, in case other want to use this feature as well. I could also change this PR and implement it as a separate `Collector`. 

Looking forward to your feedback.

I attached an exemplary grafana dashboard showing the threads grouped by state: 

screenshot 2018-06-21 10 22 22

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
@@ -84,27 +84,28 @@ public class ThreadExports extends Collector {
       Collections.singletonList("state"));
 
     Map<Thread.State, Integer> threadStateCounts = getThreadStateCountMap();
-    for (Thread.State state : Thread.State.values()) {
-      String stateName = state.toString().toLowerCase();
+    for (Map.Entry<Thread.State, Integer> entry : threadStateCounts.entrySet()) {
       threadStateFamily.addMetric(
-        Collections.singletonList(stateName),
-        (threadStateCounts.containsKey(stateName)) ? threadStateCounts.get(stateName) : 0.0d
+        Collections.singletonList(entry.getKey().toString()),
+        entry.getValue()
       );
     }
   }
 
   private Map<Thread.State, Integer> getThreadStateCountMap() {
     ThreadInfo[] allThreads = threadBean.getThreadInfo(threadBean.getAllThreadIds(), StackTraceDepth);
-    HashMap<Thread.State, Integer> threadCounts = new HashMap<Thread.State, Integer>();
 
+    // Initialize the map with all thread states
+    HashMap<Thread.State, Integer> threadCounts = new HashMap<Thread.State, Integer>();
+    for (Thread.State state : Thread.State.values()) {
+      threadCounts.put(state, 0);
+    }
+
+    // Collect the actual thread counts
     for (ThreadInfo curThread : allThreads) {
       if (curThread != null) {
         Thread.State threadState = curThread.getThreadState();
-        if (threadCounts.containsKey(threadState)) {
-          threadCounts.put(threadState, threadCounts.get(threadState) + 1);
-        } else {
-          threadCounts.put(threadState, 1);
-        }
+        threadCounts.put(threadState, threadCounts.get(threadState) + 1);
       }
     }
 
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
@@ -7,7 +7,11 @@ import io.prometheus.client.GaugeMetricFamily;
 import java.lang.management.ManagementFactory;
 import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Exports metrics about JVM thread areas.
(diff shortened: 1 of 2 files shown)


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -5,9 +5,9 @@
 import io.prometheus.client.GaugeMetricFamily;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
brian-brazil: Imports should be left explicit

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -73,8 +73,38 @@ void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
         "jvm_threads_deadlocked_monitor",
         "Cycles of JVM-threads that are in deadlock waiting to acquire object monitors",
         nullSafeArrayLength(threadBean.findMonitorDeadlockedThreads())));
+
+    Map<Thread.State, Integer> threadStateCounts = getThreadStateCountMap();
+    for (Map.Entry<Thread.State, Integer> entry : threadStateCounts.entrySet()) {
+      String stateName = entry.getKey().toString().toLowerCase();
+      sampleFamilies.add(
+        new GaugeMetricFamily(
+          String.format("jvm_threads_%s", stateName),
```
brian-brazil: This should be a label, not part of a metric name.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -73,8 +73,38 @@ void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
...
+          String.format("Current count of threads in state '%s'", stateName),
+          entry.getValue().doubleValue()));
+    }
   }
 
+  private Map<Thread.State, Integer> getThreadStateCountMap() {
+    ThreadInfo[] allThreads = threadBean.getThreadInfo(threadBean.getAllThreadIds(), StackTraceDepth);
+    HashMap<Thread.State, Integer> threadCounts = new HashMap<Thread.State, Integer>();
```
brian-brazil: These timeseries will not always be present, you should create all known states in advance

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
Teots: Sure, I'll change that.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
Teots: Ok

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -73,8 +77,42 @@ void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
...
         nullSafeArrayLength(threadBean.findMonitorDeadlockedThreads())));
+
+    GaugeMetricFamily threadStateFamily = new GaugeMetricFamily(
+      "jvm_threads_state",
+      "Current count of threads by state",
+      Collections.singletonList("state"));
+
+    Map<Thread.State, Integer> threadStateCounts = getThreadStateCountMap();
+    for (Thread.State state : Thread.State.values()) {
+      String stateName = state.toString().toLowerCase();
```
brian-brazil: Why lowercase?

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
```diff
@@ -73,8 +77,42 @@ void addThreadMetrics(List<MetricFamilySamples> sampleFamilies) {
...
         "Cycles of JVM-threads that are in deadlock waiting to acquire object monitors",
         nullSafeArrayLength(threadBean.findMonitorDeadlockedThreads())));
+
+    GaugeMetricFamily threadStateFamily = new GaugeMetricFamily(
+      "jvm_threads_state",
+      "Current count of threads by state",
+      Collections.singletonList("state"));
+
+    Map<Thread.State, Integer> threadStateCounts = getThreadStateCountMap();
+    for (Thread.State state : Thread.State.values()) {
```
brian-brazil: It'd be a bit simpler to do this down in the function before fetching the data.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
Teots: Doesn't make sense anymore. Before those names where used in the metrics name, but since that changed, I can just use the normal name - without lower case.

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/ThreadExports.java
Teots: Ok I'll move the code.

brian-brazil:
Okay that looks good, you just need to add a DCO to your commits.

Teots:
I added a DCO to all my commits.

brian-brazil:
Thanks!

klingtnet:
@brian-brazil 

> What are the performance implications of gathering this data on a thread-heavy JVM?

I couldn't find an answer to this question in this pull request, but I'm curious to know if there is a significant performance impact.
At least, I would expect because this is basically a thread-dump each time the metrics are served.
Did you measure the impact?

astral303:
The performance of ThreadsExports is already abysmal due to `jvm_threads_deadlocked_monitor` and `jvm_threads_deadlocked`. That's basically at least two thread dumps, so what's a third one? This one at least runs less logic over that list.

PR #392, opened 2018-06-22
Add Gradle assets to README

amdelamar:
@beorn7 - Just adding to the README. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,24 @@ version can be found on in the maven repository for
 </dependency>
 ```
 
+Or if you use Gradle:
+
+```gradle
+dependencies {
+    // The client
+    compile 'io.prometheus:simpleclient:0.4.0'
+    
+    // Hotspot JVM metrics
+    compile 'io.prometheus:simpleclient_hotspot:0.4.0'
+    
+    // Exposition HTTPServer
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. The existing examples has to be updated by hand on every release, and I'd prefer not to have additional updates to do.

amdelamar:
Ok let's improve it then.

We can substitute variables for the version numbers. Like so:

```gradle
dependencies {
    compile 'io.prometheus:simpleclient:' + prometheusVersion
}
```

We can do the same for Maven:

```xml
<dependencies>
  ...
  <dependency>
    <version>${prometheus.version}</version>
  </dependency>
  ...
</dependencies>
```

brian-brazil:
That would stop the current list being copy&pasteable.

amdelamar:
True. But its a common way of listing assets.

Example: [fluent/fluent-logger-java](https://github.com/fluent/fluent-logger-java)

brian-brazil:
That's not what I'm aiming for here, I like things to be as copy&pastable as possible to reduce friction.

amdelamar:
Ok. No problem. We'll close it then for now

PR #393, opened 2018-07-02
Adding the _sum metric to histogram

m1k3:
As some prometheus parsers will crash if they don’t see it.

e.g. https://github.com/Stackdriver/stackdriver-prometheus/issues/5

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -79,7 +79,8 @@ public class DropwizardExports extends io.prometheus.client.Collector implements
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.98"), snapshot.get98thPercentile() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.99"), snapshot.get99thPercentile() * factor),
                 new MetricFamilySamples.Sample(name, Arrays.asList("quantile"), Arrays.asList("0.999"), snapshot.get999thPercentile() * factor),
-                new MetricFamilySamples.Sample(name + "_count", new ArrayList<String>(), new ArrayList<String>(), count)
+                new MetricFamilySamples.Sample(name + "_count", new ArrayList<String>(), new ArrayList<String>(), count),
+                new MetricFamilySamples.Sample(name + "_sum", new ArrayList<String>(), new ArrayList<String>(), Arrays.stream(snapshot.getValues()).sum())
         );
         return Arrays.asList(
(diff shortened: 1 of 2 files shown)


brian-brazil:
The _sum is not required for a histogram, and parsers should be able to deal with that. When I last looked this value is only the sum of the current samples, not the true sum that we need so it should remain missing.

m1k3:
They `should`, but the one I linked does not 😉 .  The `CollectorRegistry` will also add it in the `collectorNames` method https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L110.

brian-brazil:
I can only accept this change if it makes sense in terms of the metric itself, which I don't believe it does.

I'm not willing to expose incorrect data to workaround faulty parsers.

m1k3:
> I'm not willing to expose incorrect data to workaround faulty parsers.

In a perfect world I would agree, but does it hurt to add the `_sum`? Especially since I saw it added to the collector names [here](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L110).

brian-brazil:
There's no meaningful _sum we can expose, so not exposing it is the right thing to do.

That code is only saying that the histogram can have a sum, not that it must.

m1k3:
> When I last looked this value is only the sum of the current samples, not the true sum that we need so it should remain missing.
> There's no meaningful _sum we can expose

I was trying to look up some documentation about it, but I only found this found https://prometheus.io/docs/concepts/metric_types/#histogram and it says `the total sum of all observed values, exposed as <basename>_sum` for both histogram and summary. 

I wouldn't ask to expose a metric that we don't have data for, but I was under the impression that the sum is the sum of all observed values. That should be what we get if we do `Arrays.stream(snapshot.getValues()).sum()`, or am I missing something?

Could you please explain in more detail why it should remain missing?

brian-brazil:
This is a question of Dropwizard semantics, not Prometheus semantics. As far as I remember, that expression does not produce the sum we need.

m1k3:
Ah, you are correct. Dropwizard snapshot does not have all the values. I did find [this](https://github.com/dropwizard/metrics/pull/1254) though. It should be exactly what is needed to expose the sum. Would you consider using 5.0.0 as the dependency once they release the final, or do you want to maintain backwards compatibility?

brian-brazil:
I want to maintain backwards compatibility, as rather old versions tend to be in use. If there's a sane way to do that and support this new feature I'd be interested.

m1k3:
I don't know of a sane way of doing it without resorting to reflections. Since you are not willing to bump the dependency to 5.0.0 once it is released, I am closing this PR.

brian-brazil:
We have resorted to reflection elsewhere. There's nothing stopping someone maintaining a version that depends on 5.0.0.

PR #394, opened 2018-07-02
Fixing filtering with CollertorRegistry

m1k3:
It would cause an unsupported expection on an interator [here](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L201) when there is more than one metric in a collector since `Arrays.asList` will produce an immutable list.

see https://stackoverflow.com/questions/2965747/why-do-i-get-an-unsupportedoperationexception-when-trying-to-remove-an-element-f

Also, the `describe` method was always returning an empty list causing the filtering [here](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L98) to not find anything.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -32,7 +32,8 @@ public class DropwizardExports extends io.prometheus.client.Collector implements
         String name = sanitizeMetricName(dropwizardName);
         MetricFamilySamples.Sample sample = new MetricFamilySamples.Sample(name, new ArrayList<String>(), new ArrayList<String>(),
                 new Long(counter.getCount()).doubleValue());
-        return Arrays.asList(new MetricFamilySamples(name, Type.GAUGE, getHelpMessage(dropwizardName, counter), Arrays.asList(sample)));
+        return Arrays.asList(new MetricFamilySamples(name, Type.GAUGE, getHelpMessage(dropwizardName, counter),
+                new ArrayList<MetricFamilySamples.Sample>(Arrays.asList(sample))));
     }
 
     private static String getHelpMessage(String metricName, Metric metric){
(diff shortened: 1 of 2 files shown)


simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
```diff
@@ -32,7 +32,8 @@ public DropwizardExports(MetricRegistry registry) {
         String name = sanitizeMetricName(dropwizardName);
         MetricFamilySamples.Sample sample = new MetricFamilySamples.Sample(name, new ArrayList<String>(), new ArrayList<String>(),
                 new Long(counter.getCount()).doubleValue());
-        return Arrays.asList(new MetricFamilySamples(name, Type.GAUGE, getHelpMessage(dropwizardName, counter), Arrays.asList(sample)));
+        return Arrays.asList(new MetricFamilySamples(name, Type.GAUGE, getHelpMessage(dropwizardName, counter),
+                new ArrayList<MetricFamilySamples.Sample>(Arrays.asList(sample))));
```
brian-brazil: Is this the most efficient way to produce a list?

simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
```diff
@@ -2,16 +2,14 @@
...
 import io.prometheus.client.CollectorRegistry;
 import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
brian-brazil: imports should be explicit

m1k3:
I also had a version where the filter method was not trying to modify the underlying iterator, but as was working with other collectors I thought it made sense to modify the `DropwizardExport` to work with the code in the current `CollectorRegistry`.

simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
m1k3: Not sure about most efficient, but it is suggested as the solution in the thread I linked.

m1k3:
Should I close this PR in favor of fixing it in the `filter`?

brian-brazil:
Yes please.

m1k3:
Closing in favor of #395

PR #395, opened 2018-07-02
Fixing filtering with CollectorRegistry

m1k3:
Using `iterator.remove()` can cause a problem with Lists that are produced using `Arrays.asList`. e.g. https://github.com/prometheus/client_java/blob/master/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java#L75-L83

see https://stackoverflow.com/questions/2965747/why-do-i-get-an-unsupportedoperationexception-when-trying-to-remove-an-element-f

Also, the `describe` method of `DropwizardExports` was always returning an empty list causing the filtering to not find anything.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -157,6 +157,6 @@ public class DropwizardExports extends io.prometheus.client.Collector implements
 
     @Override
     public List<MetricFamilySamples> describe() {
-      return new ArrayList<MetricFamilySamples>();
+      return collect();
     }
 }
--- a/simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
+++ b/simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
@@ -9,9 +9,12 @@ import org.junit.Test;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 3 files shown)


m1k3:
I would have used something more elegant like

```java
next.samples.stream()
                .filter(sample -> includedNames.contains(sample.name))
                .collect(Collectors.<Collector.MetricFamilySamples.Sample>toList());
```

but we are [targetting 1.6](https://github.com/prometheus/client_java/blob/master/pom.xml#L71-L72) 😢

simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
```diff
@@ -157,6 +157,6 @@ public static String sanitizeMetricName(String dropwizardName) {
 
     @Override
     public List<MetricFamilySamples> describe() {
-      return new ArrayList<MetricFamilySamples>();
+      return collect();
```
brian-brazil: Why are you changing this?

simpleclient_dropwizard/src/test/java/io/prometheus/client/dropwizard/DropwizardExportsTest.java
```diff
@@ -9,15 +9,15 @@
...
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
brian-brazil: Please keep imports explicit

simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
m1k3: I mentioned it in #394 and here in the description above. 

If describe always returns an empty list, it causes the filtering [here](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L98) to not find anything.

brian-brazil: I believe I disabled this as dropwizard semantics and standard usage allow for the set of metrics to be not particularly static so this feature wouldn't work well for them.

m1k3: The whole filtering feature does not work for `DropwizardExports` then. If I revert it to always return an empty list, the test I added will not work.

brian-brazil: You could write a small custom collector for the test.

m1k3: What would that accomplish? I am trying to fix the filtering with the actual `DropwizardExports`.

brian-brazil: It verifies the bugfix to the CollectorRegistry.

Filtering can't be made to reliably work with DropwizardExports, which is why it is explicitly disabled.

m1k3: If filtering can't be made to work reliably with `DropwizardExports `, then the behavior should be more meaningful than just not returning anything. When I tried the [example repo](https://github.com/RobustPerception/java_examples/blob/master/java_dropwizard/src/main/java/io/robustperception/java_examples/JavaDropwizard.java) that is using the java_client with Dropwizard and I use the filtering that the [servlet](https://github.com/prometheus/client_java/blob/master/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java#L51) obviously provides I get an empty result instead of a something telling me that filtering is not supported for Dropwizard.

m1k3:
While there may be good reasons for filtering to not work reliably for DropwizardExports, the communications of those reasons is anything but. Closing.

simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
brian-brazil: It's a limitation of how it works, we need a list of all possible metric names in advance which custom collectors cannot always provide resulting in false positives - such as you saw. It always works for direct instrumentation and usually for simpler custom collectors.

m1k3: I can understand that it is a limitation, but what I find baffling is that it does not work when using `DropwizardExports` with no clue as to why. As it is, the example simply does not work as expected with no indication why.

A simple comment explaining the situation above the `describe` method could clear it up...

Issue #364, opened 2018-04-05
Pushgateway client does not support Basic Authentication


tom-mi:
When running the pushgateway behind a reverse proxy with basic auth enabled, as [recommended in the FAQ](https://prometheus.io/docs/introduction/faq/#why-don%27t-the-prometheus-server-components-support-tls-or-authentication?-can-i-add-those?), it is not possible to use the Java Client, as it does not support Basic Authentication.

Please fix.

cluttazi:
+1

bitsal:
Please look at my PR. I am open to polish it as needed in order to merge the PR.

PR #396, opened 2018-07-04
Added ability to add authentication logic into pushgateway request

bitsal:
Added Basic Auth. implementation.

@brian-brazil 
See https://github.com/prometheus/client_java/issues/364.


brian-brazil:
This only supports basic auth. I'd prefer to make this pluggable so any auth method can be used.

bitsal:
@brian-brazil 
1. Now you don't have any auth
2. I've implemented only one auth mechanism. Anyone else will be able to implement own even outside of the library (I've left a door for this)

If you have any code suggestions/critics just write it down I'm opened to implement it of course if it is not so much.

>I'd prefer to make this pluggable so any auth method can be used
Probably, I don't understand what exactly you mean. Add more details about your idea(s).

brian-brazil:
How would this work if someone wanted say TLS client cert auth, or to adjust TLS settings?

I'd like to avoid the situation where I have to maintain every possible http configuration.

bitsal:
Easy. It is necessary to create a class like `TlsPushGateway` and override `authenticate()` based on this e.g. https://stackoverflow.com/questions/30121510/java-httpsurlconnection-and-tls-1-2
```
protected void authenticate(HttpURLConnection connection) {
     SSLContext sc = SSLContext.getInstance("TLSv1");
     // Init the SSLContext with a TrustManager[] and SecureRandom()
     sc.init(null, trustCerts, new java.security.SecureRandom());
     
     ((HttpsURLConnection) connection).setSSLSocketFactory(sc.getSocketFactory());
}
```

amlozano:
I think that a better solution would be to extract the code from the Http client outside PushGateway code, and inject it into the constructor. In that case you can compose PushGateway with the HttpClient that you want, allowing any kind of extension open for the user.

bitsal:
@amlozano Agree. A composition is much better. I've tried not to add a lot of code because I don't know what limitations are of using other libraries for http(s) requests.

@brian-brazil Could you tell me about such limitations? Like Java version, dependencies and so on.

amlozano:
@bitsal What I meant is to extract the line of 
`HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();`
The class PushGateway depends on the abstract class HttpURLConnection, so the creation of the exact implementation is outside the code. In that way you can create an HttpsURLConnection configuring the 
`connection.setRequestProperty("Authorization", basicAuthHeader);`
or the 
`connection.setSSLSocketFactory(sc.getSocketFactory());`
outside PushGateway.

brian-brazil:
The only real consideration is that we need to stick with Java 6 features.

bitsal:
@amlozano It was clear, thanks. I will add composition and try to be back-compatible just not to break existing code.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/BasicAuthHttpConnectionFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    public HttpURLConnection create(String url) throws IOException {
+        HttpURLConnection connection = originConnectionFactory.create(url);
+        connection.setRequestProperty("Authorization", basicAuthHeader);
+        return connection;
+    }
+
+    private String encode(String user, String password) {
+        try {
+            byte[] message = (user + ":" + password).getBytes("UTF-8");
+            String encoded = javax.xml.bind.DatatypeConverter.printBase64Binary(message);
```
brian-brazil: This should be imported.

bitsal: Fixed

brian-brazil:
Can you add the DCO and update the docs please?

bitsal:
Updated docs and added sign-off commit.

bitsal:
@brian-brazil DCO is still red. I wonder how it works.
I've installed it using 
```
gem install dco
dco enable
```

README.md
```diff
@@ -568,12 +568,34 @@ void executeBatchJob() throws Exception {
 }
  ```
 
-A separate registry is used, as the default registry may contain other metrics
-such as those from the Process Collector. See the 
-[Pushgateway documentation](https://github.com/prometheus/pushgateway/blob/master/README.md)
-for more information.
+ A separate registry is used, as the default registry may contain other metrics
```
brian-brazil: No leading spaces on these lines

amlozano:
@bitsal I still see it better to be in a constructor rather than in the setConnectionFactory. Otherwise you will create something that later on has to be replaced. I.E. 
`myGateway = new PushGateway(url);
myGateway.setConnectionFactory(factory);`

the second line is replacing the URL configuration, which is mandatory for no reason.
I see it more natural to define a new constructor with the HttpConnectionFactory, so you can use either the address, the URL or the ConnectionFactory to create the connection.

BTW, I am not part of the project, but we use it in our team, and it would be nice as an user to have that in place.

README.md
bitsal: Removed

bitsal:
@amlozano `BasicAuthHttpConnectionFactory` should add only basic auth based on single responsibility principle if you want to configure HTTPS you should add own `HttpsConnectionFactory` that will decorate `BasicAuthHttpConnectionFactory` and add HTTPS logic.
```java
new BasicAuthHttpConnectionFactory(new HttpsConnectionFactory(), 'username', 'verySecretPassword');
```

bitsal:
@brian-brazil Probably I should remove `BasicAuthPushGateway` beacuse it is more as an example than real helpful gateway. What do you think?

amlozano:
@bitsal I agree on your comment, I just said that the BasicAuthPushGateway is enforcing you to use an HTTP connection, since it uses the constructor from PushGateway that sets an http url.

I think that class doesn't need to exist, since it is already easy to configure the BasicAuth just with composition.

brian-brazil:
Yeah we can simplify, as long as it's covered in the docs.

bitsal:
Done

brian-brazil:
That all looks good, can you ensure all commits have the DCO?

bitsal:
@brian-brazil not all of course, first ones are not
Is there a way to add to these first commits DCO?

brian-brazil:
The easiest way is to squash your commits.

bitsal:
I have never done it before. @brian-brazil do you have an exact command(s) how to do it without recreation a branch?

brian-brazil:
`git rebase -i master` and set all but one commit to squash.

bitsal:
@brian-brazil create a new PR https://github.com/prometheus/client_java/pull/398

PR #406, opened 2018-07-30
corrected name of method in README

rbarazzutti:
Hi @brian-brazil,
Small fix in the documentation. Method name mismatch with the actual library.

Thanks.


brian-brazil:
Thanks!

PR #407, opened 2018-08-03
Fix overlap between invoke() and value() in Spring MVC

patrick-webs:
I found that sometimes when I hit the `/prometheus` endpoint on my Spring Boot application, I got a blank page back. Other times it returned the expected set of complete results. This behavior seemed to change between restarts.

After investigation, I think the problem is that `EndpointMvcAdapter.invoke()` has a default `@RequestMapping` that matches the `@RequestMapping` of the value(name) method defined on `PrometheusMvcEndpoint`.

https://github.com/spring-projects/spring-boot/blob/5863e6f78c78f840127b2ae64ea9c422a06df9f3/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/mvc/EndpointMvcAdapter.java#L41-L46

Defining a distinct `/filtered` request mapping on the value method allows filtered results to consistently work by hitting the URL `/prometheus/filtered?name[]=foo`, while not having the issue of unfiltered results sometimes not working when hitting `/prometheus`.

For whatever reason, with my version of Spring Boot (1.3.7) and Spring Framework (4.2.7), the defaultValue of "" for value does not turn into an empty Set. I do see other people on the Internet claiming this works, so I'm guessing it is a behavior that was added in a later version of Spring. This change thus also improves compatibility with older spring versions.

Edit: made adjustments to the code once I realized there were some unit tests that were failing.

### DIFF
--- a/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMvcEndpoint.java
+++ b/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot/PrometheusMvcEndpoint.java
@@ -1,6 +1,9 @@
 package io.prometheus.client.spring.boot;
 
+import static org.springframework.http.HttpHeaders.CONTENT_TYPE;
+
 import io.prometheus.client.exporter.common.TextFormat;
+
 import org.springframework.boot.actuate.endpoint.mvc.EndpointMvcAdapter;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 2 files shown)


brian-brazil:
I don't know enough Spring to understand this change, however how the endpoints behave should not change.

patrick-webs:
closing this based on the above comment. I think the solution for anyone who runs into this is to update to a newer version of spring, spring boot 1.3.x has exited support a while ago and from what I can tell, this library worked fine with spring boot 1.5.x at the time I made this PR.

Issue #381, opened 2018-05-21
Improve MetricsFilter to report on HTTP status code


umatrangolo:
Hi,

I'm using the MetricsFilter to plot RPS and P99/P95 of my services but I have just found out that there is no trace of successful/failed requests. 

This metric would be useful to fire alerts whenever the count of 500s/404s or any code != 200 gets over a given threshold.

Given that sounds so obvious and it is surprising that is not there already I was wondering if this has been not implemented by design.

If not I could give it a shot.

danieldabate:
Totally agree with @umatrangolo. We have the same use case and think that this is really useful for everyone.

+1

rbarazzutti:
I also agree with @umatrangolo  and @danieldabate.

The documentation of Prometheus shows even some [nice examples](https://prometheus.io/docs/prometheus/latest/querying/examples/) on what do with that value.

+1

rbarazzutti:
Finally, I wrote some code and did [a pull request](https://github.com/prometheus/client_java/pull/408) for this feature.

kdombeck:
It seems like the [PR](https://github.com/prometheus/client_java/pull/408) is at an impasse.

Currently today the metrics that are created are like the following
```
# HELP http_request The time taken fulfilling servlet requests
# TYPE http_request histogram
http_request_bucket{path="/my/api/1",method="GET",le="0.1",} 3.0
http_request_bucket{path="/my/api/1",method="GET",le="0.25",} 3.0
http_request_bucket{path="/my/api/1",method="GET",le="0.5",} 3.0
http_request_bucket{path="/my/api/1",method="GET",le="0.75",} 3.0
http_request_bucket{path="/my/api/1",method="GET",le="1.0",} 3.0
http_request_bucket{path="/my/api/1",method="GET",le="+Inf",} 4.0
http_request_count{path="/my/api/1",method="GET",} 4.0
http_request_sum{path="/my/api/1",method="GET",} 1.5861285180000002
http_request_bucket{path="/my/api/2",method="GET",le="0.1",} 2.0
http_request_bucket{path="/my/api/2",method="GET",le="0.25",} 2.0
http_request_bucket{path="/my/api/2",method="GET",le="0.5",} 2.0
http_request_bucket{path="/my/api/2",method="GET",le="0.75",} 2.0
http_request_bucket{path="/my/api/2",method="GET",le="1.0",} 2.0
http_request_bucket{path="/my/api/2",method="GET",le="+Inf",} 2.0
[...]

PR #408, opened 2018-08-05
MetricsFilter to report on HTTP status code

rbarazzutti:
Hi @brian-brazil,

As other users mentioned in [issue 381](https://github.com/prometheus/client_java/issues/381), it would be quite convenient to have the servlet's MetricsFilter reporting on HTTP status code.

I implemented it and wrote a dedicated unit test.

Thanks.

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
@@ -1,6 +1,7 @@
 package io.prometheus.client.filter;
 
 import io.prometheus.client.Histogram;
+import io.prometheus.client.SimpleTimer;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -9,6 +10,7 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
 /**
@@ -65,7 +67,8 @@ public class MetricsFilter implements Filter {
     private String help = "The time taken fulfilling servlet requests";
     private double[] buckets = null;
 
-    public MetricsFilter() {}
+    public MetricsFilter() {
+    }
 
     public MetricsFilter(
             String metricName,
@@ -91,7 +94,7 @@ public class MetricsFilter implements Filter {
             return str;
         }
         int count = 0;
-        int i =  -1;
+        int i = -1;
         do {
             i = str.indexOf("/", i + 1);
             if (i < 0) {
@@ -107,7 +110,7 @@ public class MetricsFilter implements Filter {
     @Override
     public void init(FilterConfig filterConfig) throws ServletException {
         Histogram.Builder builder = Histogram.build()
-                .labelNames("path", "method");
+                .labelNames("path", "method", "status_code");
 
         if (filterConfig == null && isEmpty(metricName)) {
             throw new ServletException("No configuration object provided, and no metricName passed via constructor");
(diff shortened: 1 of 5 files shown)


brian-brazil:
Thanks for your PR. It is recommended not to break out latency by success/failure, as that often leads to users only graphing success and missing important data. This also is a breaking change, and may have cardinality issues. If you want to add this it should be as a separate counter metric.

rbarazzutti:
Hi @brian-brazil,

Thanks for your answer, the issues you mentioned do make sense. It is right that this change might be somehow breaking.

IMHO, it'd be nice to have a metric that contains both latencies and status codes. To illustrate this, I'd say that 500s responses might be served very quickly or pretty slowly (i.e. database connectivity loss).

Would it ok to be able to enable this with a parameter on the MetricsFilter? 

Thanks.

brian-brazil:
A given metric shouldn't really vary it's labels across processes. If you
need something custom you can always do direct instrumentation of your app
if you have custom needs, this code is more to help you get something basic
running out of the box.

rbarazzutti:
Hi @brian-brazil,

Thanks for sharing your point of view.

I understand that a metric that changes its labels values across processes might be misleading. But in this specific case, the labels aren't changing; one of them, the status, is attributed once the HTTP response is ready. This behavior seems to be pretty common and is even presented in the querying example of the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/examples/). And I'm pretty sure that this feature might interest other programmers according to [issue 381](https://github.com/prometheus/client_java/issues/381).

Following your remark that this change might be breaking,  I redefined this feature as an option of the MetricsFilter (that is disabled by default).  So it won't be breaking for anyone.

Thanks.
Cheers.

brian-brazil:
> I redefined this feature as an option of the MetricsFilter (that is disabled by default).

Metrics still shouldn't vary their labels across processes. This should be a separate counter metric.

MattiasSeeman:
So now that it is a separate metric, what's the view on this PR?

@brian-brazil, you mention that this
> is a breaking change

Did you have something specific in mind (e.g. "A given metric shouldn't really vary it's labels across processes"), or something else?

For people instrumenting legacy applications, this PR introduces a dependency on Servlet 3.0 that _would be_ breaking (for us).

brian-brazil:
> Did you have something specific in mind

Adding instrumentation labels is a breaking change, and this still appears not to be a separate metric.

rbarazzutti:
> Adding instrumentation labels is a breaking change, and this still appears not to be a separate metric.

Hi @brian-brazil, the filter defined in the class `StatusAwareMetricFilter` provides a separate metric. The user has to choose between the two approaches/metrics, between `MetricFilter` and `StatusAwareMetricFilter`.

`StatusAwareMetricFilter` also builds a histogram, because, in my opinion, it is useful when somebody wants to record SLIs about the latency of a web service. I.e. to be able to select `GET`s with status `200` and not take into account `GET`s with status `3xx` (redirects that are likely not part of the service to be evaluated).
Like in [https://engineering.bitnami.com/articles/implementing-slos-using-prometheus.html](https://engineering.bitnami.com/articles/implementing-slos-using-prometheus.html)

Last point: as mentioned @MattiasSeeman the retrieval of status (`HttpServletResponse::getStatus()`) was not in early specifications of `servlet-api` (came with version 2.5), so it may be problematic when instrumenting legacy applications,  this change is not _breaking_ for these users as long as they stick to the usual `MetricFilter`.

brian-brazil:
> The user has to choose between the two approaches/metrics, between MetricFilter and StatusAwareMetricFilter.

The user shouldn't have to choose. This is intended as a simple way to get some quick metrics, not a full-on solution.

>  I.e. to be able to select GETs with status 200 and not take into account GETs with status 3xx

As I mentioned elsewhere this is a bad idea, as while that is a rare use case which exists and what actually generally happens is users fail to include any failures in their latency graphs.

rbarazzutti:
Okay. Got your point. Thanks, I'll see how I can adapt this.

rbarazzutti:
Looks like https://github.com/prometheus/client_java/pull/560 fits the requirements.

PR #409, opened 2018-08-15
Swap out usages of synchronized for a ReentrantReadWriteLock

AgentK20:
This PR swaps out usages of synchronized in CollectorRegistry for a ReentrantReadWriteLock, improving performance in situations where new collectors are being added or removed concurrently while the collector is being run.

(Pinging @brian-brazil as per CONTRIBUTING.md)

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -11,6 +11,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


AgentK20:
Force pushed an amendment to add sign-off commit

brian-brazil:
May I ask why you are expecting to be changing collectors while things are being collected?

AgentK20:
Apologies for the delay on this. I did a bit of further digging and in almost all scenarios it seems that the existing synchronized code outperforms a reentrant lock, so this PR isn't needed.

As for my use case, I've got an event-driven async system with various plugins that are loaded and unloaded throughout the lifetime of the process and wanted to make sure that this wouldn't be a bottleneck.

PR #411, opened 2018-08-17
set Java source and target to 1.8

rbarazzutti:
hi @brian-brazil,

Maven was complaining because `maven-compiler-plugin` version wasn't specified.

This can be done without calling an extra plugin, just setting the `maven.compiler.source` and `maven.compiler.source` variables.

Cheers,

Raphaël

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,9 +67,6 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-
-        <maven.compiler.source>1.6</maven.compiler.source>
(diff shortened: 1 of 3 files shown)


brian-brazil:
There's some build tool out there that doesn't support variables, so we avoid using them. Can you fix this without them?

rbarazzutti:
Hi,

Good to know. I just fixed it.

To be more consistent, I also changed `jetty-jdk8`'s POM.

Thanks

brian-brazil:
Thanks!

Issue #412, opened 2018-08-21
simpleclient_httpserver not respecting HTTP 1.1 spec


LucasBoisserie:
Hello,

Source issue: prometheus/jmx_exporter#304

According to the http 1.1 spec:

>  A sender MUST NOT send a Content-Length header field in any message
   that contains a Transfer-Encoding header field. (https://tools.ietf.org/html/rfc7230#section-3.3.2).

In [HTTPServer implemetation](https://github.com/prometheus/client_java/blob/master/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java#L68), the header `Content-Length` is always set but the chunked mode is enable in gzip part when in `sendResponseHeaders` method the `responseLength` is set to 0.
```
t.getResponseHeaders().set("Content-Length", String.valueOf(response.size()));
if (HTTPServer.shouldUseCompression(t)) {
    t.getResponseHeaders().set("Content-Encoding", "gzip");
    t.sendResponseHeaders(200, 0L);
```

PR #413, opened 2018-08-21
Fix HTTP header content-length when using compression

LucasBoisserie:
This PR fix issues: prometheus/client_java#412 and prometheus/jmx_exporter#304


brian-brazil:
Thanks!

PR #414, opened 2018-08-22
Add port public getter in HTTPServer

kovalevm:
If I don't want to configurate port for metrics I can pass '0' and it means use random avalible port, it's very useful but how I can know which port was used?


brian-brazil:
Can you explain your setup where this comes up? How is Prometheus discovering this port?

brian-brazil:
That makes sense. Thanks!

PR #417, opened 2018-08-27
Javadoc fix in simpleclient_servlet

rbarazzutti:
hi @brian-brazil,

a small PR with fixes in the JavaDoc (displays properly the XML code)

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
@@ -14,6 +14,26 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
+/**
+ * The MetricsServlet class exists to provide a simple way of exposing the metrics values.
+ *
+ * <p>This servlet might be used with a filter such the {@link io.prometheus.client.filter.MetricsFilter}
+ *
+ * <p>Usage example (in webapp's {@code web.xml}):
+ * <pre>{@code
+ * <!-- instance of the servlet -->
+ * <servlet>
+ *   <servlet-name>metrics</servlet-name>
+ *   <servlet-class>io.prometheus.client.exporter.MetricsServlet</servlet-class>
+ * </servlet>
+ *
+ * <!-- map it to an URL -->
+ * <servlet-mapping>
+ *   <servlet-name>metrics</servlet-name>
+ *   <url-pattern>/metrics/*</url-pattern>
+ * </servlet-mapping>
+ * }</pre>
+ */
 public class MetricsServlet extends HttpServlet {
 
   private CollectorRegistry registry;
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
@@ -17,7 +17,6 @@ import java.util.Set;
 /**
  * The MetricsServlet class exists to provide a simple way of exposing the metrics values.
  *
- * <p>This servlet might be used with a filter such the {@link io.prometheus.client.filter.MetricsFilter}
  */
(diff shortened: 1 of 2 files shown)


simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
```diff
@@ -14,6 +14,23 @@
 import java.util.HashSet;
 import java.util.Set;
 
+/**
+ * The servlet exposes the metrics
+ *
+ * <p>Example config to export metrics on path {@code /metrics}:
```
brian-brazil: This makes no sense to me as it lacks context. What is it that takes this configuration?

As this is not tied to the servlet itself, I don't think this should be here.

rbarazzutti: thanks for the feedback. tried to address that

simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
```diff
@@ -14,6 +14,26 @@
 import java.util.HashSet;
 import java.util.Set;
 
+/**
+ * The MetricsServlet class exists to provide a simple way of exposing the metrics values.
+ *
+ * <p>This servlet might be used with a filter such the {@link io.prometheus.client.filter.MetricsFilter}
```
brian-brazil: That's for something else.

simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
```diff
@@ -14,6 +14,26 @@
...
+ *
+ * <p>Usage example (in webapp's {@code web.xml}):
```
brian-brazil: What's a web.xml?

This seems specific to one particular framework, and only one way that this could be used. The changes to this file don't belong here.

rbarazzutti: The web.xml file is the deployment descriptor for a Servlet-based Java web application. It's pretty common.
The XML snippet that is in MetricsFilter.java is also referring to that web.xml.

A user of the servlet may want to know how to use it, this way is the most straightforward.

brian-brazil: I personally have never heard of it, nor have I ever needed to use it to use servlets. This sounds like something specific to the framework you are using, and not generally applicable and thus doesn't belong here.

brian-brazil: I presumed it was something Spring related, as it came in in the same PR. Claiming that this applies to all web applications seems a bit broad.

rbarazzutti: This PR is "Spring-free" ...

I don't claim it applies to all web application. But for sure, its the most common way of deploying a very simple servlet (especially when not using an extra framework). It's a simple example, that helps many.

But I'm still trying to figure out why this syntax is fine in `MetricsFilter.java` and not with the servlet. I added it for consistency.

rbarazzutti: Hi @brian-brazil,

I'll see how to rephrase this.

But it is supposed to work on any web app container that supports the servlet 3.0 specs (which is referred by the pom.xml of this module). 

Cheers

simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
brian-brazil: This line shouldn't be here, exposition and instrumentation are separate concerns.

brian-brazil:
Thanks!

PR #419, opened 2018-09-03
Fix thread state metrics not being exposed

napstr:
@brian-brazil

#391 introduced thread state metrics, however while they are collected, they are not actually exposed. Adding this line should fix that.


napstr:
I have also expanded the test case to cover the thread state metric. I'm new to Mockito, hope my test code fits this project's expectations.

brian-brazil:
Thanks!

PR #423, opened 2018-09-11
Put response body in pushgateway exception (#83)

ibalashov:
Following suggestions from SO:

[How to get response body using HttpURLConnection, when code other than 2xx is returned?](https://stackoverflow.com/questions/25011927/how-to-get-response-body-using-httpurlconnection-when-code-other-than-2xx-is-re)


brian-brazil:
Thanks!

PR #427, opened 2018-10-04
Add support for using non-default CollectorRegistry with the logback …

davidh87:
…InstrumentedAppender

We have need of specifying a non-default CollectorRegistry to use with the logback InstrumentedAppender. Only added to logback, can port to log4j / log4j2 too if preferred.

Tests updated to support both defaultRegistry and a newly created one.

First time contributor - I believe I need to ping @brian-brazil on pull requests!?

### DIFF
--- a/simpleclient_logback/pom.xml
+++ b/simpleclient_logback/pom.xml
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.11</version>
+            <version>4.12</version>
             <scope>test</scope>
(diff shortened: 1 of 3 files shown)


brian-brazil:
Can you expand on why you want this?

brian-brazil:
This has gone stale, so closing as they never got back to me.

agavrilov76:
@brian-brazil We would like to have this feature as well. 
Using non-default CollectorRegistry in the application can be useful, for example, for unit testing the metrics generated by different classes. So it would be nice to have logging and application metrics in one registry

PR #430, opened 2018-10-19
Allow configuring MetricsFilter to use other instance of CollectorReg…

llech:
Added collectorRegistry field to MetricsFilter.
It can be set from outside, making MetricsFilter using other collectorRegistry (useful when the lifecycle of http server differs from JRE).
Deregister Collector in destroy() method: allows stopping and starting container without triggering collector name conflict (because it would cause Collector to be re-registered).

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
@@ -1,5 +1,6 @@
 package io.prometheus.client.filter;
 
+import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.Histogram;
 
 import javax.servlet.Filter;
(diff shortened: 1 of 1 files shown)


simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -175,5 +184,6 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 
     @Override
     public void destroy() {
+      collectorRegistry.unregister(histogram);
```
brian-brazil: This shouldn't be removed, the values should continue to be exposed until the process terminates.

evfool: I second that this should happen for metricsFilter: in our use-case a web application which uses MetricsFilter can not be hot-redeployed to the running tomcat due to the metric being already registered. With this fix it would work: on the filter destroy (happening on redeploy) the metrics would be reset. This in some cases is not ideal indeed, but the current way it breaks the whole idea of "hot-deployable" war files to a running container.

tkgregory:
Can we please get this merged in? I'm also having to maintain a separate copy of `MetricsFilter`, as it's not really designed for extension ;)

brian-brazil:
This has gone stale and doesn't make sense in Prometheus terms, if you need something a bit more custom than MetricsFilter offers then I'd suggest direct instrumentation.

PR #440, opened 2018-11-20
Improvements on Meter export, allowing to get the requests per second.

reinaldomjr:
total -> total requests received
mean -> a mean of requests/second since the service started
1min -> requestes/second during last minute
5min -> requestes/second during last 5 minutes
15min -> requestes/second during last 15 minutes

This is useful to generate metrics to detect increase in usage and scale services accordingly


brian-brazil:
You can already calculate these from the _total, so there's no need to have the extras using up resources.

reinaldomjr:
You're right, these values can be calculated, but that will require changes on the consumer of the metrics, that's not always possible.
In my case, the consumer just show the value, and I can put triggers on these metrics, but I can't customize it to calculate the weighted moving average.

PR #442, opened 2018-11-27
Formatter fixed and unified for Quantile and Item classes

Draczech:
@brian-brazil 
Minor fix of formatter in [CKMSQuantiles](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java) *Item* and *Quantile* classes. Now both use the same format.


Draczech:
I had to align my private and office address in signoff, now everything should be OK.

brian-brazil:
Thanks!

PR #446, opened 2018-12-17
Add mean, min, max and stddev dropwizard metrics to prometheus samples

usulkies:


brian-brazil:
See https://prometheus.io/docs/instrumenting/writing_exporters/#drop-less-useful-statistics

brian-brazil:
Average is covered by https://www.robustperception.io/common-query-patterns-in-promql. Min and max are not possible, which is why these are excluded.

usulkies:
I don't think this article is relevant for the exported Dropwizard metrics.
When I use the @Timed annotation on a method (and my code have pretty much of those) - then Dropwizard creates a timer metrics for those methods. I want to monitor the duration time of these functions: maybe minimum and maximum are not possible unless I'm using a custom sliding time window, but there should be a relatively simple way to calculate the average execution duration for a given time range. If I have the count **and sum**, I guess can calculate the average using the increase in the required time range. But if all I have is the count and the quantiles data - then it is much more complicated to solve these equations, if even possible.
I feel like some essential data from the timer is missing here, and it should be added. WDYT?

brian-brazil:
> I don't think this article is relevant for the exported Dropwizard metrics.

It was created specifically with Dropwizard in mind.

> I feel like some essential data from the timer is missing here, and it should be added. WDYT?

Yes, and that's something to take up with Dropwizard. I think the very latest version might have finally added a sum.

PR #447, opened 2018-12-17
Dropwizard: Pass on labels when no custom mapping matched

twz123:
In case of no matching `MapperConfig`, quantiles were broken, since the `CustomMappingSampleBuilder` swallowed up the labels.


brian-brazil:
Thanks!

PR #449, opened 2018-12-18
Added exclusive/shared recording modes

franz1981:
Single-writer recording mode added to improve performance
on event-loop based applications.

### DIFF
--- a/benchmark/src/main/java/io/prometheus/benchmark/CounterBenchmark.java
+++ b/benchmark/src/main/java/io/prometheus/benchmark/CounterBenchmark.java
@@ -3,13 +3,16 @@ package io.prometheus.benchmark;
 import com.codahale.metrics.MetricRegistry;
 
 import java.util.concurrent.TimeUnit;
+
+import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
 import org.openjdk.jmh.annotations.Mode;
-import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.BenchmarkParams;
 import org.openjdk.jmh.runner.Runner;
 import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
@@ -17,6 +20,9 @@ import org.openjdk.jmh.runner.options.OptionsBuilder;
 
 @State(Scope.Benchmark)
 public class CounterBenchmark {
+
+  @Param({"true", "false"})
+  public boolean shared;
 
   MetricRegistry registry;
   com.codahale.metrics.Counter codahaleCounter;
@@ -29,7 +35,10 @@ public class CounterBenchmark {
   io.prometheus.client.Counter prometheusSimpleCounterNoLabels;
 
   @Setup
-  public void setup() {
+  public void setup(BenchmarkParams params) {
+    if (params.getThreads() > 1 && !shared) {
+      throw new IllegalStateException("Exclusive counters are supported only if threads = 1");
+    }
     prometheusCounter = io.prometheus.client.metrics.Counter.newBuilder()
       .name("name")
       .documentation("some description..")
@@ -39,12 +48,13 @@ public class CounterBenchmark {
     prometheusSimpleCounter = io.prometheus.client.Counter.build()
       .name("name")
       .help("some description..")
-      .labelNames("some", "group").create();
+      .labelNames("some", "group").shared(shared).create();
     prometheusSimpleCounterChild = prometheusSimpleCounter.labels("test", "group");
 
     prometheusSimpleCounterNoLabels = io.prometheus.client.Counter.build()
       .name("name")
       .help("some description..")
+      .shared(shared)
       .create();
 
     registry = new MetricRegistry();
@@ -107,7 +117,7 @@ public class CounterBenchmark {
       .include(CounterBenchmark.class.getSimpleName())
       .warmupIterations(5)
       .measurementIterations(4)
-      .threads(4)
+      .threads(1)
       .forks(1)
       .build();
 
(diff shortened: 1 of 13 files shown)


franz1981:
I will add some test (probably adding a parametrized JUnit test to the existing ones) if it will seems ok :+1:

brian-brazil:
The existing code should keep to one bucket if there's only one thread in use, so I'd expect no differences.

franz1981:
The difference is of near 4X for the single writer case and the reasons are thanks to "lazySet" and a better locality (is a double[]).

brian-brazil:
You're not doing any locking, which is going to make a difference. This code must be threadsafe.

franz1981:
"You're not doing any locking": the point of being single-writer is that there is just one thread calling observe  so the only property required is to perform atomic writes and lazySet is exactly providing it.
What do you mean by locks here?

brian-brazil:
You can't presume that, there could be threads in some of the libraries you use for example.

franz1981:
Help me to understand, please.
If I'm the user of this client library and I know that my observations are happening with single-writer semantic why I can't have the benefit of lightweight observations?
It is now more clear what use cases I'm considering?

brian-brazil:
What if someone uses your library in a threaded fashion?

brian-brazil:
Such a thing might be at quite a distance from the user, and not plumbed through. There's nothing stopping you having local collectors for custom purposes, but I don't think this is safe to have in the standard library.

franz1981:
I do understand your point, but in that case there is already the default behaviour (that's safe) ie shared = true no?

franz1981:
> And libraries are the main intended users.

I want to integrate this lib into the Artemis broker (https://github.com/apache/activemq-artemis), but we have on the hot paths several cases where counters are just single-writer: having this feature will make much more lighweight (4X is not negligible) to use it.
If you feel that would be better to have a separate factory for each counter allocation/creating, instead that a global builder option, I can change it too :+1:

brian-brazil:
What i'm saying is that it'd be unsafe for this sort of thing to live in this library, as it'd end up misused. I'd also suggest you do the math on what the impact would be. 20ns for 1M events in a second would be 2% of your potential capacity.

franz1981:
> What i'm saying is that it'd be unsafe for this sort of thing to live in this library, as it'd end up misused

That's why the default is with `shared = true` + adding proper docs too to avoid miusing: I agree that it won't save the day if a user won't use it properly, but it can be true with any options.

> I'd also suggest you do the math on what the impact would be. 20ns for 1M events in a second would be 2% of your potential capacity.

I'm more concerned about latency spikes TBH and if we (on hot path) handle brust of activities with 10 M msg/sec, having latency spikes due to added observations will have impacts, limiting both the quality of the observations and the response time of the path monitored.
Consider that the benchmarks (the `SummaryBenchmark` to be specific) of this library are not showing the case where many buckets are written and I'm quite sure that ~20 ns would became more in that case.

brian-brazil:
> I'm more concerned about latency spikes TBH and if we (on hot path) handle brust of activities with 10 M msg/sec, having latency spikes due to added observations will have impacts, limiting both the quality of the observations and the response time of the path monitored.

If you have a highly constrained use case, there's nothing stopping writing a custom collector.

> Consider that the benchmarks (the SummaryBenchmark to be specific) of this library are not showing the case where many buckets are written and I'm quite sure that ~20 ns would became more in that case.

That's not how the DoubleAdder works, a single thread will always write to a single bucket. In addition, it's only on read that you'd need to care about how many buckets there were.

franz1981:
> If you have a highly constrained use case, there's nothing stopping writing a custom collector.

Not following you: why if we can just improve what we have? Having event-loop based application is not an uncommon pattern

> That's not how the DoubleAdder works, a single thread will always write to a single bucket. In addition, it's only on read that you'd need to care about how many buckets there were.

it is not what I meant, sorry: I mean this one:
```
        @Override
        public void observe(double amt) {
            for (int i = 0; i < upperBounds.length; ++i) {
                // The last bucket is +Inf, so we always increment.
                if (amt <= upperBounds[i]) {
                    cumulativeCounts[i].add(1);
                    break;
                }
            }
            sum.add(amt);
        }
```
`amt` need to have a specific value to trigger several DoubleAdder instances to get updated.
`DoubleAdder` with a single threaded usage will always perform a compareAndSwap operation that need to be sequential consistent leading to an added overhead if compared with a plain lazySet operation.

brian-brazil:
> Not following you: why if we can just improve what we have? Having event-loop based application is not an uncommon pattern

Because it's not safe in all use cases, and an event loop design isn't exactly common in Java.

> amt need to have a specific value to trigger several DoubleAdder instances to get updated.

It'll only ever update one bucket.

> compared with a plain lazySet operation.

Looking at the spec, the code you propose is not safe as there's no bound on when the set actually happens. It could be after the next read of the value.

franz1981:
> Because it's not safe in all use cases, and an event loop design isn't exactly common in Java.

vert-x based applications are not so uncommon, to quote one type, but is just an opinion.

> It'll only ever update one bucket.

I have already corrected my previous comment about it: I have understood your point here :)

> Looking at the spec, the code you propose is not safe as there's no bound on when the set actually happens. It could be after the next read of the value.

Not sure I'm following you on this: it gives the same behaviour of DoubleAdder if used with single threaded usage, please take a look here: http://psy-lob-saw.blogspot.com/2012/12/atomiclazyset-is-performance-win-for.html

franz1981:
> and it indicates that there's no guarantee that the write will happen before subsequent reads.

Where you have read it? The lazySet doc is not correct as the bug link shows.

franz1981:
There are lot of core high-performance libraries that make use of this pattern: https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/atomic/SpscAtomicArrayQueue.java#L183

franz1981:
Probably this article is much better to understand it: http://psy-lob-saw.blogspot.com/2016/12/what-is-lazyset-putordered.html

franz1981:
@brian-brazil Any other info needed for this re `lazySet` behaviour? 
Probably @njhill could take a look too?

njhill:
@franz1981 LGTM! Had you considered a middle-ground of using ThreadLocal + your `lazySet`? May retain most of the perf improvement while staying threadsafe (hopefully addressing @brian-brazil's concerns). There is the caveat of possibly increased mem overhead in certain scenarios, but those with high thread churn likely aren't ones who care so much about perf characteristics anyhow.

Here's _another_ Nitsan reference (I guess you may have already seen... oldish but I think mostly still applicable) that seems very relevant to this .. and he describes the TL approach at the end, even with example code: http://psy-lob-saw.blogspot.com/2013/06/java-concurrent-counters-by-numbers.html

brian-brazil:
This has gone stale, and I don't think makes sense in a generic client library.

PR #450, opened 2018-12-19
Hotspot: DefaultExports with a custom CollectorRegistry

twz123:
Add a `register(CollectorRegistry)` method, so that DefaultExports can also be used with CollectorRegistries besides the default one.


bjoernhaeuser:
With spring boot 2 there is the possibility to have a autowired / spring managed collector registry, which is not the default one. The default exports then register against another collector registry, which is not getting exported.

If the details are cleared I can provide a PR for this. :)

brian-brazil:
Thanks!

PR #451, opened 2018-12-19
Add metrics for Retrofit 2 calls

kdombeck:
This creates an interceptor for [Retrofit](https://square.github.io/retrofit/). It was modeled after the [MetricsFilter](https://github.com/prometheus/client_java/blob/master/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java)

It was made possible after the implementation of these issues
https://github.com/square/okhttp/issues/270
https://github.com/square/okhttp/issues/4087
https://github.com/square/retrofit/issues/2732

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,22 @@ new QueuedThreadPoolStatisticsCollector()
     .register();
 ```
 
+### Retrofit
+
+There is an interceptor for OkHttp which allows for the measuring the duration taken for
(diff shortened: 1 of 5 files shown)


brian-brazil:
Thanks for your PR. I'm not accepting new modules, so I'd suggest having this live in your own repo.

kdombeck:
I understand and I really appreciate that this project allows me to easily see how to configure different projects to use Prometheus. Thanks for getting back to me so soon.

PR #456, opened 2019-01-13
Set thread names in HTTPServer

snago:
Name threads in the HTTPServer thread pool to make them easily identifiable in e.g. thread dumps.

Before: `pool-1-thread-3`
After: `prometheus-http-1-3`

@brian-brazil


brian-brazil:
Thanks!

PR #462, opened 2019-03-08
Update caffeine version to latest

davidwrpayne:
# What
The Caffeine Library has been updated.
https://github.com/ben-manes/caffeine/releases

# Why
It would be good to use latest version of caffeine for next release

First time using pom in a long time. Can't really remember if I'm missing anything important. If I am, just let me know. Happy to fix.

@brian-brazil

### DIFF
--- a/simpleclient_caffeine/pom.xml
+++ b/simpleclient_caffeine/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
-            <version>2.3.0</version>
+            <version>2.7.0</version>
         </dependency>
         
 


brian-brazil:
We prefer to stick with older versions, so we don't accidentally depend on a newer feature.

ben-manes:
In this case, you might want to upgrade to use `AsyncCache` instead, which was split off from `AsyncLoadingCache` . I try to be good about semver so there shouldn't be any compatibility problems, only bug fixes and various improvements.

https://github.com/prometheus/client_java/blob/923d23e33135044b59d1d736ebdfd29154ecca02/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java#L75-L77

brian-brazil:
We don't actually use this code, we're just providing metrics from it. Accordingly we use the lowest version possible to ensure that we're forwards compatible.

davidwrpayne:
Alright. Closing the PR. In the past have been bitten by dependency hell and this was my first instinct. Thanks for the quick response!

PR #465, opened 2019-03-24
Introduce ExtremumCollector, collector that can keep track of actual …

Falland:
This collector can keep track of actual max/min values through time. Other collectors can only give extrema for averaged time-series value which sometimes is hiding actual worst-case scenarios values.
There is no way currently in getting this information from Prometheus. This class is trying to solve this problem.
No adjustments on Prometheus server side is required, as this collector looks like simple Gauge, the only difference is stores only extremum value for the observed period.
Inside it keeps several sample buckets for configured sampling period (e.g. 30s and 10s for sample bucket), within the bucket only extremum value is being stored. The overall extremum is simply calculated over all the buckets.
This approach allows to provide sliding min/max for a metric with little memory and cpu overhead.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorClock.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorClock.java
@@ -1,0 +1,26 @@
+package io.prometheus.client;
+
+public interface CollectorClock {
+
+    CollectorClock DEFAULT_CLOCK = new DefaultClock();
+
+    long millis();
+
+    long nanos();
+
+    class DefaultClock implements CollectorClock {
+
+        DefaultClock() {}
+
+        @Override
+        public long millis() {
+            return System.currentTimeMillis();
+        }
+
+        @Override
+        public long nanos() {
+            return System.nanoTime();
+        }
+    }
+
+}
--- a/simpleclient/src/main/java/io/prometheus/client/Extremum.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Extremum.java
@@ -10,9 +10,10 @@ import java.util.concurrent.atomic.AtomicReferenceArray;
 import static io.prometheus.client.CollectorClock.DEFAULT_CLOCK;
 
 /**
- * Extremum metric, to report maximum/minimum values observed within within configurable sliding window.
+ * Extremum metric, to report non-averaged  maximum/minimum values observed within configurable sliding window.
+ * This metric can be helpful in addition to Histogram or Summary metrics to show the worst/best case.
  * <p>
- * Examples of ExtremumCollector include:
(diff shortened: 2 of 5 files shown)


brian-brazil:
Thank you for your PR. The PromQL execution model is that you can calculate over arbitrary periods, not just at 10s/30s boundaries so this doesn't make sense overall.

Falland:
PromQL does not allow to get the exact min/max values, only averaged in time. This PR is not aiming to calculate anything, the only aim is to keep track of worst-case scenarios. And then with PromQL one can calculate over it.

In our project we've tried to track the maximum latency of the handling a request . We've tried histograms but they are not showing the exact values only number of requests that were falling in the specific bucket. This class allowed us to get this information.

brian-brazil:
> PromQL does not allow to get the exact min/max values, only averaged in time. 

Yes, over arbitrary time periods. Any proposals must work over arbitrary time periods, which isn't possible for this sort of data with metrics.

Falland:
That is not true. Everything in Prometheus is quantized by scraping period, so PromQL can not give you time discretion smaller than that period. This collector should be configured to collect values with higher rate that scraping period (at least two times more often). So having that, it will allow PromQL to run queries over arbitrary time span, higher than scraping period.

brian-brazil:
A client does not however know the scrape interval, nor should it.

Falland:
Yes, I agree. Maybe I've put it wrong in the description of the PR, the sampling period should long enough and the amount of samples should be big enough to be able to track min/max values correctly. Maybe I should remove any scraping period metioning from the documentation. I know this collector is a little bit trickier to configure, then other ones. But I believe this metric can come very handy in many applications

Falland:
I've removed the scraping period from javadoc.

Falland:
BTW, look like all the build are failing due to javadoc check. I'm not sure what is causing it, but could be new version of java on the build agent or maven plugin version update (which I haven't seen)

Falland:
I have updated the javadoc to make the case a little bit more clear. I've use Summary's sliding window description, although the options names are different they are basically represent the same concepts.
I've also renamed the class to Extremum.
I can add performance tests to it as well.
But let's first agree on do you think this class should be added to collectors or not.

From my perspective the class can be really handy for the cases when you want to see the exact worst case in addition to Summary or Histogram.
I understand that 100th (q=1.0) quantile in Summary can give the same effect, but using summary just for this is not efficient and resource consuming.
Extremum is just a lightweight way of getting 100th quantile for the metric.

brian-brazil:
I don't think this belongs in the core client, as it is not in line with the PromQL model nor the future OpenMetrics thinking. You can always maintain it as a 3rd party library.

Issue #470, opened 2019-04-05
Add SSL support to HTTPServer


thecoop:
Allow `HTTPServer` to be created to use SSL.

This can be as simple as adding a constructor overload allowing an instance of `com.sun.net.httpserver.HttpServer` to be passed in, which could then be set as a configured instance of `HttpsServer`. Currently it's hardcoded to plain `HttpServer`.

brian-brazil:
Would you like to send a PR?

PR #471, opened 2019-04-08
Allow an HttpServer to be passed in

thecoop:
@brian-brazil PR for #470 - allowing a HttpServer to be passed in, so HttpsServer could be used instead


brian-brazil:
Can you add the DCO?

thecoop:
Done!

brian-brazil:
Thanks!

PR #473, opened 2019-04-10
Update logback to 1.2.0 to fix CVE-2017-5929

### LINKED ISSUE #472: Vulnerability in simpleclient_logback
There seems to be a vulnerability in `simpleclient_logback` version 0.6.0.

    maven: io.prometheus:simpleclient_logback:0.6.0  Confidence:Highest
    cpe: cpe:/a:logback:logback:0.6.0  Confidence:Low

    CVE-2017-5929

    Severity:High
    CVSS Score: 7.5 (AV:N/AC:L/Au:N/C:P/I:P/A:P)
    CWE: CWE-502 Deserialization of Untrusted Data

QOS.ch Logback before 1.2.0 has a serialization vulnerability affecting the SocketServer and ServerSocketReceiver component.

Please update logback in `simpleclient_logback` to version 1.2.0.

sebastianhaeni:
Fixes #472

### DIFF
--- a/simpleclient_logback/pom.xml
+++ b/simpleclient_logback/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.1.2</version>
+            <version>1.2.0</version>
         </dependency>
         
 


brian-brazil:
We purposefully use old versions to ensure we don't accidentally depend on newer features. The issue you mention does not affect this code, and you'd be using whatever version you have in your own deps.

sebastianhaeni:
Fact is, that your library produces a vulnerability warning for everybody using it and checking it with OWASP dependency check.

The logback authors clearly mention that everybody should be upgrading to at least 1.2.0 because of the vulnerability.

BTW the `Encoder` interface has changed. Other than that there are only bug fixes.
* You're not using `Encoder`. So you're good in theory.
* But if you would be using `Encoder`, you would use an old interface and thus it would not be compatible with projects that use 1.2.0 or newer.

See https://logback.qos.ch/news.html

brian-brazil:
That sounds like a false positive then, as it's the effective version that should matter rather than the unused version in a dep.

sebastianhaeni:
Ok.

If anybody encounters the same false positive, this is how you can suppress it:

    <suppress>
        <gav regex="true">io\.prometheus:simpleclient_logback:.*</gav>
        <cve>CVE-2017-5929</cve>
    </suppress>

Issue #466, opened 2019-03-28
CollectorRegistry.filteredMetricFamilySamples should filter on metrics name, not samples name


peergynt:
If I have an Histogram named `my_histogram`, I need to pass the 3 strings:
-  `my_histogram_count`
-  `my_histogram_sum`
-  `my_histogram_bucket`

to `CollectorRegistry.filteredMetricFamilySamples()` to see all the samples of my histogram.

I was thinking that it would make more sense to filter on the metric name, and not have the client trying to figure out all the possible sample names.

brian-brazil:
This is how it's implemented in both clients that have it currently. One reason is so that you can pick out just the sum/count of a histogram, without having to transfer all the buckets.

peergynt:
Opened PR #474 to clarify behavior.

PR #474, opened 2019-04-20
Added comment to filteredMetricFamilySamples to clarify behavior

peergynt:
Added comment to clarify behavior which did not seem obvious to user (see #466)

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -130,6 +130,13 @@ public class CollectorRegistry {
     return new MetricFamilySamplesEnumeration();
   }
 
+  /**
+   * Enumeration of metrics matching the specified names.
+   * <p>
+   * Note that the provided set of names will be matched against the time series
+   * name and not the metric name. For instance, to retrieve all samples from a
+   * histogram, you must include the '_count', '_sum' and '_bucket' names.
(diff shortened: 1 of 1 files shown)


simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -130,6 +130,13 @@ public void clear() {
     return new MetricFamilySamplesEnumeration();
   }
 
+  /**
+   * Enumeration of metrics matching the specified names.
+   * <p>
+   * Note that the set of names will be matched against the metric sample names
```
brian-brazil: Time series name and metric name is the appropriate naming.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -130,6 +130,13 @@ public void clear() {
...
+   * and not the parent metric name. For instance, to retrieve all samples from
+   * an histogram, you must include the '_count', '_sum' and '_bucket' names.
```
brian-brazil: a histogram

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
peergynt: Tweaked comment based on suggestion.

PR #476, opened 2019-04-24
Do not throw IllegalArgumentException on null label or less number labels than predefined

solosky:
Hi everyone, 

Recently, I got some error on null label exception in production, and it caused some impact on our customers. 

```
Caused by: java.lang.IllegalArgumentException: Label cannot be null.
	at io.prometheus.client.SimpleCollector.labels(SimpleCollector.java:68)
```

First I admit the error should never happened in production, it should be well tested before rolling out to production. Then I begun to think that how to prevent this error from begining.

So I wrote this pull request to silent any instrumentation exception on null label or unequals labels. 
I have docuemented well for this change, for best undertanding, I picked piece of changed code here. 

I think this change may not be agreed from some developers, but it's fine, let's disucss here.  I was open for that. 

Thank you. 

```
  /**
   * Return the Child with the given labels, creating it if needed.
   * <p>
   * It should be passed the same number of labels are were passed to {@link #labelNames}.
   * If null or less than number of labels defined, a default label {@link #DEFAULT_LABEL_VALUE} value was applied
   * to the label value.
   */
  public Child labels(String... labelValues) {
    /*
      For best practise, we should void any fatal exception on instrumentation
      The basic principal was that the instrumentation code in application should never impact the business logic
      We can accept monitoring data error but no application fault on instrumentation error
      The silent strategy may case some metric data error, but it's a trade off between service error and monitoring data error
    */
    List<String> key = new ArrayList<String>(labelNames.size());
    for (int i = 0; i< labelNames.size(); i++) {
        if (labelValues != null
                && labelValues.length > i
                && labelValues[i] != null) {
          key.add(labelValues[i]);
        } else {
          key.add(DEFAULT_LABEL_VALUE);
        }
    }
```

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
@@ -25,7 +25,7 @@ import java.util.List;
  * The fullname of the metric is <code>namespace_subsystem_name</code>, but only <code>name</code> is required.
  *
  * <h2>Labels</h2>
- * {@link SimpleCollector.Builder#labelNames labelNames} specifies which (if any) labels the metrics will have, and 
+ * {@link SimpleCollector.Builder#labelNames labelNames} specifies which (if any) labels the metrics will have, and
  * {@link #labels} returns the Child of the metric that represents that particular set of labels.
  * {@link Gauge}, {@link Counter} and {@link Summary} all offer convenience methods to avoid needing to call
  * {@link #labels} for metrics with no labels.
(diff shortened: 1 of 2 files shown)


brian-brazil:
Passing an incorrect number of labels is a programming error, and should result in an exception.

solosky:
> Passing an incorrect number of labels is a programming error, and should result in an exception.

Hi, I got your point, but how about the null label error? Do you think it's a programning error too ?

brian-brazil:
Yes, null is not a String and we need a string.

solosky:
> Yes, null is not a String and we need a string.

Understand. 
I agree with you that they are all programming error.

But first, let's begin the discuss with the ultimate question, 

> should the business logic be interrupted by an instrumentation exception?

My short answer was No. 
We can accept metric data error, but no business logic interruption.

please kindly share your thought. Thanks.

brian-brazil:
Coding bugs are coding bugs, and can cause problems in many places. It's also unsafe to blindly replace bad values with unknown, which could clash with an existing label.

I'm going to close this off for now, let me know if you've other ideas.

Issue #480, opened 2019-05-22
Thread contention in TimeWindowQuantiles because of `synchronized`


t3hnar:
Hello guys,
Use of Summary gives considerable performance penalty because of multiple `synchronized` calls in `.observe` flow.
Will you able to fix this?

brian-brazil:
This is a fundamental issue with client-side quantiles. If you remove the quantiles or use a Histogram instead, you can avoid this specific problem.

t3hnar:
> This is a fundamental issue with client-side quantiles

do you mean that this is not solvable without `synchronized ` problem in computer science ?

brian-brazil:
I've yet to see a solution that doesn't require this.

rrakos-evo:
I'll investigate if we could use `atomics` / `compare-and-swap` instead of `locks` / `synchronized`, as that might yield better performance for our use cases.

rrakos-evo:
I might have found something strange, but I could be misunderstanding the details. 🤔

It looks like [TimeWindowQuantiles#get](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/TimeWindowQuantiles.java#L30) reads `current bucket`, but [TimeWindowQuantiles#insert](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/TimeWindowQuantiles.java#L35) writes `all buckets`.
Note that [CKMSQuantiles#insert](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java#L90) is `synchronized`, so the writes are causing unnecessary thread contention.
I'm wondering what's the point of writing buckets which are never read?

brian-brazil:
They're rotated through. Having a quantile that represents recent data (rather than since process start) is tricky.

rrakos-evo:
I’ll check out `HdrHistogram` as the next experiment, as it looks like it’s designed exactly for our use case, but I’m not sure how well it integrates with Prometheus.

rrakos-evo:
We'll continue with profiling the alternatives and experiments (#481, #482, #483, #484) under realistic workload(s), as I can't produce similar thread contention with synthetic benchmarks.

rrakos-evo:
I'm still working on testing and profiling #482 + #484 using different use cases.
I won't be able to test them in isolation as I was asked to test them together.
[...]

PR #482, opened 2019-06-05
Use atomics instead of synchronized for Gauge

rrakos-evo:
Idea:
> We could implement `Gauge#set` with initializing a `new DoubleAdder` to avoid exposing invalid state, but that might have significant allocation (GC) overhead.

Pros:
* avoids thread contention for highly concurrent workloads (`get`, `set`)
* reduces constant overhead (`get`, `set`)

Cons:
* introduces allocations (`set` and `inc / dec`)
* increases GC overhead for high throughput workloads (`set` and `inc / dec`)

Relates to #480

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
+++ b/simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
@@ -118,6 +118,37 @@ public class DoubleAdder extends Striped64 implements Serializable {
         internalReset(0L);
     }
 
+    public void set(double x) {
+        for (;;) {
+            Cell[] as;
+            if ((as = cells) != null) { // have cells
+                if (busy == 0 && casBusy()) {
+                    try {
+                        if (cells == as) { // recheck under lock
+                            // allocate new cells instead of updating existing cells
+                            int n = as.length;
+                            Cell[] rs = new Cell[n];
+                            for (int i = 0; i < n; ++i) {
+                                Cell a = as[i];
+                                if (a != null) // avoid unused cells
+                                    rs[i] = new Cell(0L);
+                            }
+                            // update cells and base (not atomic)
+                            cells = rs;
+                            base = Double.doubleToLongBits(x);
+                            break;
+                        }
+                    } finally {
+                        busy = 0;
+                    }
+                }
+            } else { // no cells
+                base = Double.doubleToLongBits(x); // update base (atomic)
+                break;
+            }
+        }
+    }
+
     /**
      * Equivalent in effect to {@link #sum} followed by {@link
      * #reset}. This method may apply for example during quiescent
--- a/simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
+++ b/simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
@@ -94,13 +94,17 @@ public class DoubleAdder extends Striped64 implements Serializable {
      * @return the sum
      */
     public double sum() {
+        // On concurrent `sum` and `set`, it is acceptable to `get` an outdated `value`.
+        // On concurrent `sum` and `add`, it is acceptable to `get` an outdated `value`.
+        // On concurrent `sum` and `set` and `add`, it is possible to `get` an outdated `value`.
+
         // Correctness is guaranteed by `volatile` memory access ordering and visibility semantics.
         // Program order:
         //  - writes in `set` - `busy` (CAS), `cells` (Wc), `base` (Wb), `busy`
         //  - reads in `sum` - `cells` (Rc), `base` (Rb), `busy`, `cells` (Cc), `base` (Cb)
         // Note that:
         //  - `busy` is written after `cells` and `base`
-        //  - `busy` is read after `cells` and `base`, `cells` and `base` is re-read after `busy`
+        //  - `busy` is read after `cells` and `base`, then `cells` and `base` is re-read after `busy`
         // In other words:
         //  - if we see the write to `busy`, then we must see the write to `cells` and `busy` on re-read
         //  - if we don't see the write to `busy`, then we must retry as we have no guarantees
@@ -115,7 +119,8 @@ public class DoubleAdder extends Striped64 implements Serializable {
         //    - invalid result - Wc, Rc, Rb, Wb, Cc, Cb - retry
         Cell[] as = cells; long b = base;
         while (as != null && !(busy == 0 && cells == as && base == b)) {
-            Thread.yield(); // busy waiting, retry loop
+            // busy waiting, retry loop
+            Thread.yield();
             as = cells; b = base;
         }
 
@@ -143,22 +148,25 @@ public class DoubleAdder extends Striped64 implements Serializable {
     }
 
     public void set(double x) {
+        // On concurrent `set` and `set`, it should be acceptable to lose one `set` measurement.
+        // On concurrent `set` and `add`, it should be acceptable to lose the `add` measurement.
+
+        // Correctness is ensured by different techniques:
+        //  - `set` waits on contention (blocking)
+        //  - `add` avoids contention (non-blocking)
+        //  - `sum` retries on conflicts (non-blocking)
+        // Performance characteristics by use cases:
+        //  - only `set` - `cells` is always `null` - no allocations
+        //  - only `add` - `cells` allocated on contention
+        //  - mixed `set` and `add` - `cells` allocated on contention, `cells` deallocated on `set`
         for (;;) {
             Cell[] as;
             if ((as = cells) != null) { // have cells
                 if (busy == 0 && casBusy()) {
                     try {
                         if (cells == as) { // recheck under lock
-                            // allocate new cells instead of updating existing cells
-                            int n = as.length;
-                            Cell[] rs = new Cell[n];
-                            for (int i = 0; i < n; ++i) {
-                                Cell a = as[i];
-                                if (a != null) // avoid unused cells
-                                    rs[i] = a.value == 0L ? a : new Cell(0L); // reuse or initialize cell
-                            }
                             // update cells and base (not atomic)
-                            cells = rs;
+                            cells = null;
                             base = Double.doubleToLongBits(x);
                             break;
                         }
--- a/simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
+++ b/simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
@@ -94,8 +94,32 @@ public class DoubleAdder extends Striped64 implements Serializable {
      * @return the sum
      */
     public double sum() {
-        Cell[] as = cells;
-        double sum = Double.longBitsToDouble(base);
+        // Correctness is guaranteed by `volatile` memory access ordering and visibility semantics.
+        // Program order:
+        //  - writes in `set` - `busy` (CAS), `cells` (Wc), `base` (Wb), `busy`
+        //  - reads in `sum` - `cells` (Rc), `base` (Rb), `busy`, `cells` (Cc), `base` (Cb)
+        // Note that:
+        //  - `busy` is written after `cells` and `base`
+        //  - `busy` is read after `cells` and `base`, `cells` and `base` is re-read after `busy`
+        // In other words:
+        //  - if we see the write to `busy`, then we must see the write to `cells` and `busy` on re-read
+        //  - if we don't see the write to `busy`, then we must retry as we have no guarantees
+        // Execution order (in the former case):
+        //  - serial
+        //    - old result - Rc, Rb, Cc, Cb, Wc, Wb
+        //    - new result - Wc, Wb, Rc, Rb, Cc, Cb
+        //  - concurrent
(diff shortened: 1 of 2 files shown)


rrakos-evo:
I ran multiple benchmarks with `24 threads` and picked the result with the smallest error, but of course I can’t know how it behaves under different workloads.

Benchmark:
```
# JMH version: 1.21
# VM version: JDK 11.0.2, OpenJDK 64-Bit Server VM, 11.0.2+9
# VM invoker: /Library/Java/JavaVirtualMachines/openjdk-11.0.2.jdk/Contents/Home/bin/java
# VM options: -javaagent:/Applications/IntelliJ IDEA CE.app/Contents/lib/idea_rt.jar=49590:/Applications/IntelliJ IDEA CE.app/Contents/bin -Dfile.encoding=UTF-8
# Warmup: 5 iterations, 10 s each
# Measurement: 4 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 24 threads, will synchronize iterations
```

Before:
```
Benchmark                                                 Mode  Cnt     Score     Error  Units
GaugeBenchmark.prometheusSimpleGaugeChildDecBenchmark     avgt    4    31.720 ±   1.850  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildIncBenchmark     avgt    4    33.041 ±   0.608  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildSetBenchmark     avgt    4  1441.987 ±  29.700  ns/op
GaugeBenchmark.prometheusSimpleGaugeDecBenchmark          avgt    4   154.518 ±  17.714  ns/op
GaugeBenchmark.prometheusSimpleGaugeIncBenchmark          avgt    4   155.453 ±   7.384  ns/op
GaugeBenchmark.prometheusSimpleGaugeSetBenchmark          avgt    4  3047.817 ± 168.989  ns/op
```

After:
```
Benchmark                                                 Mode  Cnt    Score    Error  Units
GaugeBenchmark.prometheusSimpleGaugeChildDecBenchmark     avgt    4   30.618 ±  0.867  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildIncBenchmark     avgt    4   31.543 ±  5.164  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildSetBenchmark     avgt    4  314.646 ±  4.892  ns/op
GaugeBenchmark.prometheusSimpleGaugeDecBenchmark          avgt    4  150.270 ±  9.382  ns/op
GaugeBenchmark.prometheusSimpleGaugeIncBenchmark          avgt    4  156.412 ±  9.367  ns/op
GaugeBenchmark.prometheusSimpleGaugeSetBenchmark          avgt    4  537.230 ± 67.680  ns/op
```

brian-brazil:
What I was thinking about back when adding this is that we could adjust DoubleAdder to handle this more efficiently if it ever became an issue.

rrakos-evo:
I’m just thinking out loudly:
We could introduce `DoubleAdder#set` as an atomic alternative to `reset` and `add`.
I'm not sure how to update `base` and `cells` atomically without locking using `busy`.

simpleclient/src/main/java/io/prometheus/client/Gauge.java
```diff
@@ -234,9 +231,10 @@ public double setToTime(Runnable timeable){
      * Get the value of the gauge.
      */
     public double get() {
-      synchronized(this) {
-        return value.sum();
-      }
+      // On concurrent `get` and `set`, it is acceptable to `get` an outdated `value`.
+      // On concurrent `get` and `inc / dec`, it is acceptable to `get` an outdated `value`.
+      // On concurrent `get` and `set` and `inc / dec`, it is possible to `get` an invalid `value`.
```
rrakos-evo: I believe this tradeoff should be acceptable, as mixed `set` and `inc / dec` use cases should be rare.

rrakos-evo:
`81283a20`:
```
Benchmark                                                 Mode  Cnt    Score    Error  Units
GaugeBenchmark.codahaleCounterDecBenchmark                avgt    4   35.166 ± 16.052  ns/op
GaugeBenchmark.codahaleCounterIncBenchmark                avgt    4   31.423 ± 10.979  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildDecBenchmark     avgt    4   38.966 ±  3.249  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildIncBenchmark     avgt    4   37.798 ±  9.753  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildSetBenchmark     avgt    4  429.808 ± 92.764  ns/op
GaugeBenchmark.prometheusSimpleGaugeDecBenchmark          avgt    4  182.829 ± 38.544  ns/op
GaugeBenchmark.prometheusSimpleGaugeIncBenchmark          avgt    4  184.254 ± 43.754  ns/op
GaugeBenchmark.prometheusSimpleGaugeSetBenchmark          avgt    4  582.121 ± 86.330  ns/op
GaugeBenchmark.prometheusSimpleGaugeNoLabelsDecBenchmark  avgt    4   37.136 ±  6.599  ns/op
GaugeBenchmark.prometheusSimpleGaugeNoLabelsIncBenchmark  avgt    4   39.167 ±  3.439  ns/op
GaugeBenchmark.prometheusSimpleGaugeNoLabelsSetBenchmark  avgt    4  456.339 ± 61.951  ns/op
```

simpleclient/src/main/java/io/prometheus/client/Gauge.java
brian-brazil: > On concurrent `get` and `set` and `inc / dec`, it is possible to `get` an invalid `value`.

This is not acceptable. You should get some value that some serialisation of the set/inc/get could produce.

simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
```diff
@@ -118,6 +118,37 @@ public void reset() {
...
+                if (busy == 0 && casBusy()) {
+                    try {
+                        if (cells == as) { // recheck under lock
+                            // allocate new cells instead of updating existing cells
+                            int n = as.length;
+                            Cell[] rs = new Cell[n];
+                            for (int i = 0; i < n; ++i) {
+                                Cell a = as[i];
+                                if (a != null) // avoid unused cells
+                                    rs[i] = new Cell(0L);
```
brian-brazil: This could be a lot of allocations.

rrakos-evo: I think we could reuse (steal) the cell if it's `0`:
```
rs[i] = a.value == 0L ? a : new Cell(0L);
```

Let me try to think about the use cases:
* only `set` - `cells` is always `null` - no allocations
* only `inc / dec` - `cells` allocated on contention - few allocations
* mostly `inc / dec` and occasional `set`- `cells` allocated on contention, `cells` reallocated on `set` - some allocations
* mostly `set` and occasional `inc / dec` - `cells` allocated rarely, `cells` reallocated on `set` - some allocations
* mixed `set` and `inc / dec` - `cells` allocated on contention, `cells` reallocated on `set` - a lot of allocations

I'm not sure how to find a balance between the tradeoffs for each use case. 🙁

brian-brazil: Performance wise we only need to worry about the first two cases, however correctness wise we need to worry about everything.

simpleclient/src/main/java/io/prometheus/client/Gauge.java
rrakos-evo: The invalid value is when `sum` reads the new `cells` and old `base`.

Let me try to think about if it's possible to block  `sum` on concurrent `set` without affecting anything else.

rrakos-evo: I think I managed to fix it with busy waiting / retry loop based on the idea of optimistic concurrency / [Software Transactional Memory](https://en.wikipedia.org/wiki/Software_transactional_memory).

simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
```diff
@@ -118,6 +142,38 @@ public void reset() {
...
+                        if (cells == as) { // recheck under lock
+                            // allocate new cells instead of updating existing cells
+                            int n = as.length;
+                            Cell[] rs = new Cell[n];
+                            for (int i = 0; i < n; ++i) {
+                                Cell a = as[i];
+                                if (a != null) // avoid unused cells
+                                    rs[i] = a.value == 0L ? a : new Cell(0L); // reuse or initialize cell
+                            }
+                            // update cells and base (not atomic)
```
brian-brazil: Is this still true?

rrakos-evo: It depends on how we look at it. 😉 
It's not written atomically, but it's read "atomically". 

On concurrent `get` and `set` and `inc / dec`:
* it is possible that `set` writes `cells`, then `inc / dec` writes  `base` (no contention), then `set` writes `base`
* it is possible that `sum` reads intermediate state, but it detects this and retries, so it won't use invalid state

simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
rrakos-evo: If performance is not that important for mixed `set` and `inc / dec`, then we could just set `cells` to `null` in `set`, and let `inc / dec` reallocate cells on contention.

simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
```diff
@@ -118,6 +142,38 @@ public void reset() {
...
     }
 
+    public void set(double x) {
+        for (;;) {
+            Cell[] as;
+            if ((as = cells) != null) { // have cells
+                if (busy == 0 && casBusy()) {
+                    try {
+                        if (cells == as) { // recheck under lock
+                            // allocate new cells instead of updating existing cells
```
rrakos-evo: > Thinking about it, it'd be simpler to just set the values on the Cells - but we'd still want to keep the cas on busy around for the right semantics in sum.

I don't think we could ensure correctness in `sum` with updating existing cells in `set`, as it's not possible to take a snapshot of `cells` atomically, so there would be no way to detect the intermediate state of `cells` caused by concurrent updates of individual `cells`.

rrakos-evo:
2874c5ee:
```
Benchmark                                                 Mode  Cnt    Score    Error  Units
GaugeBenchmark.codahaleCounterDecBenchmark                avgt    4   35.740 ± 20.478  ns/op
GaugeBenchmark.codahaleCounterIncBenchmark                avgt    4   32.936 ±  1.891  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildDecBenchmark     avgt    4   36.128 ±  3.627  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildIncBenchmark     avgt    4   38.462 ±  1.829  ns/op
GaugeBenchmark.prometheusSimpleGaugeChildSetBenchmark     avgt    4  444.700 ± 97.494  ns/op
GaugeBenchmark.prometheusSimpleGaugeDecBenchmark          avgt    4  163.684 ±  4.272  ns/op
GaugeBenchmark.prometheusSimpleGaugeIncBenchmark          avgt    4  161.102 ± 12.008  ns/op
GaugeBenchmark.prometheusSimpleGaugeSetBenchmark          avgt    4  492.110 ± 50.022  ns/op
GaugeBenchmark.prometheusSimpleGaugeNoLabelsDecBenchmark  avgt    4   37.190 ± 12.319  ns/op
GaugeBenchmark.prometheusSimpleGaugeNoLabelsIncBenchmark  avgt    4   35.756 ±  5.226  ns/op
GaugeBenchmark.prometheusSimpleGaugeNoLabelsSetBenchmark  avgt    4  471.402 ± 92.968  ns/op
```

simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
brian-brazil: That would be a lot simpler.

simpleclient/src/main/java/io/prometheus/client/DoubleAdder.java
brian-brazil: The comment as written hints that there's a race here, but there isn't.

rrakos-evo:
I rebased the changes on `master`.
I'll share load testing and profiling results soon™️.

rrakos-evo:
I couldn't find `Gauge` in YourKit snapshots and screenshots from July. 🙁 
I'll need to profile the changes again, so I'll try to prove that it eliminated blocking.

rrakos-evo:
I can confirm that there's no more blocking. I've been trying to get profiling results to have something to show that proves that, but because it's so fast now it doesn't show up during sampling using a reasonable sampling rate, that does not affect performance by itself.

Do you have any further concerns about merging this?

brian-brazil:
Thanks!

rrakos-evo:
Thank you. 🙂

Issue #159, opened 2016-11-17
If parameters include '/' ,It will response code 400.


Nodeman123:
If tags parameters include '/' ,It will response code 400. How to do? PushGateway.push(CollectoRegistry,jobname,map)

brian-brazil:
This is https://github.com/prometheus/pushgateway/issues/97

brian-brazil:
This has been fixed.

PR #491, opened 2019-07-29
Handle slashes in pgw grouping label values.

brian-brazil:
Good base64 support was only added in Java 8,
we support back to Java 6 so this is a little unclean.

@beorn7


brian-brazil:
Fixes #159

brian-brazil:
Done

PR #497, opened 2019-08-26
Don't depend on obsolete sonatype-oss-parent pom

jerboaa:
See:
https://github.com/sonatype/oss-parents

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,12 +7,6 @@
     <artifactId>parent</artifactId>
     <version>0.6.1-SNAPSHOT</version>
 
-    <parent>
-        <groupId>org.sonatype.oss</groupId>
-        <artifactId>oss-parent</artifactId>
-        <version>7</version>
-    </parent>
-
     <name>Prometheus Java Suite</name>
     <url>http://github.com/prometheus/client_java</url>
     <description>


jerboaa:
@brian-brazil PTAL. Projects aren't supposed to depend on this deprecated pom. One could manage configs coming from https://github.com/sonatype/oss-parents/blob/master/oss-parent/pom.xml by including them in the `client_java/pom.xml` file. Thoughts?

brian-brazil:
I'm not familiar with this particular bit of maven, is it really recommended for every Java project in existence to copy&paste this?

jerboaa:
> I'm not familiar with this particular bit of maven, is it really recommended for every Java project in existence to copy&paste this?

Yes, that's my understanding. Consider that sonatype-oss-parent is ~7 years old and obsolete. If you need to share some config across projects consider doing something like apache-parent:

https://github.com/apache/maven-apache-parent/blob/apache-21/pom.xml
https://maven.apache.org/pom/asf/

brian-brazil:
The ASF does its own thing, I wouldn't take that as the way the central repository should be handled. Are there docs about this somewhere?

jerboaa:
@brian-brazil Not sure what the question is exactly. This perhaps? https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance_vs_Project_Aggregation

Why are you inheriting from sonatype-oss-parent? What are you trying to achieve?

jerboaa:
FWIW, https://github.com/sonatype/oss-parents also mentions:

> For deployments to the Central Repository and the needed setup, please refer to http://central.sonatype.org

brian-brazil:
This isn't causing any current issues, so I'm not going to worry about this right now.

PR #498, opened 2019-09-04
Don't sanitize away colons in metric names

ugniusb:
The [docs](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) specify that colons (`:`) are allowed in metric names, but `Collector.sanitizeMetricName` removes them.
This PR makes it keep the colons.

cc @brian-brazil


brian-brazil:
Thanks!

PR #505, opened 2019-10-07
Add a class cast check to avoid ClassCastException

### LINKED ISSUE #504: ClassCastException in MemoryAllocationExports
Hello, 
I am working with an IBM jdk which seems to have a **GarbageCollectorMXBean** not being a **NotificationEmitter**. That's why I get a **ClassCastException** on line **29** of class **MemoryAllocationExports**.
Could it be possible to check if the **GarbageCollectorMXBean** is an instance of **NotificationEmitter** before casting it ?
Regards

kebuu:
FIX #504
I added the method **getGarbageCollectorMXBeans** to be able to test my modifications (otherwise I don't know how to fake **ManagementFactory.getGarbageCollectorMXBeans()**).

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryAllocationExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryAllocationExports.java
@@ -25,8 +25,10 @@ public class MemoryAllocationExports extends Collector {
 
   public MemoryAllocationExports() {
     AllocationCountingNotificationListener listener = new AllocationCountingNotificationListener(allocatedCounter);
-    for (GarbageCollectorMXBean garbageCollectorMXBean : ManagementFactory.getGarbageCollectorMXBeans()) {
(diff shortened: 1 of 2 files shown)


simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryAllocationExports.java
```diff
@@ -97,4 +99,8 @@ private static long getAndSet(Map<String, Long> map, String key, long value) {
       return last == null ? 0 : last;
     }
   }
+
+  protected List<GarbageCollectorMXBean> getGarbageCollectorMXBeans() {
```
brian-brazil: Why did you add this function?

kebuu:
I added this function to be able to test what I coded. Otherwise I don't
know how to mock  ManagementFactory.getGarbageCollectorMXBeans().

brian-brazil:
Thanks!

PR #486, opened 2019-06-19
Zero gc labels lookup

brian-brazil:
@njhill @franz1981 @Falland 

I've finally had time to dig into the various PRs (#445 #459 #460), and what I've done is taken the benchmarks from 460 and used 445 as a base and then eliminated the remaining allocs from it. 460 still has notably better performance, however I remain uncomfortable with a bespoke hashtable implementation and Java's profiling tools aren't giving me enough to figure out why it's slower.

As a side effect, it's also simpler to add other numbers of labels in the future (not that even 4 should be common in the first place).

PR #514, opened 2019-11-22
Make Collector labels-to-child map implementation pluggable

njhill:
#460 proposed an optimized zero-GC version of the child lookup logic which was deemed too specialized for inclusion in the library.

Subjectively less complex alternatives were also proposed which provide some but not as much performance/garbage improvement, and rely for example on some per-thread overhead.

This PR aims to add a minimally-invasive mechanism to allow users to plug in an implementation of their choice, so that performance sensitive consumers can opt for minimal overhead without the core library having to include the corresponding code.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/ConcurrentChildMap.java
+++ b/simpleclient/src/main/java/io/prometheus/client/ConcurrentChildMap.java
@@ -1,0 +1,94 @@
+package io.prometheus.client;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for your PR, I don't believe this sort of thing should be pluggable - that's asking for someone to make an incompatible implementation and causing problems down the line. We should have one implementation that works well such as proposed in #486, and if someone is in a particularly niche setup they can always create custom collectors.

njhill:
@brian-brazil the only "niche" aspect of my usage is the desire for minimal runtime overhead. I am using a few of the different standard collectors and so using custom ones will mean duplicating all of them which I was hoping to avoid but I guess is the only option.

PR #527, opened 2020-02-13
Remove the mention of JMX from the included connectors

maxstreese:
This is a very minor PR but I think it is still important. Currently the README suggests that simply calling `DefaultExports.initialize()` will enable JMX metrics collection.

This is not the case though and has led to some issues opened here in that past (it did not just confuse me is what I am saying I guess). JMX metrics are not collected when calling the above. Only garbage collection, memory pools, class loading and thread counts are collected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ class YourClass {
 
 ## Included Collectors
 
-The Java client includes collectors for garbage collection, memory pools, JMX, classloading, and thread counts.
+The Java client includes collectors for garbage collection, memory pools, classloading, and thread counts.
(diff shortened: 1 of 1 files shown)


brian-brazil:
All those metrics come from JMX though, so it's not technically wrong...

maxstreese:
Hmm it is just that the way it is currently written strongly suggests to me that the call would enable collection of every and all JMX metrics and not that the metrics which are collected just happen to come via JMX. If you disagree feel free to close the PR.

brian-brazil:
Thanks!

Issue #529, opened 2020-02-14
Make HTTPMetricHandler public


janssk1:
The HTTPServer is nice little server exposing stats. 
However, it would be even better if i can easily expose metrics in a (sun) HTTP server that i control myself. Then i can setup the server the way i want (eg include other handlers)
In this case, i only need the httpmetricHandler from the prometheus client library, but unfortunately the handler is package private

brian-brazil:
Would you like to send a PR?

janssk1:
Sure

PR #530, opened 2020-02-14
#529: Expose HTTPMetricHandler

janssk1:


brian-brazil:
This change is bigger than it needs to be, you only needed to change the visibility.

janssk1:
Created it as a toplevel class. That makes it clear that the handler is a standalone class. Instantiating HttpServer.HTTPMetricHandler might indicate that a server is still involved. 
Given that there are good tests, moving a class should not be problem.

brian-brazil:
I'm not sure how good our tests are here, and I'd don't like splitting logic like this across files too much.

brian-brazil:
Given that my comments remain unaddressed I'm going to close this off. If you're still interesting in a change that only makes this public, let me know.

PR #532, opened 2020-02-26
Make it easier to customize serialization of registry data.

vim345:
This would allow applications to subclass `MetricsServlet` and access registry data to serve them in non http endpoints such as gRPC or different formats that Prometheus/gRPC based collectors use.

I wasn't too sure how to bump up the version. Do I need to bump up the version in all pom.xml files?


brian-brazil:
You shouldn't be subclassing registry, if you want to expose other formats use metricFamilySamples just like the main exposition does.

PR #534, opened 2020-02-28
Added health check api to the client.

bristy:
Added health check api to the client.
When "/-/healthy" is hit, server will return "Exporter is Healthy."
Added Unit test to verify changes are proper and no existing functionality is broken.
https://github.com/prometheus/jmx_exporter/issues/465

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -42,15 +42,13 @@ public class HTTPServer {
         }
     }
 
-    private static final String HEALTH_PATH = "/-/healthy";
-
     /**
      * Handles Metrics collections from the given registry.
      */
     static class HTTPMetricHandler implements HttpHandler {
         private CollectorRegistry registry;
         private final LocalByteArray response = new LocalByteArray();
-        private final static String RESPONSE = "Exporter is Healthy.";
+        private final static String HEALTHY_RESPONSE = "Exporter is Healthy.";
 
         HTTPMetricHandler(CollectorRegistry registry) {
           this.registry = registry;
@@ -64,9 +62,8 @@ public class HTTPServer {
             ByteArrayOutputStream response = this.response.get();
             response.reset();
             OutputStreamWriter osw = new OutputStreamWriter(response);
-            if (HEALTH_PATH.equals(contextPath)) {
-
-                osw.write(RESPONSE);
+            if ("/-/healthy".equals(contextPath)) {
+                osw.write(HEALTHY_RESPONSE);
             } else {
                 TextFormat.write004(osw,
                         registry.filteredMetricFamilySamples(parseQuery(query)));
@@ -166,7 +163,7 @@ public class HTTPServer {
         HttpHandler mHandler = new HTTPMetricHandler(registry);
         server.createContext("/", mHandler);
         server.createContext("/metrics", mHandler);
-        server.createContext(HEALTH_PATH, mHandler);
+        server.createContext("/-/healthy", mHandler);
         executorService = Executors.newFixedThreadPool(5, NamedDaemonThreadFactory.defaultThreadFactory(daemon));
         server.setExecutor(executorService);
         start(daemon);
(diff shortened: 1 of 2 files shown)


bristy:
@brian-brazil Please have a look on my PR please.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -42,9 +42,15 @@ protected ByteArrayOutputStream initialValue()
         }
     }
 
+    private static final String HEALTH_PATH = "/-/healthy";
```
brian-brazil: You can hardcode this.

bristy: I don't actually like writing hardcode magic values. So this should be fine.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -54,16 +60,21 @@ protected ByteArrayOutputStream initialValue()
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
 
+            String contextPath = t.getHttpContext().getPath();
             ByteArrayOutputStream response = this.response.get();
             response.reset();
             OutputStreamWriter osw = new OutputStreamWriter(response);
-            TextFormat.write004(osw,
-                    registry.filteredMetricFamilySamples(parseQuery(query)));
+            if(HEALTH_PATH.equals(contextPath)) {
```
Hangleton: Minor: an alternative could be to have a distinct implementation of `HttpHandler` to respond to health check. And res-use code via composition or inheritance.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -42,9 +42,15 @@ protected ByteArrayOutputStream initialValue()
...
 
+    private static final String HEALTH_PATH = "/-/healthy";
+
+    /**
+     * Handles Metrics collections from the given registry.
+     */
     static class HTTPMetricHandler implements HttpHandler {
         private CollectorRegistry registry;
         private final LocalByteArray response = new LocalByteArray();
+        private final static String RESPONSE = "Exporter is Healthy.";
```
Hangleton: Name could benefit from being more specific.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -54,16 +60,21 @@ protected ByteArrayOutputStream initialValue()
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
 
+            String contextPath = t.getHttpContext().getPath();
             ByteArrayOutputStream response = this.response.get();
```
Hangleton: Not related to this PR, but why do we cache the output stream?

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
brian-brazil: I don't like having to indirect to find simple non-magic constants.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
bristy: https://github.com/prometheus/client_java/issues/529
there is issue for this. This could be taken as part of this issue.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
bristy: I believe it because, we dont want to create new object. Anyways we are writing metrics to  bytestream for each call.

Hangleton:
o

brian-brazil:
Thanks!

PR #535, opened 2020-02-28
Add ExecutionAvgTime metrics in HibernateStatisticsCollector

dengliming:


brian-brazil:
You can already calculate this in a more useful fashion from the existing metrics.

PR #537, opened 2020-02-29
Fix #536 - Allow log4j2 appenders to have filters

### LINKED ISSUE #536: Log4j2 appender should allow filtering
Log4j2 [appenders can have a filter defined](https://logging.apache.org/log4j/2.x/manual/filters.html) on them. Currently, when this is done, the filter is ignored. This should be fixed.

travisspencer:
This PR fixes the log4j2 appender to support filtering.

### DIFF
--- a/simpleclient_log4j2/pom.xml
+++ b/simpleclient_log4j2/pom.xml
@@ -48,7 +48,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.11</version>
+            <version>4.13</version>
             <scope>test</scope>
         </dependency>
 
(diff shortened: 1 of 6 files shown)


simpleclient_log4j2/pom.xml
```diff
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-core</artifactId>
-            <version>2.1</version>
+            <version>2.13.1</version>
```
brian-brazil: We purposefully stick with older versions, so we don't accidentally depend on new features.

travisspencer: Understood. Fixed

travisspencer:
> There appear to be a number of unrelated changes in this PR, can you trim it down to what's actually required?

I can but then I'll have to make another PR to allow the counter name to be set via config. Perhaps that's best though for tracking. Will split this apart.

PR #540, opened 2020-03-12
Wrap PrintWriter with BufferedWriter

after-the-sunrise:
Hi @brian-brazil,

Generating servlet output for a large number of metrics seems to be inefficient, since the underlying `PrintWriter` implementation is generating too many `org.eclipse.jetty.io.ByteArrayBuffer` instances.

Performance improved by simply wrapping the `PrintWriter` with `BufferedWriter`. Microbenchmark (`ExampleBenchmark`, 10k gauges, 100 iterations, laptop machine) shows that it takes around 2.5 sec with the wrapper, and around 5.5 sec without the wrapper.

### DIFF
--- a/simpleclient_servlet/src/test/java/io/prometheus/client/exporter/ExampleBenchmark.java
+++ b/simpleclient_servlet/src/test/java/io/prometheus/client/exporter/ExampleBenchmark.java
@@ -23,13 +23,13 @@ public class ExampleBenchmark {
         context.setContextPath("/");
         context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
 
-        Server server = new Server(1234);
+        Server server = new Server(0);
         server.setHandler(context);
         server.start();
         Thread.sleep(1000);
 
         byte[] bytes = new byte[8192];
-        URL url = new URL("http://localhost:1234/metrics");
+        URL url = new URL("http", "localhost", server.getConnectors()[0].getLocalPort(), "/metrics");
 
         long start = System.nanoTime();
         for (int i = 0; i < 100; i++) {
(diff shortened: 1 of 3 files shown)


simpleclient_servlet/src/test/java/io/prometheus/client/exporter/ExampleBenchmark.java
```diff
@@ -0,0 +1,52 @@
...
+        Gauge gauge = Gauge.build().name("labels").help("foo").labelNames("bar").register();
+        for (int i = 0; i < 10000; i++) {
+            gauge.labels(UUID.randomUUID().toString()).set(Math.random());
+        }
+
+        ServletContextHandler context = new ServletContextHandler();
+        context.setContextPath("/");
+        context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
+
+        Server server = new Server(1234);
```
brian-brazil: Do we actually need to run a HTTP server to test this?

brian-brazil: This hasn't been addressed.

after-the-sunrise: It's required since bottleneck is container-specific. For example, with Jetty 8.1.7 the bottleneck seems to be the ByteArrayBuffer instances. But with Jetty 9.4.27, ByteArrayBuffer is no longer an issue but another part of the code is consuming CPU cycles. In either case, the wrapper patch improved the performance by approximately the same amount.

brian-brazil: Can we at least bind to 0 so we don't clash with any existing listeners?

brian-brazil:
Thanks!

after-the-sunrise:
Thanks for the review & merge :)

PR #541, opened 2020-03-18
make fields in SimpleCollector.Builder visible for other packages.

ye-will:
Fields in SimpleCollector.Builder should be 'protected' so that custom Builder class in other packages can do some validation or set dontInitializeNoLabelsChild flag. Otherwise, we have to do the dirty work in constructor of our CustomCollector.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
@@ -179,14 +179,14 @@ public abstract class SimpleCollector<Child> extends Collector {
    * Builders let you configure and then create collectors.
    */
   public abstract static class Builder<B extends Builder<B, C>, C extends SimpleCollector> {
-    String namespace = "";
-    String subsystem = "";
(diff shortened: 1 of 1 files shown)


brian-brazil:
SimpleCollector is an internal class that subclassing of by 3rd party code is not supported.

ye-will:
There's CounterMetricFamily, GaugeMetricFamily & SummaryMetricFamily for custom metrics, but none for custom Histogram. I's a lot of work if we extend the Collector.MetricFamilySamples directly...
Other ways are extending Histogram.Child or SimpleCollector, both have fields visible issue...

brian-brazil:
Given how histograms work, a HistogramMetricFamily wouldn't actually save you any code - which is why it doesn't exist.

ye-will:
Maybe fields in SimpleCollector should not be 'protected' if it's a internal class... :P

brian-brazil:
Java as far as I know doesn't offer a way to have a class only be subclassed by its package, so it's only in the javadoc.

PR #549, opened 2020-05-07
Bump up caffeine dependencies

eschizoid:
The purpose of this PR is to add support for the `AsyncCache` and not only the`AsyncLoadingCache`

### DIFF
--- a/simpleclient_caffeine/pom.xml
+++ b/simpleclient_caffeine/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
-            <version>2.3.0</version>
+            <version>2.7.0</version>
         </dependency>
         
 
--- a/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
+++ b/simpleclient_caffeine/src/main/java/io/prometheus/client/cache/caffeine/CacheMetricsCollector.java
@@ -1,6 +1,6 @@
 package io.prometheus.client.cache.caffeine;
 
-import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
+import com.github.benmanes.caffeine.cache.AsyncCache;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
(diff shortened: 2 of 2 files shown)


eschizoid:
hey @checketts wondering if you can give it a quick look to this little guy, thanks!

brian-brazil:
We purposefully list old versions so that we don't inadvertently break backwards compatibility.

eschizoid:
>We purposefully list old versions so that we don't inadvertently break backwards compatibility.

hey @brian-brazil thanks for your quick reply. Is there any other we can add support for this new type of caching that caffeine supports?

This upgrade should be breaking previous implementations since now both classes shared a common interface, hope that helps :)

brian-brazil:
So if someone was running 2.3.0, then this wouldn't break them?

eschizoid:
hey @brian-brazil attached my answers.

>So if someone was running 2.3.0, then this wouldn't break them?
Thats is correct, given the following interface:

```
package com.github.benmanes.caffeine.cache;
..
public interface AsyncLoadingCache<K, V> extends AsyncCache<K, V> {
...
}
```

Caffeine now offers some other methods that allow you to have an instance `AsyncCache`

brian-brazil:
I'm not convinced, there is no AsyncCache in 2.3.0.

eschizoid:
> I'm not convinced, there is no AsyncCache in 2.3.0.
correct, so folks that remain on version 0.8.1 won't have problems


let's say we merge this PR and folks migrate to version 0.8.2, they won't have to change their working code since `AsyncCache` is the base class for `AsyncLoadingCache `, meaning they won't have to change their code at all other than bumping the dependency.

Would it be helpful if add some explanations on the REAMDE file explaining how inheritance works in java and add some examples of how to migrate to the new version?

If you feel more comfortable we can also do this :)


```
public void addCache(String cacheName, AsyncLoadincCache cache) {
    public void addCache(String cacheName, AsyncLoadincCache cache) {
        children.put(cacheName, cache.synchronous());
    }

public void addCache(String cacheName, AsyncCache cache) {
    public void addCache(String cacheName, AsyncCache cache) {
        children.put(cacheName, cache.synchronous());
    }
```

i feel is a little bit redundant given what i explain, please let me know what you think :)

brian-brazil:
> correct, so folks that remain on version 0.8.1 won't have problems

But folks who upgrade to the next version of client_java would. I don't like forcing users to upgrade to new versions of dependencies unless there's a non-trivial gain in terms of the metrics available. Especially when the new version hasn't even been out two weeks yet.

eschizoid:
> I don't like forcing users to upgrade to new versions of dependencies unless there's a non-trivial gain in terms of the metrics available.

I see. I think t the java client will get some benefits on the metric side since now it will support the two types of caches.

> Especially when the new version hasn't even been out two weeks yet.

What about if we bump up to the version where the interface `AsyncCache` was introduced?

brian-brazil:
That sounds more reasonable.

eschizoid:
>That sounds more reasonable.
Done :)

brian-brazil:
Thanks!

PR #555, opened 2020-05-19
Avoid NullPointerException when double unregister

runzhiwang:
If unregister one collector twice, it will throw NullPointException when do for `(String name : names)`.


runzhiwang:
@brian-brazil Could you help review this patch, and release a new snapshot with this patch? Thank you very much.

brian-brazil:
Double unregistering is a programming error, so should throw an exception. In addition unregister in general should be very rarely used - basically only in unittests.

brian-brazil:
Closing, as an exception is the correct behaviour.

PR #559, opened 2020-07-13
Mockito Matchers has been deprecated and replaced with ArgumentMatchers

kdombeck:
https://github.com/mockito/mockito/blob/release/3.x/src/main/java/org/mockito/Matchers.java

Also removed a couple of unused imports in the test classes I was already modifying.


brian-brazil:
Thanks!

PR #560, opened 2020-07-15
Count HTTP statuses returned along with the HTTP response times

kdombeck:
This will add metrics for HTTP status codes like the following along side the existing Histogram.
```
# HELP http_request_status HTTP status codes of The time taken fulfilling servlet requests
# TYPE http_request_status counter
http_request_status{path="/my/api/1",method="GET",status="200",} 4.0
http_request_status{path="/my/api/2",method="GET",status="500",} 2.0
```

@brian-brazil I currently have the code deriving the **name** and **help** for the Counter from the existing parameters that are used in the constructor or init. Would you prefer I add 2 new parameters for the name and help for the counter or leave it like it is?

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
@@ -1,5 +1,6 @@
 package io.prometheus.client.filter;
 
+import io.prometheus.client.Counter;
 import io.prometheus.client.Histogram;
 
 import javax.servlet.Filter;
@@ -9,13 +10,14 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
 /**
  * The MetricsFilter class exists to provide a high-level filter that enables tunable collection of metrics for Servlet
  * performance.
  *
- * <p>The Histogram name itself is required, and configured with a {@code metric-name} init parameter.
+ * <p>The metric name itself is required, and configured with a {@code metric-name} init parameter.
  *
  * <p>The help parameter, configured with the {@code help} init parameter, is not required but strongly recommended.
  *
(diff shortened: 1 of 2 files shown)


simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -149,6 +155,10 @@ public void init(FilterConfig filterConfig) throws ServletException {
                 .help(help)
                 .name(metricName)
                 .register();
+
+        statusCounter = Counter.build(metricName + "_status", "HTTP status codes of " + help)
```
brian-brazil, line 159: Counters should end in _total

Having a label name in a metric name is generally an anti-pattern, as it won't make sense if that label is aggregated away. It may be okay here though.

kdombeck: I remove **status** from the metric name. It will now look like this
```
# HELP http_request_status HTTP status codes of The time taken fulfilling servlet requests
# TYPE http_request_status counter
http_request_total{path="/my/api/1",method="GET",status="200",} 4.0
http_request_total{path="/my/api/2",method="GET",status="500",} 2.0
```

brian-brazil: That'd end up kinda clashing with the histogram though, which isn't desirable.

kdombeck: I added **status** back in. It will now look like this.
```
# HELP http_request_status HTTP status codes of The time taken fulfilling servlet requests
# TYPE http_request_status counter
http_request_status_total{path="/my/api/1",method="GET",status="200",} 4.0
http_request_status_total{path="/my/api/2",method="GET",status="500",} 2.0
```

brian-brazil: I do want it removed, but we can't for other reasons :)

brian-brazil:
Thanks!

vizmi:
When can we expect this to hit the official maven repository? 0.9.0 is the latest I can see and it is from May 2020: https://mvnrepository.com/artifact/io.prometheus/simpleclient

PR #561, opened 2020-07-15
Add Bill of Materials

### LINKED ISSUE #508: Publish a BOM
Hey there! Love the project, but I would love it if there was a BOM file that is published for each version. If you're not familiar with a BOM,it stands for "Bill of Materials" it helps your Gradle (in my case, but it's originally a Maven thing) file look like this:

```
implementation("io.prometheus:simpleclient-bom:0.6.0")
implementation("io.prometheus:simpleclient_common")
implementation("io.prometheus:simpleclient_dropwizard")
implementation("io.prometheus:simpleclient_servlet")
```

Instead of this:

```
val prometheusVersion by extra { "0.6.0" }

implementation("io.prometheus:simpleclient:$prometheusVersion")
implementation("io.prometheus:simpleclient_common:$prometheusVersion")
implementation("io.prometheus:simpleclient_dropwizard:$prometheusVersion")
implementation("io.prometheus:simpleclient_servlet:$prometheusVersion")
```

Notice how you just leave the versions off when you use a BOM. This can help reduce the number of dependency compatibility surprises one can encounter, especially if a transitive dependency brings in a newer version of one of the components (it'll be reduced to the BOM's version).

Here's a deeper dive into how a BOM works:

https://howtodoinjava.com/maven/maven-bom-bill-of-materials-dependency/

twz123:
Fixes #508.

### DIFF
--- a/simpleclient_bom/pom.xml
+++ b/simpleclient_bom/pom.xml
@@ -29,97 +29,97 @@
             <dependency>
                 <groupId>io.prometheus</groupId>
                 <artifactId>simpleclient</artifactId>
-                <version>${project.version}</version>
+                <version>0.9.1-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>io.prometheus</groupId>
                 <artifactId>simpleclient_common</artifactId>
-                <version>${project.version}</version>
+                <version>0.9.1-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>io.prometheus</groupId>
                 <artifactId>simpleclient_caffeine</artifactId>
-                <version>${project.version}</version>
+                <version>0.9.1-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>io.prometheus</groupId>
                 <artifactId>simpleclient_dropwizard</artifactId>
-                <version>${project.version}</version>
+                <version>0.9.1-SNAPSHOT</version>
             </dependency>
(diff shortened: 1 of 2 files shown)


simpleclient_bom/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.prometheus</groupId>
+                <artifactId>simpleclient</artifactId>
+                <version>${project.version}</version>
```
brian-brazil, line 32: Please don't use variables, there's some build system out there that can't handle them.

twz123: Okay, didn't know. Which one?

I'll try to fix that. Without variables, we'll need some special handling for releases, so that versions are changed accordingly. Let's see...

brian-brazil: It was some years ago, so I don't recall offhand.

I'd expect that mvn would update everything automatically, as it already does that for the other versions.

twz123: You're right. The release plugin handles this out of the box just fine. It seems to rewrite versions for all projects in the reactor everywhere. I thought it would just update the parent version.

brian-brazil:
Thanks!

PR #564, opened 2020-07-29
Specify the charset of HTTPServer response

joaopdelboni:
In some Operational Systems (Like IBM Z/OS and IBM Z/Unix) the JRE runs with a charset different of UTF-8, those default charsets produce wrong behavior and truncated responses.

To avoid this kind of situations I put the specifc charset in the output writer.

This create a consistent response between different platforms.

### DIFF
--- a/simpleclient_graphite_bridge/src/main/java/io/prometheus/client/bridge/Graphite.java
+++ b/simpleclient_graphite_bridge/src/main/java/io/prometheus/client/bridge/Graphite.java
@@ -5,8 +5,10 @@ import io.prometheus.client.CollectorRegistry;
 
 import java.io.BufferedWriter;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.net.Socket;
+import java.nio.charset.Charset;
 import java.util.Collections;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


brian-brazil:
Is there anywhere else we need this?

joaopdelboni:
> Is there anywhere else we need this?

Only this class for the tested use cases. But this class are used by jmx_exporter (jmx_prometheus_javaagent and jmx_prometheus_httpserver) those projects will be affected by this change.

brian-brazil:
We've other places we produce HTTP requests/responses with metrics, would you mind checking them?

joaopdelboni:
@brian-brazil sorry for the delay, I've made a review on the other projects of the repository and thats my view:
- SpringBoot: The spring boot project made the enconding based on request encoding header. So, we don't need specify this type of detail;
- VertX: VertX respect the same behavior of spring.
- Servlet: The metrics servlet endpoint uses the default ServeletResponse buffer, so we don't need to made any type of ajustment.
- Graphite Bridge:  I made a change to include charset on the request. To prevent the same issues related before.
- Push Gateway: This project specify the enconding, so .. is all right here 
- LogBack, Log4j2, Log4j, Hotspot, SpringWeb, etc..: They don't expose any type of metrics http endpoints.

brian-brazil:
> The spring boot project made the enconding based on request encoding header. 

We should always expose UTF-8, no matter what the request headers say.

joaopdelboni:
Ok Brian,

I made a mistake on the previous description, the spring boot have the property named _spring.http.encoding.charset_, this property have a default value defined to UTF-8.
The only way to spring return something in other encoding is changing this property value.

I can made a change to in the metrics specific endpoint aways return values encoded in UTF-8, this will result in a method with raw bytearray return. This not smell good. But i can do.

What do you think about ?

brian-brazil:
It sounds like Spring do the right thing by default, so we could leave that until it's actually a problem for someone.

joaopdelboni:
Great, so at this point it is all changes done.

brian-brazil:
Thanks!

PR #566, opened 2020-08-08
Verify exception messages in tests

stefanbirkner:
Improve test coverage and make refactorings more safe.

@brian-brazil


brian-brazil:
Thanks!

PR #572, opened 2020-10-05
Fix code example for YourCustomCollector in README

charneykaye:
In order to run the example, this method must have public visibility.


brian-brazil:
Thanks!

PR #575, opened 2020-10-08
Provide JVM Mixin.

jeschkies:
This proposes a [monitoring mixing](https://monitoring.mixins.dev/) based on the dashboard
https://grafana.com/grafana/dashboards/3066 by @jotak.

The datasource is templated in addition to the original dashboard.

Outstanding:
- [ ] Alerts
- [ ] CI integration

### DIFF
--- a/mixin/README.md
+++ b/mixin/README.md
@@ -1,0 +1,6 @@
+# JVM Mixin 
+
+This mixin works with the [Java client](https://github.com/prometheus/client_java). It requires each JVM application to export the metrics with the client.
+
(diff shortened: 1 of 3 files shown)


jeschkies:
There are jsonnet formatting and linting tools. However, they require go. How should I integrate them? Also, should I provide an example service?

brian-brazil:
Thanks for your PR. I'm not sure it makes sense to define alerts and dashboards for a client library, the Java build ecosystem is fun enough without mixing in a completely different language in, and I'm overall not looking to add more to maintain to this repo.

I'd suggest putting this in some 3rd party repo and we could link from the readme for it. I also note that raw counters aren't useful to graph, you should be taking a rate of the GC count.

jeschkies:
Won't do

Issue #576, opened 2020-10-11
Would it make sense to add to Default exporters JVM's Safepoint metrics?


mr-git:
Would it make sense to add to Default exporters JVM's Safepoint metrics?

`HotspotRuntimeMBean` exposes:
* `getSafepointCount()`
* `getSafepointSyncTime()`
* `getTotalSafepointTIme()`

In applications with many threads and long computations, these metrics can manifest how much time is spent before GC to get all threads to Safepoint and pause it.

brian-brazil:
This sounds reasonable to add, though we'd need to check that they work on the various JVM implementations.

darkl0rd:
Any idea when this functionality is going to be added? In the interim, is there a way to manually add these to the JMX exporter?

dhoard:
I have some test code that calls `sun.management.HotspotRuntimeMBean` via reflection and support is not great.

For Java 6 OpenJDK-based VMs...

- **fails**

For Java 8 OpenJDK-based VMs...

- **works**

For Java 11 OpenJDK-based VMs...

- **works with a warning**

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by io.prometheus.jmx.collector.HotSpotCollector (file:/temp/jmx_prometheus_javaagent.jar) to method sun.management.ManagementFactoryHelper.getHotspotRuntimeMBean()
WARNING: Please consider reporting this to the maintainers of io.prometheus.jmx.collector.HotSpotCollector
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

For Java 17 OpenJDK-based VMs...

- **fails**

```
java.lang.IllegalAccessException: class io.prometheus.jmx.collector.HotSpotCollector cannot access class sun.management.ManagementFactoryHelper (in module java.management) because module java.management does not export sun.management to unnamed module @7cf10a6f
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:392)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:674)
	at java.base/java.lang.reflect.Method.invoke(Method.java:560)
	at io.prometheus.jmx.collector.HotSpotCollector.accessHotspotRuntimeMBean(HotSpotCollector.java:68)
	at io.prometheus.jmx.collector.HotSpotCollector.<init>(HotSpotCollector.java:48)
[...]

PR #577, opened 2020-10-11
#576 add SafepointExports

mr-git:
seems that the management API has been around since at least JDK 6 (http://srcrr.com/java/oracle/openjdk/6/reference/sun/management/HotspotRuntimeMBean-source.html)

Sadly, I cannot get Maven to compile locally, I suspect, that there is PR build before it gets merged


mr-git:
@brian-brazil, please review - my local `mvn` is broken, but I checked the management API - it works at least on OpenJDK 14 and GraalVM 20.0.0 (java 1.8.0_242)

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/SafepointExports.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  public List<MetricFamilySamples> collect() {
+    CounterMetricFamily safepointCount = new CounterMetricFamily(
+        "jvm_safepoint_count",
+        "The number of safepoints taken place since the JVM started.",
+            hotspotRuntimeMBean.getSafepointCount());
+
+    CounterMetricFamily safepointTime = new CounterMetricFamily(
+        "jvm_safepoint_total_time_seconds",
```
brian-brazil, line 45: The total here is a little confusing, I'd remove it.

Might these two make more sense as one Summary?

simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/SafepointExports.java
```diff
@@ -0,0 +1,61 @@
...
+    this(ManagementFactoryHelper.getHotspotRuntimeMBean());
+  }
+
+  SafepointExports(HotspotRuntimeMBean hotspotRuntimeMBean) {
+    this.hotspotRuntimeMBean = hotspotRuntimeMBean;
+  }
+
+  public List<MetricFamilySamples> collect() {
+    CounterMetricFamily safepointCount = new CounterMetricFamily(
+        "jvm_safepoint_count",
```
brian-brazil, line 40: Counters should end in _total.

brian-brazil (commented):
Have you tested it on the IBM JVM? It's not the hotspot ones I'm worried about.

mr-git:
>Have you tested it on the IBM JVM? It's not the hotspot ones I'm worried about.
No, I haven't :(
Which IBM JVM would you recommend to try?

I found some links, that IBM is contributing to J9 and there is hotspot runtime mbean in place:
https://github.com/search?q=org%3AAdoptOpenJDK+getHotspotRuntimeMBean&type=code

brian-brazil:
> Which IBM JVM would you recommend to try?

Whatever is easiest to get your hands on.

PR #578, opened 2020-10-12
Reduce copying and memory usage in HttpServer

sfriberg:
Reduce copying by directly writing to body.
Reduce memory usage by avoiding keeping ByteBuffer alive in memory.

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -5,6 +5,7 @@ import io.prometheus.client.exporter.common.TextFormat;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


brian-brazil:
We need to wait for write004 to complete to see if there's an exception before sending the response, does this preserve that?

sfriberg:
if write004 throws an exception that should stop the output and get propagated to the next level for error handling. So would expect that HTTP error codes are set like before.

Is there a dependency that it does not write anything in case of error?
This would potentially start streaming output until the error is hit and no further writes would happen.

brian-brazil:
> This would potentially start streaming output until the error is hit and no further writes would happen after that and http error code should be set.

You can't send a HTTP status code once you've started sending output, as the status code is the very first thing sent.

sfriberg:
Let me do a couple of tests and see how things behave if you throw an exception while writing.

sfriberg:
Played around a bit with the JDK basic HttpServer and indeed as you write to the stream data is sent as buffers fills up or you flush the stream.

Is it a hard requirement not to send any data if part of the response creation fails?

Since the output would be chunked the scraper would detect that not all data was not received. So it would be the same case if there for some reason was some exception when writing the ByteBuffer (or jvm exiting). Surely lower probability than encoding, but would still need to be handled and treated as a failed scrape? Would this be something that can be relied upon?

brian-brazil:
> Is it a hard requirement not to send any data if part of the response creation fails?

Yes, partial responses are very hard to deal with for end users.

sfriberg:
Too bad, will close this case and see if there is a different way to improve this.

Issue I'm hitting is with a service capturing a large number of metrics so need to run with a pretty large heap to ensure things do not OOM.

Having scrapers run with compression/gzip should reduce the total used heap, but would have some risk that any scrape without compression would result in the ByteArrayOutputStream would explode in size.

PR #598, opened 2020-10-14
Make sure GZIPOutputStream is always closed

turbanoff:
Leave GZIPOutputStream open is dangerous as it leads to native memory leak.
Use try-finally to make sure stream is always closed

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -10,20 +10,20 @@ import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
 import java.net.URLDecoder;
 import java.nio.charset.Charset;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.HashSet;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.zip.GZIPOutputStream;
 
+import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpHandler;
 import com.sun.net.httpserver.HttpServer;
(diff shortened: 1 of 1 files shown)


simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -47,41 +48,42 @@ protected ByteArrayOutputStream initialValue()
...
             if ("/-/healthy".equals(contextPath)) {
                 osw.write(HEALTHY_RESPONSE);
             } else {
                 TextFormat.write004(osw,
                         registry.filteredMetricFamilySamples(parseQuery(query)));
             }
 
-            osw.flush();
             osw.close();
-            response.flush();
```
brian-brazil, line 75: Is it safe to remove all of these?

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -47,41 +48,42 @@ protected ByteArrayOutputStream initialValue()
      * Handles Metrics collections from the given registry.
      */
     static class HTTPMetricHandler implements HttpHandler {
-        private CollectorRegistry registry;
+        private final CollectorRegistry registry;
         private final LocalByteArray response = new LocalByteArray();
         private final static String HEALTHY_RESPONSE = "Exporter is Healthy.";
+        private final Charset UTF_8 = Charset.forName("UTF-8");
```
brian-brazil, line 54: This new indirection just makes the code harder to read for me.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
turbanoff: 1. `flush` call before `close` is useless. According to javadoc - `close` should call `flush` first.
2. `ByteArrayOutputStream.flush` and `ByteArrayOutputStream.close` are no-op. So it's safe to remove them.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
brian-brazil: This is still here.

brian-brazil (commented):
Can you also add the DCO?

brian-brazil:
Thanks!

PR #599, opened 2020-10-14
Unregistering a collector that is present at the registry should not throw an NPE

marcospereira:
@brian-brazil got this in a system that was using `clear` to _shutdown_  the `registry`, but another place was `unregister`ing the metric, and the getting an NPE.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -67,6 +67,7 @@ public class CollectorRegistry {
   public void unregister(Collector m) {
     synchronized (namesCollectorsLock) {
       List<String> names = collectorsToNames.remove(m);
+      if (names == null) return;
       for (String name : names) {
         namesToCollectors.remove(name);
       }
--- a/simpleclient/src/test/java/io/prometheus/client/CollectorRegistryTest.java
+++ b/simpleclient/src/test/java/io/prometheus/client/CollectorRegistryTest.java
@@ -139,6 +139,13 @@ public class CollectorRegistryTest {
     registry.register(h);
   }
 
+  @Test
+  public void testUnregisterWhenItIsNotRegisteredIsANoOp() {
+    Histogram h = Histogram.build().name("s").help("h").create();
+    // This doesn't throw.
+    registry.unregister(h);
+  }
+
   class MyCollector extends Collector {
     public List<MetricFamilySamples> collect() {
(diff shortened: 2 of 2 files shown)


brian-brazil:
Thowing an exception seems appropriate to me if you're trying to remove something that isn't there, as that indicates a programming error. Both clear and unregister should be so rarely used that this should never come up in practices - they're only for unittests basically.

marcospereira:
> Throwing an exception seems appropriate to me if you're trying to remove something that isn't there, as that indicates a programming error.

Hum, but `CollectorRegistry` does not have a way to check if the collector is there before removing it. And in my case, it was a concurrency issue (one place calls `clear`, and another tries to `unregister` a collector). And while this concurrency access can also be considered a programming error, throwing an NPE does not sound like the correct exception to report. Maybe an `IllegalArgumentException` would be more informative.

To give you more context, this is not a regular application, but instead, it is an instrumentation product that provides an API with multiple implementations (to support distinct metrics backends). And users can create new metrics using the abstraction layer, but they can also "cleanup" such metrics in some cases.

brian-brazil:
> Hum, but CollectorRegistry does not have a way to check if the collector is there before removing it. 

You should always know what is or is not registered, the reason you generally wouldn't is if you're following certain anti-patterns around how to use direct instrumentation which also happen to be inefficient.

> And users can create new metrics using the abstraction layer, but they can also "cleanup" such metrics in some cases.

That's a bit more complex than the usual use case. In general removing existing metrics in such a case is an anti-pattern. Why do you need a clear though?

marcospereira:
> You should always know what is or is not registered, the reason you generally wouldn't is if you're following certain anti-patterns around how to use direct instrumentation which also happen to be inefficient.

Can you elaborate a little more about what do you mean by "direct instrumentation"?

The product I'm working on is an agent that instruments code and integrates with multiple backends. Meaning that it can send metrics to NewRelic, Datadog, Prometheus, and others. For some, having a way to destroy/remove the metrics is necessary in order to release resources when those metrics aren't necessary anymore.

> throwing an NPE does not sound like the correct exception to report. Maybe an `IllegalArgumentException` would be more informative.

What do you think about this?

brian-brazil:
> Can you elaborate a little more about what do you mean by "direct instrumentation"?

Roughly anything that's not a custom collector.

> For some, having a way to destroy/remove the metrics is necessary in order to release resources when those metrics aren't necessary anymore.

That a resource has gone away doesn't make counters about it invalid.

> What do you think about this?

It doesn't sound unreasonable.

brian-brazil:
Closing per above, throwing an exception is a correct behaviour in the face of programming error.

Issue #573, opened 2020-10-07
Add ability to filter metrics in DropwizardExports


vgv:
We are migrating from dropwizard metrics to prometheus right now. We exported all old dropwizard metrics to prometheus and added all metrics to prometheus one by one. However, there is a problem – metric is added to prometheus is exported via DropwizardExports too, so, we see two similar metrics in output. 

It will be great if we can filter metrics are exported to Dropwizard by standard MetricFilter mechanism.

vladimir-bukhtoyarov:
If I understand correctly this fix has not been released. So, bellow is a powerful workaround, that can be used now:

```java
public static MetricRegistry decorateMetricRegistryByFilter(MetricRegistry sourceRegistry, MetricFilter filter) {
        MetricRegistry decoratedRegistry = new MetricRegistry();

        MetricRegistryListener listener = new MetricRegistryListener() {
            @Override
            public void onGaugeAdded(String name, Gauge<?> gauge) {
                if (filter.matches(name, gauge)) {
                    decoratedRegistry.register(name, gauge);
                }
            }

            @Override
            public void onGaugeRemoved(String name) {
                decoratedRegistry.remove(name);
            }

            @Override
            public void onCounterAdded(String name, com.codahale.metrics.Counter counter) {
                if (filter.matches(name, counter)) {
                    decoratedRegistry.register(name, counter);
                }
            }

            @Override
            public void onCounterRemoved(String name) {
                decoratedRegistry.remove(name);
            }

            @Override
            public void onHistogramAdded(String name, Histogram histogram) {
                if (filter.matches(name, histogram)) {
                    decoratedRegistry.register(name, histogram);
                }
            }

            @Override
            public void onHistogramRemoved(String name) {
                decoratedRegistry.remove(name);
            }

            @Override
            public void onMeterAdded(String name, Meter meter) {
                if (filter.matches(name, meter)) {
                    decoratedRegistry.register(name, meter);
                }
            }

            @Override
            public void onMeterRemoved(String name) {
                decoratedRegistry.remove(name);
            }

            @Override
            public void onTimerAdded(String name, Timer timer) {
                if (filter.matches(name, timer)) {
                    decoratedRegistry.register(name, timer);
                }
            }

            @Override
            public void onTimerRemoved(String name) {
                decoratedRegistry.remove(name);
            }
        };
        sourceRegistry.addListener(listener);

        return decoratedRegistry;
    }
```
Then all we need is passing the decorated registry to the instead of the original.

[...]

PR #601, opened 2020-11-13
#573 optimize filtering algorithm for DropwizardExports

vladimir-bukhtoyarov:
The intention of this optimization has been described there https://github.com/prometheus/client_java/issues/573#issuecomment-726971059


brian-brazil:
Do you have benchmarks to demonstrate how useful this is?

Also please use explicit imports, it's hard to follow what is what otherwise.

vladimir-bukhtoyarov:
I can not imagine how to write the benchmark here. The penalty of filtering depends on two factors:
* Complexity of filter itself.
* Amount of metrics in the registry.

This pull request allows applying filter only one for each metric, filter does not need to be applied on each reporting cycle. For example, if we have 3 regular expressions in the filter, and 5000 metrics in the registry, then 15000 regexps are not executed on each reporting cycle, because pre-filtering of registry makes this work unnecessary. A huge metric-registry with thousands of metrics is a typical case for systems like Cassandra or Kafka.

brian-brazil:
It'd be good to have some rough idea, for example if it turns out that other processing is substantially larger then there's not much point in having this from a complexity standpoint.

I also presume it's safe to assume that filters don't vary over time?

vladimir-bukhtoyarov:
Looks I can just continue to wrap the registry by decorator before passing to the collector.

PR #606, opened 2020-11-28
Make DropwizardExports methods protected

Alexey1Gavrilov:
Allow overriding DropwizardExports methods in subclasses.

The purpose of this change to be able easily customize the format
of Dropwizard metrics in Prometheus, for example, to reduce the number
of quintiles from histograms.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -83,7 +83,7 @@ public class DropwizardExports extends io.prometheus.client.Collector implements
     /**
      * Export counter as Prometheus <a href="https://prometheus.io/docs/concepts/metric_types/#gauge">Gauge</a>.
      */
-    MetricFamilySamples fromCounter(String dropwizardName, Counter counter) {
+    protected MetricFamilySamples fromCounter(String dropwizardName, Counter counter) {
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. I don't think I want this class subclassed, as I can't guarantee things like a stable API. If you only want to drop quantiles, you can always use metric_relabel_configs on the Prometheus side. The existing sample builder logic should also cover  any renaming or related tweaks to metrics.

Alexey1Gavrilov:
Thank you for the quick reply!
We'd like to be able to filter out unused quantiles on the client side to avoid unnecessary load in the Prometheus server while scraping the clients. We care only about 0.5 and 0.99 quantiles, dropping the rest currently using not very elegant solution with a custom sample builder matching unwanted samples by a label name/value.

We'll check out `metric_relabel_configs`, thank for the pointer.

brian-brazil:
It should be quite cheap on the Prometheus side, though unless you're getting into hundreds of thousands of metrics it's unlikely you'd even notice all of the quantiles being ingested in the first place.

Alexey1Gavrilov:
Each of our service instances normally exports around 5K of metrics. Having those extra 4 quantiles making it 20K, and with 5 running instances reaches a hundred thousand which is not a problem but it requires slightly bigger Prometheus server instance to avoid crashing.

It would be nice to have a way to control how Dropwizard's histograms are presented in Prometheus, a protected method of some sort.

brian-brazil:
As I said you can use metric_relabel_configs. If you want finer control I'd suggest using the client library directly rather than via Dropwizard, it's faster generally and better aligned with the Prometheus data model.

brian-brazil:
Closing per above.

PR #608, opened 2020-12-01
Extend DropwizardExports

catap:
Dropwizard historgram and meter provides a bit more information that is quite useful like `stdDev` and precomputed rate for `1m`, `5m` and `15m`.

Let export it also.


brian-brazil:
Thanks for your PR. These are excluded on purpose, see https://prometheus.io/docs/instrumenting/writing_exporters/#drop-less-useful-statistics

catap:
@brian-brazil very good point.

Gr33nbl00d:
Quite igorant less usefull does not mean useless. I prefer to have a median rather than just mean value. I dont like software that tries to educate me and denys me to do things for no reason. Its easy to add these values back to dropwizward export i will post a code example tomorrow. Good reason to never include prometheus client library in our software ;)

PR #611, opened 2021-01-03
java's method Double.toString is able to convert not-a-number values to "NaN" string

turbanoff:


brian-brazil:
Can you add the DCO?

brian-brazil:
Thanks!

PR #614, opened 2021-01-18
Support custom metrics contexts

thanosIrodotou:
Hi there,

We recently starting using thie lib which is a life saver but has one small limitation which I'm addressing here.
That's the ability to support a user-provided metrics context.
At the moment the handler is instantiated in the constructor and there's no accessor, so retrieving metrics is only available when hitting "/metrics" which I guess should be fine for most people.

Unfortunately there are use cases where a prometheus configuration is outside of one's control and Prometheus is then configured to look for and scrape for metrics under a custom path, i.e. "/metricz" is a common one. Hence, the changes here add support for using something other than just "/metrics".

@brian-brazil  hope this makes sense and is in accordance with the contributing guidelines.

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -155,27 +155,43 @@ public class HTTPServer {
 
     /**
      * Start a HTTP server serving Prometheus metrics from the given registry using the given {@link HttpServer}.
-     * The {@code httpServer} is expected to already be bound to an address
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks for your PR. This code is intended as an easy to get up an running quickly, not as a fully general HTTP serving solution. If you've need for something more niche, I'd suggest working from whatever your existing HTTP server approach is.

Also, it will already serve on /metricz.

thanosIrodotou:
ok thanks Brian, noted. Looks like you're registering "/" too so that should allow all paths anyway. Closing this, cheers.

PR #615, opened 2021-01-18
Convert client_java to OpenMetrics

brian-brazil:
This switches the core data model to OM, adds support for negotiating and exposing OpenMetrics format, adds units, created time, and adds new direct instrumentation for Info and Enum.

This is done as gracefully as practical, but this will break users - in particular if a) they have counters lacking a `_total` suffix in which case the sample names will change or b) if they happen to be using `UNTYPED` which is now called `UNKNOWN` in which case they'll need up update their code.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -1,6 +1,7 @@
 
 package io.prometheus.client;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Pattern;
 
@@ -46,11 +47,27 @@ public abstract class Collector {
       if ((type == Type.INFO || type == Type.STATE_SET) && !unit.isEmpty()) {
         throw new IllegalArgumentException("Metric is of a type that cannot have a unit: " + name);
       }
+      List<Sample> mungedSamples = samples;
+      // Deal with _total from pre-OM automatically.
+      if (type == Type.COUNTER) {
+        if (name.endsWith("_total")) {
+          name = name.substring(0, name.length() - 6);
+        }
+        String withTotal = name + "_total";
+        mungedSamples = new ArrayList<Sample>(samples.size());
+        for (Sample s: samples) {
+          String n = s.name;
+          if (name.equals(n)) {
+            n = withTotal;
+          }
+          mungedSamples.add(new Sample(n, s.labelNames, s.labelValues, s.value, s.timestampMs));
+        }
+      }
       this.name = name;
       this.unit = unit;
       this.type = type;
       this.help = help;
-      this.samples = samples;
+      this.samples = mungedSamples;
     }
 
     public MetricFamilySamples(String name, Type type, String help, List<Sample> samples) {
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -20,7 +20,7 @@ public abstract class Collector {
    */
   public abstract List<MetricFamilySamples> collect();
   public enum Type {
-    UNTYPED, // XXX This is Unknown in OpenMetrics.
+    UNKNOWN, // This is untyped in Prometheus text format.
     COUNTER,
     GAUGE,
     STATE_SET,
--- a/simpleclient/src/main/java/io/prometheus/client/Collector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Collector.java
@@ -19,11 +19,14 @@ public abstract class Collector {
    */
   public abstract List<MetricFamilySamples> collect();
   public enum Type {
+    UNTYPED, // XXX This is Unknown in OpenMetrics.
     COUNTER,
     GAUGE,
-    SUMMARY,
+    STATE_SET,
+    INFO,
     HISTOGRAM,
-    UNTYPED,
+    GAUGE_HISTOGRAM,
+    SUMMARY,
   }
 
   /**
@@ -31,15 +34,27 @@ public abstract class Collector {
    */
   static public class MetricFamilySamples {
     public final String name;
+    public final String unit;
     public final Type type;
     public final String help;
     public final List<Sample> samples;
 
-    public MetricFamilySamples(String name, Type type, String help, List<Sample> samples) {
+    public MetricFamilySamples(String name, String unit, Type type, String help, List<Sample> samples) {
+      if (!unit.isEmpty() && !name.endsWith("_" + unit)) {
+        throw new IllegalArgumentException("Metric's unit is not the suffix of the metric name: " + name);
+      }
+      if ((type == Type.INFO || type == Type.STATE_SET) && !unit.isEmpty()) {
+        throw new IllegalArgumentException("Metric is of a type that cannot have a unit: " + name);
+      }
       this.name = name;
+      this.unit = unit;
       this.type = type;
       this.help = help;
       this.samples = samples;
+    }
+
+    public MetricFamilySamples(String name, Type type, String help, List<Sample> samples) {
+      this(name, "", type, help, samples);
     }
 
     @Override
@@ -49,14 +64,18 @@ public abstract class Collector {
       }
       MetricFamilySamples other = (MetricFamilySamples) obj;
       
-      return other.name.equals(name) && other.type.equals(type)
-        && other.help.equals(help) && other.samples.equals(samples) ;
+      return other.name.equals(name)
+        && other.unit.equals(unit)
+        && other.type.equals(type)
+        && other.help.equals(help)
+        && other.samples.equals(samples);
     }
 
     @Override
     public int hashCode() {
       int hash = 1;
       hash = 37 * hash + name.hashCode();
+      hash = 37 * hash + unit.hashCode();
       hash = 37 * hash + type.hashCode();
       hash = 37 * hash + help.hashCode();
       hash = 37 * hash + samples.hashCode();
@@ -65,7 +84,7 @@ public abstract class Collector {
 
     @Override
     public String toString() {
-      return "Name: " + name + " Type: " + type + " Help: " + help + 
+      return "Name: " + name + " Unit:" + unit + " Type: " + type + " Help: " + help +
         " Samples: " + samples;
     }
 
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -105,6 +105,11 @@ public class CollectorRegistry {
     List<String> names = new ArrayList<String>();
     for (Collector.MetricFamilySamples family : mfs) {
       switch (family.type) {
+        case COUNTER:
+          names.add(family.name + "_total");
+          names.add(family.name + "_created");
+          names.add(family.name);
+          break;
         case SUMMARY:
           names.add(family.name + "_count");
           names.add(family.name + "_sum");
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -113,11 +113,19 @@ public class CollectorRegistry {
         case SUMMARY:
           names.add(family.name + "_count");
           names.add(family.name + "_sum");
+          names.add(family.name + "_created");
           names.add(family.name);
           break;
         case HISTOGRAM:
           names.add(family.name + "_count");
           names.add(family.name + "_sum");
+          names.add(family.name + "_bucket");
+          names.add(family.name + "_created");
+          names.add(family.name);
+          break;
+        case GAUGE_HISTOGRAM:
+          names.add(family.name + "_gcount");
+          names.add(family.name + "_gsum");
           names.add(family.name + "_bucket");
           names.add(family.name);
           break;
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -129,6 +129,10 @@ public class CollectorRegistry {
           names.add(family.name + "_bucket");
           names.add(family.name);
           break;
+        case INFO:
+          names.add(family.name + "_info");
+          names.add(family.name);
+          break;
         default:
           names.add(family.name);
       }
--- a/simpleclient/src/main/java/io/prometheus/client/Counter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Counter.java
@@ -74,6 +74,10 @@ public class Counter extends SimpleCollector<Counter.Child> implements Collector
   public static class Builder extends SimpleCollector.Builder<Builder, Counter> {
     @Override
     public Counter create() {
+      // Gracefully handle pre-OpenMetrics counters.
+      if (name.endsWith("_total")) {
+        name = name.substring(0, name.length() - 6);
+      }
       return new Counter(this);
     }
   }
@@ -158,7 +162,7 @@ public class Counter extends SimpleCollector<Counter.Child> implements Collector
   public List<MetricFamilySamples> collect() {
     List<MetricFamilySamples.Sample> samples = new ArrayList<MetricFamilySamples.Sample>(children.size());
     for(Map.Entry<List<String>, Child> c: children.entrySet()) {
-      samples.add(new MetricFamilySamples.Sample(fullname, labelNames, c.getKey(), c.getValue().get()));
+      samples.add(new MetricFamilySamples.Sample(fullname + "_total", labelNames, c.getKey(), c.getValue().get()));
     }
     return familySamplesList(Type.COUNTER, samples);
   }
--- a/simpleclient/src/main/java/io/prometheus/client/Counter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Counter.java
@@ -64,6 +64,12 @@ import java.util.Map;
  * </pre>
  * These can be aggregated and processed together much more easily in the Prometheus
  * server than individual metrics for each labelset.
+ *
+ * If there is a suffix of <code>_total</code> on the metric name, it will be
+ * removed. When exposing the time series for counter value, a
+ * <code>_total</code> suffix will be added. This is for compatibility between
+ * OpenMetrics and the Prometheus text format, as OpenMetrics requires the
+ * <code>_total</code> suffix.
  */
 public class Counter extends SimpleCollector<Counter.Child> implements Collector.Describable {
 
--- a/simpleclient/src/main/java/io/prometheus/client/Counter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Counter.java
@@ -112,6 +112,7 @@ public class Counter extends SimpleCollector<Counter.Child> implements Collector
    */
   public static class Child {
     private final DoubleAdder value = new DoubleAdder();
+    private final long created = System.currentTimeMillis();
     /**
      * Increment the counter by 1.
      */
@@ -133,6 +134,12 @@ public class Counter extends SimpleCollector<Counter.Child> implements Collector
      */
     public double get() {
       return value.sum();
+    }
+    /**
+     * Get the created time of the counter in milliseconds.
+     */
+    public long created() {
+      return created;
     }
   }
 
@@ -163,6 +170,7 @@ public class Counter extends SimpleCollector<Counter.Child> implements Collector
     List<MetricFamilySamples.Sample> samples = new ArrayList<MetricFamilySamples.Sample>(children.size());
     for(Map.Entry<List<String>, Child> c: children.entrySet()) {
       samples.add(new MetricFamilySamples.Sample(fullname + "_total", labelNames, c.getKey(), c.getValue().get()));
+      samples.add(new MetricFamilySamples.Sample(fullname + "_created", labelNames, c.getKey(), c.getValue().created() / 1000.0));
     }
     return familySamplesList(Type.COUNTER, samples);
   }
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
@@ -216,7 +216,7 @@ public abstract class SimpleCollector<Child> extends Collector {
       return (B)this;
     }
     /**
-     * Set the uit of the metric. Required.
+     * Set the unit of the metric. Required.
      */
     public B unit(String unit) {
       this.unit = unit;
--- a/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
@@ -49,6 +49,7 @@ import java.util.List;
 public abstract class SimpleCollector<Child> extends Collector {
   protected final String fullname;
   protected final String help;
+  protected final String unit;
   protected final List<String> labelNames;
 
   protected final ConcurrentMap<List<String>, Child> children = new ConcurrentHashMap<List<String>, Child>();
@@ -145,7 +146,7 @@ public abstract class SimpleCollector<Child> extends Collector {
   protected abstract Child newChild();
 
   protected List<MetricFamilySamples> familySamplesList(Collector.Type type, List<MetricFamilySamples.Sample> samples) {
-    MetricFamilySamples mfs = new MetricFamilySamples(fullname, type, help, samples);
+    MetricFamilySamples mfs = new MetricFamilySamples(fullname, unit, type, help, samples);
     List<MetricFamilySamples> mfsList = new ArrayList<MetricFamilySamples>(1);
     mfsList.add(mfs);
     return mfsList;
@@ -159,6 +160,10 @@ public abstract class SimpleCollector<Child> extends Collector {
     }
     if (!b.namespace.isEmpty()) {
       name = b.namespace + '_' + name;
+    }
+    unit = b.unit;
+    if (!unit.isEmpty() && !name.endsWith("_" + unit)) {
+      name += "_" + unit;
     }
     fullname = name;
     checkMetricName(fullname);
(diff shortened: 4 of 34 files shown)


simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -19,27 +20,58 @@
...
+      if (!unit.isEmpty() && !name.endsWith("_" + unit)) {
+        throw new IllegalArgumentException("Metric's unit is not the suffix of the metric name: " + name);
+      }
+      if ((type == Type.INFO || type == Type.STATE_SET) && !unit.isEmpty()) {
+        throw new IllegalArgumentException("Metric is of a type that cannot have a unit: " + name);
+      }
+      List<Sample> mungedSamples = samples;
+      // Deal with _total from pre-OM automatically.
+      if (type == Type.COUNTER) {
+        if (name.endsWith("_total")) {
```
robskillington, line 53: nit: Worth making `_total` a const? Then can `name.endsWith(counterSuffix)` and `name = name.substring(0, name.length() - counterSuffix.length())`.

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -19,27 +20,58 @@
...
+          name = name.substring(0, name.length() - 6);
+        }
+        String withTotal = name + "_total";
```
robskillington, line 56: nit: This could also use the const.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -105,15 +105,32 @@ public void clear() {
...
           names.add(family.name);
           break;
         case HISTOGRAM:
           names.add(family.name + "_count");
           names.add(family.name + "_sum");
           names.add(family.name + "_bucket");
+          names.add(family.name + "_created");
+          names.add(family.name);
+          break;
+        case GAUGE_HISTOGRAM:
```
robskillington, line 126: Believe gauge histogram should also have `_created`?

simpleclient/src/main/java/io/prometheus/client/Counter.java
```diff
@@ -74,6 +74,10 @@
   public static class Builder extends SimpleCollector.Builder<Builder, Counter> {
     @Override
     public Counter create() {
+      // Gracefully handle pre-OpenMetrics counters.
+      if (name.endsWith("_total")) {
+        name = name.substring(0, name.length() - 6);
```
robskillington, line 79: nit: These could reference the `_total` const?

simpleclient/src/main/java/io/prometheus/client/Counter.java
```diff
@@ -130,6 +135,12 @@ public void inc(double amt) {
     public double get() {
       return value.sum();
     }
+    /**
+     * Get the created time of the counter in milliseconds.
+     */
+    public long created() {
```
robskillington, line 141: Should this maybe be a float in seconds? That way callers don't need to divide by 1000.0 to get the value they should expose, also since this accessor doesn't have the unit (seconds/milliseconds/etc) maybe it's best to return the unit that's expected to be exposed for exposition?

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -209,6 +215,13 @@ public B namespace(String namespace) {
       this.namespace = namespace;
       return (B)this;
     }
+    /**
+     * Set the uit of the metric. Required.
```
robskillington, line 219: Believe `uit` should be `unit`?

simpleclient/src/test/java/io/prometheus/client/SimpleCollectorTest.java
```diff
@@ -145,6 +145,15 @@ public void testReservedLabelNameThrows() {
     Gauge.build().name("a").labelNames("__name__").help("h").create();
   }
 
+  @Test
+  public void testUnitsAdded() {
+    Gauge g = Gauge.build().name("a").unit("seconds").help("h").create();
+    assertEquals("a_seconds", g.fullname);
+
+    Gauge g2 = Gauge.build().name("a_seconds").unit("seconds").help("h").create();
+    assertEquals("a_seconds", g2.fullname);
```
robskillington, line 154: Worth adding a test to observe the unit being added if the name doesn't correctly have the unit on it already when built?

simpleclient_caffeine/src/test/java/io/prometheus/client/cache/caffeine/CacheMetricsCollectorTest.java
```diff
@@ -70,6 +70,7 @@ public void loadingCacheExposesMetricsForLoadsAndExceptions() throws Exception {
         }
         cache.get("user3");
 
+
```
robskillington, line 73: nit: Probably could be removed this line change?

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -2,37 +2,108 @@
...
+  /**
+   * Return the content type that should be used for a given Accept HTTP header.
+   */
+  public static String chooseContentType(String acceptHeader) {
+    if (acceptHeader == null) {
+      return CONTENT_TYPE_004;
+    }
+
+    for (String accepts : acceptHeader.split(",")) {
+      if ("application/openmetrics-text".equals(accepts.split(";")[0].trim())) {
```
robskillington, line 34: nit: Worth making this a const? It's used both in the content-type + version string const as well as here when detecting the accept type.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -2,37 +2,108 @@
...
      * for the output format specification. */
     while(mfs.hasMoreElements()) {
       Collector.MetricFamilySamples metricFamilySamples = mfs.nextElement();
+      String name = metricFamilySamples.name;
       writer.write("# HELP ");
-      writer.write(metricFamilySamples.name);
+      writer.write(name);
+      if (metricFamilySamples.type == Collector.Type.COUNTER) {
+        writer.write("_total");
+      }
```
robskillington, line 71: Hm is Prometheus text version output also meant to have the suffixes appended? I thought only OpenMetrics format would get that?

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -53,6 +124,10 @@ public static void write004(Writer writer, Enumeration<Collector.MetricFamilySam
         writer.write('\n');
       }
     }
+    // Write out any OM-specific samples.
+    if (!omFamilies.isEmpty()) {
+      write004(writer, Collections.enumeration(omFamilies.values()));
```
robskillington, line 129: Hm, could this not be recursive since this is in `write004` method itself? i.e. could it not each time collect into `omFamilies` and recurse but not actually writing it in the child recursive call but simply recursing again infinitely?

Haven't looked to see if there's a test or not for this just yet.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -100,8 +175,101 @@ private static String typeString(Collector.Type t) {
         return "summary";
       case HISTOGRAM:
         return "histogram";
+      case GAUGE_HISTOGRAM:
+        return "histogram";
+      case STATE_SET:
+        return "gauge";
+      case INFO:
+        return "gauge";
```
robskillington, line 183: Should this be info?

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -100,8 +175,101 @@ private static String typeString(Collector.Type t) {
         return "summary";
       case HISTOGRAM:
         return "histogram";
+      case GAUGE_HISTOGRAM:
+        return "histogram";
```
robskillington, line 179: Should this be gauge_histogram? (not sure)

robskillington, line 179: Hm, isn't this written out as a gauge? As per detecting if gsum/gcount is present?

simpleclient/src/main/java/io/prometheus/client/Collector.java
brian-brazil: I don't like indirection like that, it only makes the code harder to understand for me as I have to bounce around the file for something that could be understood one its own line.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
brian-brazil: No, gauges don't have a created.

simpleclient/src/main/java/io/prometheus/client/Counter.java
brian-brazil: Java's current time function returns milliseconds, so that's the effective standard inside Java - and thus also what the client uses everywhere already.

simpleclient/src/test/java/io/prometheus/client/SimpleCollectorTest.java
brian-brazil: That's what the test right above this is doing.

simpleclient_caffeine/src/test/java/io/prometheus/client/cache/caffeine/CacheMetricsCollectorTest.java
brian-brazil: Done

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
brian-brazil: I don't think it'd aid readability to have a constant based on other constants, this is a well known string - I find it easier to have the raw text rather than digging through layers of indirection to see what this code is actually doing.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
brian-brazil: Yes, no matter which format is negotiated the exact same samples should end up inside Prometheus.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
brian-brazil: The `if` is there for a reason. Java lacks inner functions, so this is the easy way to not repeat the rendering logic.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
brian-brazil: Histogram is the closest structurally, even if it's technically a gauge. The Python client does the same.

beorn7:
Since the Prometheus server cannot be configured to _not_ negotiate OpenMetrics, is it perhaps worth it to make OpenMetrics opt-in here, at least for a transition period? I assume this instrumentation library has loads of users, and many will naively update and then be taken by surprise by the changes in naming and the potential of name collisions.

The opt-in hask worked quite well in client_golang. People still run into surprising (for them) issues, but at least they have very explicitly activated OpenMetrics and can easily switch it off again if needed.

brian-brazil:
This PR is at its core a fundamental data model change, the addition of additional output formats and new metric types are incidental to that. No matter what format is negotiated, Prometheus will ultimately ingest the same samples. Having which internal data model to use would be non-trivial, and thus it does not make sense to make this configurable. I'm taking the same approach here as with client_python, clearing calling out the change in the release notes and announcements which is fine as this library is not 1.0. If users wish to continue to use the old code, they can not upgrade.

client_golang still uses the Prometheus data model internally, so it's not a good comparison.

beorn7:
My argument is purely from the users' perspective. They don't care about the internal data model.

I'm not disputing that we are technically entitled to any breaking changes in a pre-1.0 release. However, I think we should still try to be nice to our users if we can. And even if we are technically not breaking any promises, many users will still see it as "our fault" if things break unexpectedly (from their perspective).

The Python client already caused some trouble with the many added `_created` timeseries (which could only be prevented by not upgrading the Python client). That's however, just a load problem. Name collisions and suddenly enforced `_total` suffixes on counters are way more problematic. I'd expect client_java to run more commonly in large, complex, and highly critical production environments. I'm afraid releasing this as is under the current circumstances could hit the trust in the project quite a bit.

If at least Prometheus could have a configurable priority for content negotiation, we could at least offer a quick mitigation strategy.

RichiH:
Would cutting a 0.10 (or 0.9.1) before merging this PR be a compromise?

beorn7:
It's technically a compromise, but it doesn't address the problem.

Users will suffer the same fate with 0.11. Please also take into account that instrumentation happens in code that is not your own, where you are not in control of which version to use. (See the recent Caddy issues as an example from the Go world. At least, Caddy could fix it by adding a configuration option for their users. And zero blame was put on the Prometheus project.)

I think the following migration path would be the least bad:
- 0.10 has OM as opt-in, so users can test it in their environments without any risk of activating OM without even knowing it.
- 0.11 has OM as opt-out, so whoever didn't test it and didn't get the message will now learn it the hard way, but can easily switch off.
- 0.12 will always negotiate OM if possible.

brian-brazil:
> I think the following migration path would be the least bad:

The problem with that suggested path is that it requires making a data model configurable, which is not a reasonable request as that's such a fundamental aspect of the code and would be a major amount of work. This PR was tricky enough as-is without having to duplicate all the relevant code and make it all cleanly and correctly interoperate somehow.

> has OM as opt-in, ... has OM as opt-out, ... negotiate OM if possible.

Once again this is primarily a data model change, which metric format(s) can be negotiated is irrelevant as we'll always end up with the exact same samples being ingested into Prometheus. The way you can opt in or opt out is by choosing the version of client_java to use. If you don't like the change, it's easy to downgrade again.

> Users will suffer the same fate with 0.11. Please also take into account that instrumentation happens in code that is not your own, where you are not in control of which version to use. 

That's not a problem, as that's not how Java dependency resolution works. Whoever builds the final "binary" gets to choose which version is used, not the individual library developers. An end-user can also readily produce such a "binary" with a different version. You can't somehow end up with two different versions of the relevant code inside the binary claiming to both be the simpleclient. There's also no backwards incompatible ABI changes here (I'm always careful with that), so users can roll forwards/backwards as long as they don't write code that uses any of the new features.

Ultimately this is a change which has been carefully considered and publicly communicated for well over a year, including how users can migrate in this grace period. This is the PR that ends that ends the grace period Java, so if users choose to upgrade then they get the OM data model and new metric names with that. If they want to stay as-is, they're free to do that too.

juliusv:
Without having reviewed the code or knowing Java, I'd tend to agree with @brian-brazil here if it's true that the binary author is in control of all dependency versions in Java, and if it's indeed a lot of extra effort to make the behavior configurable within the same version. But again, this is just my 2c without having taken a deep look at things.

roidelapluie:
Thanks for this pull request.


By looking at the code, it seems that we still support prometheus text format. So it seems reasonable to add a feature to not check for _total, _unit and keep the original provided name. And that if you opt-in for OpenMetrics, we serve it when negociated and run the checks.

It also seems that the pre-om functions are not marked as @deprecated, which is something that I think we should do.

brian-brazil:
> So it seems reasonable to add a feature to not check for _total, _unit and keep the original provided name. And that if you opt-in for OpenMetrics, we serve OM when negociated and run the checks about unit and total.

As I've said above, the primary goal of this PR is to change the internal data model of this client to be fully aligned with OM. Accordingly this is not possible, as `Counter` objects now always produce a `_total` for those samples as output no matter what exposition format happens to be subsequently negotiated. This is valid (and indeed recommended) Prometheus text exposition format. It would also be unwise that different data be ingested into Prometheus depending on what is negotiated, as that would be a pain for debugging and understandability - and likely cause even more breakage as we'd be talking blindly stripping _total even where it was desired.

Unit only exists within this client as of this PR, so I'm not sure what you're getting at there.

> It also seems that the pre-om functions are not marked as @deprecated, which is something that I think we should do.

No functions are being deprecated in this PR, nor are there plans to remove any of the functions touched here as they're all still needed. In some far distant future the Prometheus exposition format code may be removed, but that is many years away at best.

However separately as the next release will already contain breaking changes, there are some functions which I deprecated 6 years ago that I'm planning on finally removing.

roidelapluie:
Would it be reasonable, timewise, to duplicate the code in 2 classes, promclient and omclient, with client becoming a set of interfaces who would call omclient or promclient, depending on configuration ?

roidelapluie:
As I see this it is a breaking change for users, in a language that is more popular than golang. I would be extra cautious with our userbase. I would really like to have a migration path.

brian-brazil:
No, that wouldn't be reasonable and also isn't possible as changing a class to an interface is not ABI compatible so would cause pretty serious breakage.
Even were still desired, it would be a massive amount of work, and would likely require coordinated effort on the part of every single user of the current instrumentation libraries to deal with the breaking changes it'd involve plus we're talking having effectively two or three mutually incompatible client libraries that would both need translation layers between them - just to have to undo all that work again in immediate future.

The desired effect is already possible and allowed for by choosing which version of the simpleclient jars to include, which is very easy to do in the Java ecosystem.

>  I would really like to have a migration path.

There is and always has been a migration path, there's no forced upgrade here. As was communicated well over a year ago and will be communicated again upon release if a user wishes to go more gradually they can change non-conforming metric names at their own pace before upgrading. If they don't want to upgrade, they don't have to and their existing code will continue to work. If they upgrade and don't like it, they can downgrade again.

roidelapluie:
Thank you. I agree that the downgrade will be easy for most users. I think that one year is reasonable for users to upgrade and that we could move forward with this.

I do have one last question , would you consider releasing this as a release candidate first? like 0.10.0-RC1? I think that it would help a lot of developers.

roidelapluie:
Also, I expect most users to use the libraries provided in this repository or the jmx_exporter. For which the impact seems to be minimum.

roidelapluie (approved):
Pending Rob's approval for the java part. It seems that this is reasonable. I think releasing a RC first would still be nice.

brian-brazil:
> I do have one last question , would you consider releasing this as a release candidate first? like 0.10.0-RC1? I think that it would help a lot of developers.

I plan on releasing pretty much immediately once this and one more PR (the one to remove the old deprecated functions) is in, so I don't think that'd make any difference. If someone really wants I can push a snapshot release.

> Also, I expect most users to use the libraries provided in this repository or the jmx_exporter. For which the impact seems to be minimum.

Yeah, for the jmx exporter it'd only apply where a user has explicitly configured something to be a counter (untyped is the default) but not added the _total - and it's easy to remove that explicit type from your config. This will have no impact on the cloudwatch_exporter, as that's all gauges.

robskillington (approved):
LGTM

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
robskillington: Ah yes, I see - and these will be gauges in the recurse function. Makes sense.

roidelapluie:
I have computed some numbers. We have 339 importers, each one having even more importers. This seems like there is no way for us to know the impact upfront, except that it will be huge. However, in the importers, we see that only 72 are on the latest release: https://mvnrepository.com/artifact/io.prometheus/simpleclient

RichiH:
That only covers open source and Java will be used in closed source the
most. We could file issues with the open source projects, though.

Brian, is there a guesstimate of how many hours of work it would be to
fully address Beorn's concerns?

Sent by mobile; please excuse my brevity.

brian-brazil:
> The safe minimal move would be to release this as 1.0.0-RC1 or whatever is the proper mechanism in Java for releases, and release the final version in a few weeks.

I don't think it's reasonable to request a permanent change to the versioning strategy of this library and adding burden to all future maintenance for the sake of one particular PR. Version management is far more complicated than you'd naively expect for this repo, as it contains more than a client library, and thus there's no plans to ever go 1.0.

> Brian, is there a guesstimate of how many hours of work it would be to fully address Beorn's concerns?

As far as I can see, Beorn's concerns have already been fully addressed. In particular Beorn's concern seems to be "that instrumentation happens in code that is not your own, where you are not in control of which version to use." - however in the Java ecosystem the end user does in fact control the version used so this is not a issue. This is not Go - users can upgrade/downgrade a library without changing any code, or even recompiling anything.

Also, if a user wants to there's nothing stopping them from continuing to use just write004 (the Prometheus text format exposition function) in their own HTTP serving code. At the exposition function call level you have to change your code to be able to handle the negotiation and exposition for OM exposition, as given the library design that's not something that could happen automatically. The pushgateway integration still uses write004 directly for example.

I would once again like to remind everyone that this PR is at its core data model change. Adding an additional exposition format is orthogonal, even if the OM exposition was removed/opt-out/opt-in all counter values would be exposed with _total in the Prometheus text format.

> This seems like there is no way for us to know the impact upfront, except that it will be huge.

I think you're making some interesting assumptions there, and I don't think that'll be the case based on what we do know.

Doing some digging with Github search, usages of client_java's Counter is only half as common as client_python's. In 2+ years there were only two user reported issues from client_python taking this approach - neither of whom had apparently read the changelog, so opt-in/opt-out (if you could even reasonably do that for a data model change) wouldn't have made the slightest difference. Accordingly I am not worried in terms of impact.

> However, in the importers, we see that only 72 are on the latest release

I'm not surprised that users upgrade slowly. I'd expect that the main reason that users would upgrade is if they specifically wanted this change so I wouldn't expect much in the way of surprise - even if they didn't read the changelog.

RichiH:
I just spent an hour talking to people deep in the corporate Java world, mostly in highly regulated businesses like finance and insurance. While they are not using Prometheus, none of them were concerned with the upgrade Brian proposes.

brian-brazil:
I'll also note that in order to make the path that Beorn suggests actually useful falls into the category of hard research problems as it's not a simple opt-in/opt-out thing that can be done in a few lines of code, the sort of problem that I'd typically leave to my subconcious to tackle for a few months. Just figuring out how to design such a thing in a way that makes sense and doesn't break anyone could easily take weeks, before even starting on implementation.

beorn7:
It was brought to my attention that some people involved in this PR are assuming it's my intention to block merging this. Let me clarify:

I raised my concerns here. They weren't meant recommendations and additional arguments in the discussion. My arguments were discussed, and apparently, everyone disagrees with me. I am not convinced, though. I think my arguments have not been understood sufficiently. And I think many counter-arguments are flawed. However, I have no time to further engage in this discussion and explain my points. Neither have I time to amend this PR myself. Under those circumstances, I can hardly insist that you all follow my bidding. All I ever wanted is to voice my concerns up to a level of detail that I can afford timewise. I that's not enough, too bad…

If you believe you have enough confidence to disregard my concerns, please go ahead. I still believe, however, that this will throw a (in relative terms small, but in absolute terms significant) number of users under the bus in a way that we could have managed better, in particular making it easier for them to stay in control. It will be hard to explain to them why we haven't managed it differently, and they will be even less forgiving because the benefits of these changes are hardly noticeable for them. (And even here I have to partially agree with them. We are essentially burning our "forgiveness budget" for a change that leaves the actual problems of the current text format unaddressed (it in fact makes some of them worse), making future changes even harder to sell.)

juliusv:
@beorn7 Totally appreciate the argument about limited time. OTOH from my POV reading this as someone who doesn't know much about either client_java or Java in general, it seemed like @brian-brazil brought up some reasonable counter points, especially about versioning being in control of the binary author. If there are important caveats to that that I'm not aware of, it would be good to know. Still not sure if that would ultimately change the outcome if it's simultaneously really hard[tm] to make the library support two different data models at the same time.

brian-brazil:
Thanks for the discussion everyone, rollouts like these are always a bit tricky but all things considered I think this is the best path.

simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
robskillington: If so, should we change the proto then? HistogramValue can be used with GAUGE_HISTOGRAM:
https://github.com/OpenObservability/OpenMetrics/blob/master/proto/openmetrics_data_model.proto#L136-L138

I suppose implementors could enforce that's not combined together?

brian-brazil: We should clarify that then.

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -209,6 +215,13 @@ public B namespace(String namespace) {
       this.namespace = namespace;
       return (B)this;
     }
+    /**
+     * Set the unit of the metric. Required.
```
shakuzen, line 219: Isn't a unit optional, not required? From the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/2474ddafe93217cc4979de0b3148e47a1d9340ad/specification/OpenMetrics.md#unit):

> If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore.

brian-brazil: Indeed this is a copy&paste error, would you like to send a PR to fix this?

shakuzen: Sure. I've opened https://github.com/prometheus/client_java/pull/619

justinabrahms:
Just hopping in to mention that this broke a few services in my company due to the backwards incompatible change. :(

nobletrout:
yup this broke stuff.

deepak-auto:
We are blocked from upgrading to a newer version of Spring Boot because of this 🤷 Now our backlog involves fixing dozens, if not hundreds of alerts and dashboards.

bwplotka:
Ack, thanks for the reports. We are aware of this and we are discussing the way to unblock you and the way forward with OM transitions.

fwbrasil:
thank you @bwplotka! Please let me know if we can help. Upgrading the client would solve an important performance issue for us (https://github.com/prometheus/client_java/pull/777)

fwbrasil:
hi folks! any updates on this?

PR #616, opened 2021-01-20
Bridge refactor

joannepolsky:


brian-brazil (commented):
Thanks for your PR. As I said privately this repo is not accepting new modules, so if you want to have something like this your should maintain it in your own repo.

In addition this PR takes 100 lines of simple code that lives in one file, and changes it to 400+ lines across 8 files with several new abstractions. This makes it much harder to maintain, and it would no longer serve its primary purpose of serving as an example - while also making it harder to find and understand the core exposition logic that all users depend on.

I've also never heard of this discourse prometheus_exporter, and from a quick look it is not an exporter. It appears to be primarily a custom client library for Ruby with no connection or compatibility with the official one, but also acts as a monitoring system in that it mixes exporters with direct instrumentation and is doing some level of service discovery. The architecture suggested by this PR also doesn't make sense to me, if something is already using client_java you can have Prometheus scrape it directly rather than using push only to convert it back to pull.
Accordingly this is not something that that should be defacto endorsed in an official client library. I would recommend that if there are features lacking in client_ruby that you work with the maintainers of that library, rather than creating a balkanised situation.

PR #618, opened 2021-01-27
Use JavaDoc since tag on new public API

shakuzen:
This will help other libraries that depend on this library to identify when they are using new API that will raise the minimum version of this library required by end users.


brian-brazil:
Thanks!

PR #619, opened 2021-01-27
Document unit as optional

shakuzen:
It was incorrectly documented as required when initially introduced.
See https://github.com/prometheus/client_java/pull/615#discussion_r565027894


brian-brazil:
Thanks!

PR #620, opened 2021-01-27
Clarify Info javadoc

snago:
If you follow the example in the javadoc you get a metric named `your_build_info_info`.

Also remove unused imports.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Info.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Info.java
@@ -1,30 +1,24 @@
 package io.prometheus.client;
 
-import io.prometheus.client.CKMSQuantiles.Quantile;
-
-import java.io.Closeable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Info metric, key-value pairs.
  *
- * Examples of Info include, build information, version information, and potential target metadata,
- * The first provided state will be the default.
+ * Examples of Info include build information, version information, and potential target metadata.
+ * The first provided state will be the default. The string "_info" will be appended to the name.
  *
  * <p>
- * Example enumeration:
+ * Example Info:
  * <pre>
  * {@code
  *   class YourClass {
  *     static final Info buildInfo = Info.build()
- *         .name("your_build_info").help("Build information.")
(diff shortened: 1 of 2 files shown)


snago:
I didn't understand what `The first provided state will be the default.` was about, but now I'm thinking that it is a copy-paste error from Enumeration. Should I remove that line from the Info javadoc?

simpleclient/src/main/java/io/prometheus/client/Info.java
```diff
@@ -1,30 +1,24 @@
...
  * Info metric, key-value pairs.
  *
- * Examples of Info include, build information, version information, and potential target metadata,
- * The first provided state will be the default.
+ * Examples of Info include build information, version information, and potential target metadata.
+ * The first provided state will be the default. The string "_info" will be appended to the name.
  *
  * <p>
- * Example enumeration:
+ * Example Info:
```
brian-brazil, line 16: Can you fix the capitalisation over in Enumeration too?

simpleclient/src/main/java/io/prometheus/client/Info.java
```diff
@@ -1,30 +1,24 @@
...
-import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Info metric, key-value pairs.
  *
- * Examples of Info include, build information, version information, and potential target metadata,
- * The first provided state will be the default.
+ * Examples of Info include build information, version information, and potential target metadata.
+ * The first provided state will be the default. The string "_info" will be appended to the name.
```
brian-brazil, line 13: This is misleading, the metric name is not changed. The sample name is what will get the suffix.

snago: Sorry, I don't have a complete grasp of all the OM concepts yet. Should it say `... appended to the sample name.` instead?

simpleclient/src/main/java/io/prometheus/client/Info.java
snago: Sure thing!

simpleclient/src/main/java/io/prometheus/client/Info.java
brian-brazil: The term "metric" has a few different meanings in Prometheus, OM tries to fix that by defining it. That should do the trick.

snago: Changed!

Also removed `The first provided state will be the default.` as I believe that was a copy-paste error from Enumeration.

brian-brazil:
Thanks!

Issue #627, opened 2021-02-17
TextFormat: Provide an easy way of mutating the values written


bjung82:
Hi,

a quite common use case is (should be!) testing the whole alertchain:
we provide Prometheus-metrics and use prometheus + alertmanager for
building alertchains. It is really hard to test these chains (initially). All depends
on the value in the exported metric, finally. By providing a simple interface that allows
mutating the value written, we could really improve the situation of testing the whole
alertchain, end-to-end.

I will provide a PR for this.

Regards
Benjamin

fstab:
Thanks for the PR. Let's continue the discussion there. #628

PR #628, opened 2021-02-17
Add value filter to TextFormat for easier supplying fake values.

bjung82:
Fake values are useful for easy simulation of (temporary) alert situations. By proving a mutating ValueFilter in software that uses the simpleclient, end-to-end tests of the whole alerting chain will be much easier.

See #627

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -61,9 +61,34 @@ public class TextFormat {
   }
 
   /**
+   * Filter can be used to change the value of a sample, e.g. for simulating specific values for
+   * testing alerts.
+   */
+  public interface ValueFilter {
+    double getValue(Collector.MetricFamilySamples.Sample sample);
+  }
+
+  /**
+   * Noop filter that simple returns the original value
+   */
+  public static class NullValueFilter implements ValueFilter {
+    @Override
+    public double getValue(Collector.MetricFamilySamples.Sample sample) {
+      return sample.value;
+    }
+  }
+
+  /**
    * Write out the text version 0.0.4 of the given MetricFamilySamples.
    */
   public static void write004(Writer writer, Enumeration<Collector.MetricFamilySamples> mfs) throws IOException {
+    write004(writer, mfs, new NullValueFilter());
+  }
+
+  /**
+   * Write out the text version 0.0.4 of the given MetricFamilySamples.
+   */
+  public static void write004(Writer writer, Enumeration<Collector.MetricFamilySamples> mfs, ValueFilter valueFilter) throws IOException {
     Map<String, Collector.MetricFamilySamples> omFamilies = new TreeMap<String, Collector.MetricFamilySamples>();
     /* See http://prometheus.io/docs/instrumenting/exposition_formats/
      * for the output format specification. */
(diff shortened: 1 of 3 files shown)


fstab:
Hey @bjung82, thanks for the PR. I agree having an easy way for manipulating metrics for end-to-end tests is a good idea. However, I'm not sure if this is the best way to do it, because many users don't call `write004()` directly but through an existing exporter like the `MetricsServlet` or through a framework like [micrometer](https://micrometer.io/), and these users would find it hard to use that feature. I don't know if there is a way that works for everyone, but it would be worthwhile to consider a few alternatives before we do this. Let's take some time to think of alternatives and then get back to this and decide how to proceed.

bjung82:
Yes, we use Micrometer.. I created an issue months ago, but they are not that interested. It would be a good feature for other metric backends, too. But coordinating this ist out of scope/interest of the Micrometer Project, I assume.

If we could add this feature, I could come again and try to convince them to integrate it. Or override a public method in Micrometer..

https://github.com/micrometer-metrics/micrometer/issues/2243

bjung82:
@fstab  What is your opinion on this topic? Introducing this feature could help a lot, even if micrometer would not integrate it directly. And if it would be already integrated I would have chances to change the micrometers guy opinion. Parsing and fiddling around with the already generated prometheus output would be a pain. It would be unnecessary and inefficient and not future proof, too. And what about the new OpenMetrics-Format - I would have to change my (hopfully never written) hacky parsing code again to adapt for any future changes in the format, too. 

IMHO Providing the (optional) ValueFilter would be a clean and future proof way of achiving end-to-end tests, at least for prometheus. Which should be our main concern now ;-)

fstab:
Hi Benjamin, sorry for not getting back to this earlier. My concern is that this PR is made with a very specific test scenario in mind, and is not a generic end-to-end testing solution. You have to call `write004()` directly, so it doesn't work with standard exporters. The `ValueFilter` operates on a very low level single value, which makes it hard to implement higher level testing (e.g. shift a latency histogram by 1s).

What do you think about doing something like this to achieve your goal (didn't run the code):

```java
  public static Enumeration<Collector.MetricFamilySamples> mock(final Enumeration<Collector.MetricFamilySamples> delegate) {

    return new Enumeration<Collector.MetricFamilySamples>() {
      @Override
      public boolean hasMoreElements() {
        return delegate.hasMoreElements();
      }

      @Override
      public Collector.MetricFamilySamples nextElement() {
        Collector.MetricFamilySamples orig = delegate.nextElement();
        return new Collector.MetricFamilySamples(orig.name, orig.unit, orig.type, orig.help,
            orig.samples.stream().map(sample -> new Collector.MetricFamilySamples.Sample(
                sample.name,
                sample.labelNames,
                sample.labelValues,
                modifiedValue(sample)
            )).collect(Collectors.toList()));
      }
    };
  }

  public static double modifiedValue(Collector.MetricFamilySamples.Sample sample) {
    return sample.value * 3;
  }
```
With that you could call
```java
write004(writer, mock(mfs));
```
to have all values multiplied by 3.

If you use these 20 lines of code in your application, you can achieve the same goal but we don't need to extend the `TextFormat` interface with a very specific `ValueFilter` implementation.

bjung82:
Hi Fabian,

this looks like a good solution - I will give it a try! With the value filter, we had to 
build more complex logic (like recognizing metrics of the same family by metrics name etc.).

Changing/extending micrometer has to be done either way. I can symply subclass
io.micrometer.prometheus.PrometheusMeterRegistry and provide a more flexible
scrape method, taking a mock like you suggested.

If you want you can close this isssue. Thank you very much :-)

PR #630, opened 2021-03-02
Updated Hibernate Documentation

ablx:
Added a code snippet which shows how to get the session factory from an EntityManager instead of an EntityManagerFactory.


fstab (changes requested):
Thanks Micro. Please add a semicolon at the end of the line :)

PR #639, opened 2021-03-15
Add parameter to optionally strip deploy path.

### LINKED ISSUE #631: Avoid deploy path in the MetricsFilter
Right now if I have a WAR using `MetricsFilter` on `/*` and deploy that under `/deployPoint` I get metrics content like:

```
projectName_requests_count{path="/deployPoint/api/listUser",method="GET",} 2.0
projectName_requests_sum{path="/deployPoint/api/listUser",method="GET",} 0.228438531
```

and if on a different server (or even the same) I have the same WAR deployed under `/anotherDeploy` I get metrics that are difficult to `SUM()` together, as they have a different `path` label, while being logically "the same metric".

I'd suggest changing like this:
```patch
-        String path = request.getRequestURI();
+        String path = request.getRequestURI().substring(request.getContextPath().length());
```

(also as suggested by @jfbreault in https://github.com/prometheus/client_java/pull/168#discussion_r91014868)

lapo-luchini:
Closes #631.

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
@@ -60,6 +60,7 @@ public class MetricsFilter implements Filter {
     static final String HELP_PARAM = "help";
     static final String METRIC_NAME_PARAM = "metric-name";
     static final String BUCKET_CONFIG_PARAM = "buckets";
+    static final String STRIP_CONTEXT_PATH_PARAM = "strip-context-path";
     static final String UNKNOWN_HTTP_STATUS_CODE = "";
 
     private Histogram histogram = null;
@@ -68,6 +69,7 @@ public class MetricsFilter implements Filter {
     // Package-level for testing purposes.
     int pathComponents = 1;
     private String metricName = null;
+    private boolean stripContextPath = false;
     private String help = "The time taken fulfilling servlet requests";
     private double[] buckets = null;
 
@@ -145,6 +147,10 @@ public class MetricsFilter implements Filter {
                     buckets[i] = Double.parseDouble(bucketParams[i]);
                 }
             }
+
+            if (!isEmpty(filterConfig.getInitParameter(STRIP_CONTEXT_PATH_PARAM))) {
+                stripContextPath = Boolean.parseBoolean(filterConfig.getInitParameter(STRIP_CONTEXT_PATH_PARAM));
+            }
         }
 
         if (buckets != null) {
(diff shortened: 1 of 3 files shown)


simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -171,6 +177,8 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
         HttpServletRequest request = (HttpServletRequest) servletRequest;
 
         String path = request.getRequestURI();
+        if (stripContextPath)
```
fstab, line 180: We have `{` `}` in one line `if` statements everywhere else in the code, so please add `{` `}` here as well.

fstab (changes requested):
Thanks a lot or the PR. Here are a few minor comments:

Please update the class JavaDoc for `MetricsFilter`. Add the new option with its default to the example:
```xml
    <!-- strip-context-path is optional, defaults to false -->
   <init-param>
      <param-name>strip-context-path</param-name>
      <param-value>false</param-value>
   </init-param>
```
...and add an explanation of the `strip-context-path` option to the text above the example.

Additionally, please also update the _Servlet Filter_ section in `README.md`.

There should be a test for the new option in `MetricsFilterTest`. With copy-and-paste from `init()` it could look something like this

```java
    @Test
    public void testStripContextPath() throws Exception {
        String metricName = "foo";
        FilterConfig cfg = mock(FilterConfig.class);
        when(cfg.getInitParameter(MetricsFilter.METRIC_NAME_PARAM)).thenReturn(metricName);
        when(cfg.getInitParameter(MetricsFilter.PATH_COMPONENT_PARAM)).thenReturn("0");
        when(cfg.getInitParameter(MetricsFilter.STRIP_CONTEXT_PATH_PARAM)).thenReturn("true");
        f.init(cfg);
        assertTrue(f.stripContextPath);
        HttpServletRequest req = mock(HttpServletRequest.class);
        when(req.getRequestURI()).thenReturn("/foo/bar/baz/bang");
        when(req.getContextPath()).thenReturn("/foo/bar");
        when(req.getMethod()).thenReturn(HttpMethods.GET);
        HttpServletResponse res = mock(HttpServletResponse.class);
        FilterChain c = mock(FilterChain.class);
        f.doFilter(req, res, c);
        verify(c).doFilter(req, res);
        final Double sampleValue = CollectorRegistry.defaultRegistry.getSampleValue(metricName + "_count", new String[]{"path", "method"}, new String[]{"/baz/bang", HttpMethods.GET});
        assertNotNull(sampleValue);
        assertEquals(1, sampleValue, 0.0001);
    }
```

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
lapo-luchini: Should be fine now.

lapo-luchini:
@fstab let me know if anything more should be changed.

fstab:
Thanks a lot @lapo-luchini, and sorry that it took so long to merge this.

Issue #622, opened 2021-01-27
Add Exemplar Support to Client Java


joe-elliott:
Feature Request:
Addition of exemplar support to the previously merged OM support: https://github.com/prometheus/client_java/pull/615

Thanks for working on this!

fitzoh:
Any thoughts on API design?

For reference, the client_golang implementation was added in https://github.com/prometheus/client_golang/pull/706 :

```golang
//counters
Add(float64)
AddWithExemplar(value float64, exemplar Labels)
//histograms
Observe(float64)
ObserveWithExemplar(value float64, exemplar Labels)
```

I'm inclined to follow the golang pattern and use distinct method names for exemplar method calls to reduce the risk of people confusing exemplar labels and metric labels.

Does this seem reasonable?
```java
addWithExemplar(double value, String... exemplarLabels)
observeWithExemplar(double value, String... exemplarLabels)
```

Or would a key/value wrapper be preferred?
```java
addWithExemplar(value, Label.of("key", "value"))
```
Adding a label class that's only used for exemplars feels a little weird.

fstab:
In OpenTelemetry for Java, the current span context is always available via

```java
Span.current().getSpanContext()
```

So the question is, what should our recommended way of dealing with exemplars in Java be? Explicit API calls where the user passes the trace ID as parameter, or implicit integration with OpenTelemetry, i.e. `client_java` adds exemplars automatically under the hood if a Span context is available?

One argument for the implicit integration is that this will still work if the user does not call the `client_java` API directly, but used this library indirectly via Micrometer. In that case, the user will not be able to call `incWithExemplar()`, but the Java client library will still be able to add examplars by getting them from the current OpenTelemetry span context.

fitzoh:
One idea that had crossed my mind for exemplar recording would be initializing the counter/histogram with an optional `Supplier<Exemplar>`, where the supplier might be a lambda that pulls a value from an MCD/trace context or closes over some over system that it can get metadata from. That gets a little tricker if you need to define how often it gets invoked. Do you get an exemplar every time a value is recorded? Every nth invocation? At a fixed rate?

The other option is to completely punt on that question and keep `client_java` simple with by using something like `incWithExemplar` and pushing the complexity to micrometer/OT/spring cloud sleuth and let them control when a value is recorded with/without an exemplar.

fstab:
[...]

PR #641, opened 2021-03-17
Exemplar API sketch

fitzoh:
Expending a [comment](https://github.com/prometheus/client_java/issues/622#issuecomment-800901953) from https://github.com/prometheus/client_java/issues/622 to a PR.

This PR proposes an API for recording exemplar values on counters and histograms.

Adds new `<inc|add>WithExemplar(double value, String... exemplarLabels)` methods which take string varargs containing exemplar values.

Inspired by initial golang implementation by @beorn7 : https://github.com/prometheus/client_golang/pull/706

### Not addressed
* persisting exemplars/exposing them to scrapes (wanted to start with the validating the API) 
* histogram timer methods (adding exemplars to those could be tricky, doesn't need to block initial implementation)

### Intentionally omitted
* `Counter#incWithExemplar(String... examplarLabels)` increment by 1 convenience method (less duplication in the API, forcing the user to include an amount felt reasonable if they're already using the more verbose method).

@fstab @tomwilkie

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/Counter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/Counter.java
@@ -143,10 +143,11 @@ public class Counter extends SimpleCollector<Counter.Child> implements Collector
      * @param amt the amount to increment the counter by
      * @param exemplarLabels a set of key value pairs used to create the exemplar
      *
-     * @throws IllegalArgumentException if exemplarLabels is null
-     * @throws IllegalArgumentException if exemplarLabels does not contain an even number of values
-     * @throws IllegalArgumentException if exemplarLabels contains any invalid labels
-     * @throws IllegalArgumentException if the combined length of the exemplarLabels is greater than 64
+     * @throws IllegalArgumentException If amt is negative.
+     * @throws IllegalArgumentException if exemplarLabels is null.
+     * @throws IllegalArgumentException if exemplarLabels does not contain an even number of values.
+     * @throws IllegalArgumentException if exemplarLabels contains any invalid labels.
+     * @throws IllegalArgumentException if the combined length of the exemplarLabels is greater than 64.
      */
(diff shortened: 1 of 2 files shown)


simpleclient/src/main/java/io/prometheus/client/Counter.java
```diff
@@ -135,6 +137,24 @@ public void inc(double amt) {
...
+     *
+     * @param amt the amount to increment the counter by
+     * @param exemplarLabels a set of key value pairs used to create the exemplar
+     *
+     * @throws IllegalArgumentException If amt is negative.
+     * @throws IllegalArgumentException if exemplarLabels is null.
+     * @throws IllegalArgumentException if exemplarLabels does not contain an even number of values.
+     * @throws IllegalArgumentException if exemplarLabels contains any invalid labels.
+     * @throws IllegalArgumentException if the combined length of the exemplarLabels is greater than 64.
+     */
```
fitzoh, line 151: Initial golang documentation that I can probably borrow from a little more heavily:
https://github.com/prometheus/client_golang/blob/c32ffd121f40843d8ce747536467bb091678fd87/prometheus/counter.go#L44-L51

There were two things I wanted to point out here.

Do we want to copy the go behavior of ignoring null labels?
I'm inclined to say that null labels are invalid and that the user should use `inc` instead.

Exceptions for invalid labels scare me a bit as well. Having instrumentation code throw an exception  because your dynamically generated label ended up a little too long sounds like a not fun user experience, but I don't have a better alternative other than logging an error or incrementing an internal error counter.

fstab:
Thanks a lot for picking this up again. I continued the discussion in #622.

fstab:
I just merged Exemplar support to the `next-release` branch, and it contains API very close to what you proposed here. Thanks for your proposal.

PR #643, opened 2021-03-17
Set HTTPServer request/response timeouts

### LINKED ISSUE #636: Support request and response timeouts in HTTPServer
### Problem description

The `com.sun.net.httpserver.HttpServer` as used by `io.prometheus.client.exporter.HTTPServer` does not, by default, use any request and response timeouts at all. This means that an HTTP server set up with a limited amount of threads (say, 1 or 5) can easily be fully blocked by a malicious/misconfigured client (or a network failure), indefinitely.

We have seen this ourselves with a custom Prometheus publisher in our code (which also uses `com.sun.net.httpserver.HttpServer`), and the slightly naïve approach we took was attempting to override the default timeouts in our code using a static initializer:

```java
    static {
        System.setProperty( "sun.net.httpserver.maxReqTime", "60" );
        System.setProperty( "sun.net.httpserver.maxRspTime", "600" );
    }
```

_However_, because we also use [the Prometheus JMX exporter](https://github.com/prometheus/jmx_exporter) which in turn uses `simpleclient_httpserver` under the hood, the JMX exporter agent runs before our own code (presumably because of how Java agents work in the JVM) => it becomes impossible to override the request and response timeouts using the approach above. Here's the stack trace for when `sun.net.httpserver.ServerConfig` gets initialized (after which it becomes impossible to override these properties):

image

### Workaround

Set `-Dsun.net.httpserver.maxReqTime=60 -Dsun.net.httpserver.maxRspTime=600` on the command line instead; it will then apply to all use cases of `com.sun.net.httpserver.HttpServer` withinin the Java runmtime in question.

### Suggested changes

I am creating this issue to raise awareness of the problem and discuss potential ways to improve this before spending any time on a potential PR.

Personally, I think it's a bit problematic to rely on the JDK defaults in this case; it would make sense to set up some reasonable defaults in case the user hasn't set these system properties (since I'd claim it's rather unlikely that the user actually wants an "infinite" request/response timeout" like this). 

One important question then  is whether this should be done in `io.prometheus.client.exporter.HTTPServer` or taken care of by the JMX exporter? Feel free to suggest what you think would make most sense. :+1:

perlun:
Closes #636

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -36,6 +36,17 @@ import com.sun.net.httpserver.HttpServer;
  * </pre>
  * */
 public class HTTPServer {
+
+    static {
+        if (!System.getProperties().containsKey("sun.net.httpserver.maxReqTime")) {
+            System.setProperty("sun.net.httpserver.maxReqTime", "60");
+        }
+
+        if (!System.getProperties().containsKey("sun.net.httpserver.maxRspTime")) {
+            System.setProperty("sun.net.httpserver.maxRspTime", "600");
+        }
+    }
+
     private static class LocalByteArray extends ThreadLocal<ByteArrayOutputStream> {
         @Override
         protected ByteArrayOutputStream initialValue()
@@ -256,4 +267,3 @@ public class HTTPServer {
         return server.getAddress().getPort();
     }
 }
-


simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -36,6 +36,17 @@
...
+    static {
+        if (!System.getProperties().containsKey("sun.net.httpserver.maxReqTime")) {
+            System.setProperty("sun.net.httpserver.maxReqTime", "60");
+        }
+
+        if (!System.getProperties().containsKey("sun.net.httpserver.maxRspTime")) {
+            System.setProperty("sun.net.httpserver.maxRspTime", "600");
+        }
+    }
+
```
perlun, line 49: I tried adding a test for this as well, like this:

```diff
diff --git simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
index c9f6bee..f02dbd4 100644
--- simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
+++ simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
@@ -13,6 +13,7 @@ import java.util.zip.GZIPInputStream;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import sun.net.httpserver.ServerImplHelper;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
@@ -147,4 +148,14 @@ public class TestHTTPServer {
     String response = requestWithCompression("/-/healthy", "");
     assertThat(response).contains("Exporter is Healthy");
   }
+
+  @Test
+  public void testRequestTimeout() {
+    assertThat(ServerImplHelper.getRequestTimeout()).isEqualTo(60000);
+  }
+
+  @Test
+  public void testResponseTimeout() {
+    assertThat(ServerImplHelper.getResponseTimeout()).isEqualTo(600000);
+  }
 }
diff --git simpleclient_httpserver/src/test/java/sun/net/httpserver/ServerImplHelper.java simpleclient_httpserver/src/test/java/sun/net/httpserver/ServerImplHelper.java
new file mode 100644
index 0000000..f841435
--- /dev/null
+++ simpleclient_httpserver/src/test/java/sun/net/httpserver/ServerImplHelper.java
@@ -0,0 +1,17 @@
+package sun.net.httpserver;
+
+/**
+ * Helper class to be able to read package-internal fields from {@link ServerImpl}.
+ *
+ * @author Per Lundberg
+ */
+public class ServerImplHelper {
+
+    public static long getRequestTimeout() {
+        return ServerImpl.MAX_REQ_TIME;
+    }
+
+    public static long getResponseTimeout() {
+        return ServerImpl.MAX_RSP_TIME;
+    }
+}
```

However, this fails with an exception: `java.lang.IllegalAccessError: tried to access class sun.net.httpserver.ServerImpl from class sun.net.httpserver.ServerImplHelper`. The `sun.net.httpserver` classes are really not written with (external, non-JDK) testability in mind... :unamused:

fstab (approved):
Thanks a lot! I agree that this is hard to test.

PR #647, opened 2021-03-30
Add support to jakarta servlet

mmadoo:
The module simpleclient_servlet_jakarta is a copy of simpleclient_servlet where javax.servlet is replaced by jakarta.servlet (real changes can be seen in the second commit)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,13 @@
 .idea
 *.iml
 
+### STS ###
+.apt_generated
+.classpath
+.factorypath
+.project
+.settings
+
 target/
 simpleclient_pushgateway/mockserver.log
 simpleclient_pushgateway/mockserver_request.log
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,7 @@
         <module>simpleclient_logback</module>
         <module>simpleclient_pushgateway</module>
         <module>simpleclient_servlet</module>
+        <module>simpleclient_servlet_jakarta</module>
         <module>simpleclient_spring_web</module>
         <module>simpleclient_spring_boot</module>
         <module>simpleclient_jetty</module>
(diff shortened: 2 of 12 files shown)


fstab:
Thanks a lot. Do you see any way to reuse the existing code? For example, if we merge #639 the jakarta servlet will not have that change.

mmadoo:
I currently see no easy way to reuse the existing code.
Jetty have the same issue (the main branch is 10.0.x (using javax.servlet) and the branch 11.0.x (using jakarta.servlet) is updated to take changes from 10.0.x), but I do not know how they deal with it.

Changes from src dir are easy to report as the only changes are on import but for test directory it is not the case.

mmadoo:
I have found a solution:
I add the profile updateSource which copy the src dir simpleclient_servlet to simpleclient_servlet_jakarta and do required replace in order to allow compilation of simpleclient_servlet_jakarta. 
You could test it with: ` mvn compile -P updateSource`.

This solution is not perfect as it may requires some adpations on test evolution and requires to modify first simpleclient_servlet, but it should do the job in a first time.
If this approach fit to you, I could update the doc (just provide me the best file to update)

travisspencer:
Are there plans to merge this PR? Without this kinda change, we cannot upgrade to Jakarta and drop the old javax servlet API.

fstab (commented):
@mmadoo sorry that it took so long to get back to this. Could you rebase it to the current master branch and squash the commits into one?

fstab:
Hey, I just pushed a commit that replaces this one, so I'm closing it. The next release will have Jakarta support. The reason why I pushed my own commit is that I played with re-using code across both implementations (javax and jakarta), and I ended up having a quite complete solution. Thanks a lot for bringing this up, jakarta support is now on its way.

travisspencer:
Did this get released yet, @fstab ?

rossbu:
never

> Did this get released yet, @fstab ?

PR #652, opened 2021-04-12
add Exemplar support for OpenTelemetry tracing

fstab:
# New Section in README:

Exemplars are a feature of the [OpenMetrics](http://openmetrics.io) format that allows applications to link metrics
to example traces. Starting with version 0.11.0, `client_java` automatically adds Exemplars for applications that
are instrumented with [OpenTelemetry](https://opentelemetry.io/) distributed tracing. No code change is required.

### Running an Exemplars Example

If you want to see this in action, you can run the example from the `ExemplarsClientJavaIT`:

```
mvn package
cd integration_tests/exemplars_otel_agent/target/
curl -LO https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v1.0.1/opentelemetry-javaagent-all.jar
java -Dotel.traces.exporter=logging -Dotel.metrics.exporter=none -javaagent:./opentelemetry-javaagent-all.jar -jar ./sample-rest-application.jar
```

Now you have a Spring REST service running on [http://localhost:8080/hello](http://localhost:8080/hello) that is instrumented with the OpenTelemetry Java agent.

In order to get metrics in [OpenMetrics](http://openmetrics.io) format, run

```
curl -H 'Accept: application/openmetrics-text; version=1.0.0; charset=utf-8' http://localhost:8080/metrics
```

You should see metrics with Exemplars, for example in the `request_duration_histogram` metric:

```
request_duration_histogram_bucket{path="/god-of-fire",le="0.004"} 4.0 # {trace_id="043cd631811e373e4180a678c06b128e",span_id="cd122e457d2ca5b0"} 0.0033 1618261159.027
```

Note that this is an example application for a unit test, so durations don't represent real durations, and some example metrics might not make sense in the real world.

### Enabling And Disabling Exemplars

By default, Exemplars are enabled if OpenTelemetry tracing is detected. You can disable this globally with:

```java
ExemplarConfig.disableByDefault();
```

If you want to enable Exemplars only for a single metric, use `withExemplarSampler()` in the metric builder.

```java
// This also works with Gauges, Histograms, and Summaries.
Counter labelsCustomExemplar = Counter.build()
    .name("number_of_events_total")
    .help("help")
    .withExemplarSampler(new MyExemplarSampler())
    ...
    .register();
```

Likewise, you can disable Exemplars for a single metric like this:

```java
// This also works with Gauges, Histograms, and Summaries.
Counter labelsCustomExemplar = Counter.build()
    .name("number_of_events_total")
    .help("help")
    .withoutExemplars()
    ...
    .register();
```

See `ExemplarTest` for more examples.

### Implement Your Own Exemplar Sampling Algorithm

The `DefaultExemplarSampler` is very simple: It samples a new Exemplar if the current one is older than 7 seconds.
You might want to implement your own Exemplar sampler that provides more interesting Exemplars for your application.
In order to do so, just implement the following interfaces:

* `CounterExemplarSampler`
* `GaugeExemplarSampler`
* `HistogramExemplarSampler`
* `SummaryExemplarSampler`

You can set your implementations as default like this:

```java
ExemplarConfig.setDefaultCounterExemplarSampler(mySampler);
ExemplarConfig.setDefaultGaugeExemplarSampler(mySampler);
ExemplarConfig.setDefaultHistogramExemplarSampler(mySampler);
ExemplarConfig.setDefaultSummaryExemplarSampler(mySampler);
```

### Implement Support for Other Tracing Vendors

Version 0.11.0 implements support for OpenTelemetry. If you are a vendor for another distributed tracer,
please create a pull request adding support for your system. The idea is that copy-and-paste the
`tracer_otel` module and modify it for your needs. If this turns out to be too simplistic, please create
a GitHub issue and let us know what you need.

# Implementation Notes

This PR also introduces integration tests with [testcontainers](https://www.testcontainers.org/), which is a framework for running tests against Docker containers. Benefits:
* This allows tests like the `ExemplarsClientJavaIT` where we attach the OpenTelemetry Java agent to the system under test.
* This allows tests with different Java versions, like `JavaVersionsIT` that can be used to ensure the library is compatible with Java 6 or IBM's OpenJ9.

That means `docker` must be installed on the system where you run `mvn verify`.


simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -241,25 +241,37 @@ public static void writeOpenMetrics100(Writer writer, Enumeration<Collector.Metr
...
+          writer.write(' ');
+          writeTimestamp(writer, sample.exemplar.getTimestampMs());
         }
         writer.write('\n');
       }
     }
     writer.write("# EOF\n");
   }
 
+  private static void writeTimestamp(Writer writer, long timestampMs) throws IOException {
```
brian-brazil, line 262: omWriteTimestamp would avoid confusion, as this is OM specific logic.

brian-brazil (commented):
There's a few things here that don't jive with the OM spec.

README.md
```diff
@@ -291,6 +296,99 @@ class YourClass {
...
+See `ExemplarTest` for more examples.
+
+### Implement Your Own Exemplar Sampling Algorithm
+
+The `DefaultExemplarSampler` is very simple: It samples a new Exemplar if the current one is older than 7 seconds.
+You might want to implement your own Exemplar sampler that provides more interesting Exemplars for your application.
+In order to do so, just implement the following interfaces:
+
+* `CounterExemplarSampler`
+* `GaugeExemplarSampler`
```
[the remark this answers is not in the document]
fstab: removed

README.md
```diff
@@ -291,6 +296,99 @@ class YourClass {
...
+* `HistogramExemplarSampler`
+* `SummaryExemplarSampler`
```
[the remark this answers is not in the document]
fstab: removed

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Exemplar.java
```diff
@@ -0,0 +1,70 @@
...
+ * Immutable data class holding an Exemplar.
+ */
+public class Exemplar {
+
+  private final String traceId;
+  private final String spanId;
+  private final double value;
+  private final long timestampMs;
+
+  public Exemplar(String traceId, String spanId, double value, long timestampMs) {
```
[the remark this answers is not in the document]
fstab: fixed

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Value.java
```diff
@@ -0,0 +1,13 @@
+package io.prometheus.client.exemplars.api;
+
+/**
+ * Lazy value holder for a double value.
+ * <p/>
+ * Counters and Gauges track their current value using a DoubleAdder, so getting the value
```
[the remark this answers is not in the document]
fstab: I changed the `CounterExemplarSampler` to this
```java
Exemplar sample(double increment, Value newTotalValue, Exemplar previous);
```
The default implementation uses the `increment` and ignores the `newTotalValue`. However, I still kept the `newTotalValue` available in the API so that users can access it if needed (e.g. start sampling after the total value exceeded a threshold).

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/impl/DefaultExemplarSampler.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  interface Clock {
+    long currentTimeMillis();
+  }
+
+  static class SystemClock implements Clock {
+    @Override
+    public long currentTimeMillis() {
+      return System.currentTimeMillis();
```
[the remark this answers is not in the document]
fstab: I changed the implementation so that this is only called if OpenTelemetry tracing is available.

For the latency, I added a `ExemplarsBenchmark`. Result on my local Linux laptop:
```
Benchmark                                               Mode  Samples   Score   Error  Units
i.p.b.ExemplarsBenchmark.testCounter                    avgt      200  28.252 ± 0.312  ns/op
i.p.b.ExemplarsBenchmark.testCounterWithExemplars       avgt      200  46.042 ± 0.781  ns/op
i.p.b.ExemplarsBenchmark.testCounterWithoutExemplars    avgt      200  29.322 ± 0.475  ns/op
```
so about 17ns for the `currentTimeMillis()` call.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -241,25 +241,37 @@ public static void writeOpenMetrics100(Writer writer, Enumeration<Collector.Metr
...
-            writer.write("0");
-          }
-          if (ms < 10) {
-            writer.write("0");
-          }
-          writer.write(Long.toString(ts % 1000));
-
+          writeTimestamp(writer, sample.timestampMs);
+        }
+        if (sample.exemplar != null) {
```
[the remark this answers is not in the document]
fstab: fixed with the new `Exemplar` data model

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Exemplar.java
```diff
@@ -0,0 +1,70 @@
+package io.prometheus.client.exemplars.api;
+
+/**
+ * Immutable data class holding an Exemplar.
+ */
+public class Exemplar {
```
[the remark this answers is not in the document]
fstab: The `ExemplarSampler` implementations use the `Exemplar` data model, so if we move the model to simpleclient, we will need to move all `ExemplarSampler` implementations to simpleclient as well. Currently it's not much, but if in the future people provide more sophisticated implementations it might be worthwhile to have this in a separate module.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
fstab: changed

fstab:
Thanks a lot @brian-brazil for reviewing this, I appreciate it a lot. I pushed an update and I'd be happy if you could review it once more.

benchmark/pom.xml
```diff
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>io.prometheus</groupId>
             <artifactId>simpleclient</artifactId>
-            <version>0.10.1-SNAPSHOT</version>
+            <version>${project.version}</version>
```
brian-brazil, line 52: This will break some build systems, I forget which one but that's the reason all versions are explicit.

benchmark/src/main/java/io/prometheus/benchmark/ExemplarsBenchmark.java
```diff
@@ -0,0 +1,80 @@
+package io.prometheus.benchmark;
```
brian-brazil, line 1: The readme should be updated with the new numbers.

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/CounterExemplarSampler.java
```diff
@@ -0,0 +1,18 @@
+package io.prometheus.client.exemplars.api;
+
+/**
+ * Exemplar sampler for counter metrics.
+ */
+public interface CounterExemplarSampler {
+
+  /**
+   * @param increment     the value added to the counter on this event
+   * @param newTotalValue the new total counter value after adding the increment.
```
brian-brazil, line 10: Why would you care about this? Increments should be independent.

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Exemplar.java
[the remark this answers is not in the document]
brian-brazil: My point is that the core data model should all live in one place, not spread across N modules that have to be pieced together by someone trying to work with this. It's kinda like having an entire module just to handle timestamps.

Personally I wouldn't have the core simpleclient depending on any other modules at all, rather make this something that the user has to add an extra module to enable, rather than silently creating the possibility for dependency hell on upgrade. simpleclient thus far has purposefully avoided any dependencies for that reason, it should always be safe to have simpleclient instrumentation in place.

This is also still not the data model of an OM exemplar, as timestamps are optional.

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Exemplar.java
```diff
@@ -0,0 +1,106 @@
...
+  private final long timestampMs;
+
+  /**
+   * Create an Exemplar without a timestamp
+   *
+   * @param value  the observed value
+   * @param labels name/value pairs. Expecting an even number of strings. The combined length of the label names and
+   *               values must not exceed 128 UTF-8 characters. Neither a label name nor a label value may be null.
+   */
+  public Exemplar(double value, String... labels) {
```
brian-brazil, line 24: Passing in labels should be consistent with how it's done in Info, so a map should also be supported.

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/ExemplarConfig.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  // Disable all of the above
+
+  public static void disableExemplars() {
+    instance.counterExemplarSampler = instance.noopExemplarSampler;
+    instance.histogramExemplarSampler = instance.noopExemplarSampler;
+  }
+
+  // Methods for internal use
```
brian-brazil, line 59: If they're internal, why are they public?

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Exemplar.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * @return 0 means no timestamp.
+   */
+  public long getTimestampMs() {
+    return timestampMs;
+  }
+
+  private void validateLabels(String... labels) {
```
brian-brazil, line 67: You should also check that the label names are valid.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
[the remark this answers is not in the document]
brian-brazil: Still not quite right, 0 is a valid timestamp.

brian-brazil (commented):
All public methods should have javadoc.

benchmark/pom.xml
fstab: reverted

benchmark/src/main/java/io/prometheus/benchmark/ExemplarsBenchmark.java
fstab: done

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/CounterExemplarSampler.java
fstab: I'm not sure about this. Here's an example that might be a valid use case: A counter representing a buffer size. If the buffer can only grow, a counter should be a valid metric. A user might decide to sample exemplars only when the buffer has exceeded a certain size, thus the exemplar sampler needs the total value.

I'm not saying that this will be common, in most cases users won't need the total value. However, if we remove it from the API there will be no chance to use it even if there are rare cases where it makes sense.

Anyway, I don't have a strong opinion on this. If you still think removing it would be better I'm happy to remove it.

brian-brazil: The interesting thing about a counter is how quickly it goes up, the absolute value doesn't matter.

>  A counter representing a buffer size

That's a gauge, as you care about the absolute value.

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/Exemplar.java
fstab: I moved the `Exemplar` and all classes that depend on it to `simpleclient`. Now only the bridges to the tracing libraries live in separate modules. I agree that it should be safe to have simpleclient instrumentation in place, and that dependencies should be avoided. I hope this is the case with the current implementation: `simpleclient` does not have any runtime dependencies. OpenTelemetry is available we will load the corresponding `SpanContextSupplier`, if not we won't.

We could disable exemplars by default and have the user explicitly enable them with a `ExemplarConfig.enableExemplars()` call. Apart from that I don't see how to improve this. We could load `SpanContextSuppliers` dynamically through a class loader, but that would yield to the same optional dependency.

I added an `Exemplar` constructor with a `Map` parameter to be consistent with `Info`.

exemplars_parent/exemplars/src/main/java/io/prometheus/client/exemplars/api/ExemplarConfig.java
fstab: I removed the comment. It makes sense to have these as public methods, because a user might want to keep a copy of the original default `ExemplarSampler`.

simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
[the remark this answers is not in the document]
fstab: Ok, using `Long` now and `null` to indicate no timestamp.

fstab:
I pushed another update to make sure that `simpleclient` works even if none of the dependencies (`tracer_common`, `tracer_otel`, `tracer_otel_agen`) is present.

simpleclient/src/main/java/io/prometheus/client/exemplars/HistogramExemplarSampler.java
```diff
@@ -0,0 +1,18 @@
+package io.prometheus.client.exemplars;
+
+/**
+ * Exemplar sampler for histogram metrics.
+ */
+public interface HistogramExemplarSampler {
+  /**
+   * @param value      the value to be observed.
+   * @param bucketFrom upper boundary of the previous bucket in the histogram.
+   *                   May be {@link Double#NEGATIVE_INFINITY} if there is no previous bucket.
```
brian-brazil, line 10: Will be

There's no uncertainty here.

README.md
```diff
@@ -291,6 +296,107 @@ class YourClass {
...
+
+```
+request_duration_histogram_bucket{path="/god-of-fire",le="0.004"} 4.0 # {trace_id="043cd631811e373e4180a678c06b128e",span_id="cd122e457d2ca5b0"} 0.0033 1618261159.027
+```
+
+Note that this is an example application for a unit test, so durations don't represent real durations, and some example metrics might not make sense in the real world.
+
+### Enabling And Disabling Exemplars
+
+By default, Exemplars are enabled if OpenTelemetry tracing is detected. You can disable this globally with:
```
brian-brazil, line 334: Does this need updating?

simpleclient/src/main/java/io/prometheus/client/exemplars/ExemplarConfig.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Static configuration class for Exemplar behavior.
+ */
+public class ExemplarConfig {
+
+  private static final ExemplarConfig instance = new ExemplarConfig();
+
+  private final ExemplarSampler noopExemplarSampler;
+  private final ExemplarSampler defaultExemplarSampler;
```
brian-brazil, line 11: Shouldn't these be static? They aren't tied to an instance.

simpleclient/src/main/java/io/prometheus/client/exemplars/ExemplarConfig.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public static void setHistogramExemplarSampler(HistogramExemplarSampler histogramExemplarSampler) {
+    if (histogramExemplarSampler == null) {
+      throw new NullPointerException();
+    }
+    instance.histogramExemplarSampler = histogramExemplarSampler;
+  }
+
+  /**
+   * Disable exemplars by default. Exemplars can still be enabled for individual metrics in the metric's builder.
```
brian-brazil, line 44: Does this actually work? By the time it is called, static initialisation of metrics will likely already be done.

brian-brazil (commented):
It may make sense for an exemplar to be provided directly at inc/observe time, as exemplars are not limited to spans/traces.

simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java
```diff
@@ -0,0 +1,163 @@
...
+  private void validateLabels(String... labels) {
+    if (labels.length % 2 != 0) {
+      throw new IllegalArgumentException("labels are name/value pairs, expecting an even number");
+    }
+    int charsTotal = 0;
+    for (int i = 0; i < labels.length; i++) {
+      if (labels[i] == null) {
+        throw new IllegalArgumentException("labels[" + i + "] is null");
+      }
+      if (i % 2 == 0) { // label names should be unique
```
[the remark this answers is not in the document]
fstab: Two question:

1. What else would you test apart from not `null` and uniqueness?
2. Do you see any drawbacks if we sort the labels by name? That would give them a defined order when the labels are passed as a `HashMap`.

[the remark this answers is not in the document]
brian-brazil: That the characters match the labelname regex.

Sorting is fine, but a HashMap doesn't have ordering. You'd want a TreeMap for that.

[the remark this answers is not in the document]
fstab: added sorting and the check for the label name regex.

integration_tests/exemplars_otel/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+          <groupId>io.prometheus</groupId>
+          <artifactId>tracer_otel</artifactId>
+        </exclusion>
+        -->
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+      <version>${otel.version}</version>
```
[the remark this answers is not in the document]
fstab: I added the maven wrapper script so that users have a defined version of maven (specified in `.mvn/wrapper/maven-wrapper.properties`) if they call `./mvnw` instead of `mvn`.

integration_tests/example_application/src/main/java/io/prometheus/smoketest/Server.java
```diff
@@ -0,0 +1,18 @@
+package io.prometheus.smoketest;
```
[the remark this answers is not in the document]
fstab: moved

pom.xml
```diff
@@ -43,6 +43,7 @@
     </developers>
 
     <modules>
+        <module>tracer_parent</module>
```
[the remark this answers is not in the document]
fstab: moved to `simpleclient_tracer_parent`

simpleclient/src/main/java/io/prometheus/client/exemplars/CounterExemplarSampler.java
```diff
@@ -0,0 +1,18 @@
+package io.prometheus.client.exemplars;
+
+/**
+ * Exemplar sampler for counter metrics.
+ */
+public interface CounterExemplarSampler {
+
+  /**
+   * @param increment     the value added to the counter on this event
+   * @param newTotalValue the new total counter value after adding the increment.
```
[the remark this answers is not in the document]
fstab: ok, I removed it

simpleclient/src/main/java/io/prometheus/client/exemplars/HistogramExemplarSampler.java
fstab: fixed

README.md
fstab: I re-wrote the exemplars documentation.

simpleclient/src/main/java/io/prometheus/client/exemplars/ExemplarConfig.java
fstab: Made them static.

simpleclient/src/main/java/io/prometheus/client/exemplars/ExemplarConfig.java
fstab: I added the following to `README.md` to clarify this
> Note that `ExemplarConfig` is used when a metric is created, so changing the default has no effect on metrics that have already been created.

If we want API to change the `ExemplarSampler` for existing metrics, we'd need to make access to the `ExemplarSampler` thread safe, which would create unexpected synchronization in multi threaded user applications. I don't see a good way of doing this apart from requiring the user to call `ExemplarConfig` before metrics are created.

fstab:
Thanks again for your review.

Apart from addressing your comments, I added `...withExemplar()` versions for the `Counter`'s `inc()` and the `Histogram`'s `observe()` and `time()` methods. Now users can now control Exemplars on three levels:
* Globally by setting the default exemplar sampler
* Per metric by setting an exemplar sampler in the metric builder
* Per observation by useing the `...withExemplar()` methods

I described this in `README.md`. I also split the documentation so that generic Exemplar topics are covered in `README.md` and OpenTelemetry specific topics in `OTEL_EXEMPLARS.md`.

I did not find the time yet to re-run all benchmarks, I will do that before the merge.

I tested all combinations of available / unavailable dependencies and all available OpenTelemetry Java versions. I didn't find any combination that breaks anything.

If the OpenTelemetry tracer is detected, Exemplars for OpenTelemetry are enabled by default. I added a few options to disable this
* In code by calling `ExemplarConfig.disableExemplarSamplers()`
* At build time by excluding the `tracer_otel` and `tracer_otel_agent` dependency
* At runtime with system properties or environment variables.

I documented this in `OTEL_EXEMPLARS.md`.

I like the idea that if users attach the OpenTelemetry Java agent to get traces, they will also see Exemplars by default. And I don't see why people would use distributed tracing with OpenTelemetry without having corresponding exemplars in their metrics. However, if you think this is a bad idea, I am happy to turn this around and disable OpenTelemetry exemplars by default.

README.md
```diff
@@ -291,6 +296,125 @@ class YourClass {
...
+```
+
+### Per Observation Exemplars
+
+You can explicitly create an exemplar for an individual observation. This takes precedence over the exemplar sampler configured with the metric.
+
+The following call will increment a counter, and create an exemplar with the specified `span_id` and `trace_id` labels:
+
+```java
+myCounter.incWithExemplar(Exemplar.SPAN_ID, "abcdef", Exemplar.TRACE_ID, "123456");
```
brian-brazil, line 399: I'd include the literal string here, rather than pointing off to an unspecified class that you have to import to get to a constant.

README.md
```diff
@@ -291,6 +296,125 @@ class YourClass {
...
+```
+
+The following call will observe a value of `0.12` in a histogram, and create an exemplar with the specified `span_id` and `trace_id` labels:
+
+```java
+myHistogram.observeWithExemplar(0.12, Exemplar.SPAN_ID, "abcdef", Exemplar.TRACE_ID, "123456");
+```
+
+All methods for observing and incrementing values have `...withExemplar` equivalents. There are versions taking the exemplar labels as a `String...` as shown in the example, as well as versions taking the exemplar labels as a `Map<String, String>`.
```
brian-brazil, line 408: Have you thought about doing this as part of inc() call? This isn't Go, function overloading is possible and normal in Java.

benchmark/src/main/java/io/prometheus/benchmark/ExemplarsBenchmark.java
brian-brazil: This comment still stands.

integration_tests/exemplars_otel/pom.xml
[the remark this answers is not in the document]
brian-brazil: My comment stands, you can't assume that users are using maven.

simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java
```diff
@@ -0,0 +1,188 @@
...
+public class Exemplar {
+
+  public static final String TRACE_ID = "trace_id";
+  public static final String SPAN_ID = "span_id";
+
+  private final String[] labels;
+  private final double value;
+  private final Long timestampMs;
+
+  private static final Pattern labelNameRegex = Pattern.compile("[a-zA-Z_]\\w*"); // \w is [a-zA-Z_0-9]
```
brian-brazil, line 19: This comment only confuses matters, if you want to say `[a-zA-Z_0-9]` then say `[a-zA-Z_0-9]` directly.

simpleclient/src/main/java/io/prometheus/client/exemplars/ExemplarConfig.java
```diff
@@ -0,0 +1,78 @@
...
+    ExemplarConfig.histogramExemplarSampler = histogramExemplarSampler;
+  }
+
+  /**
+   * Disable the implicit exemplar samplers by default.
+   * Exemplars can still be enabled for individual metrics in the metric builder,
+   * or they can be provided explicitly for individual observations with the {@code ...withExemplar()} methods.
+   */
+  public static void disableExemplarSamplers() {
+    enabled = false;
```
brian-brazil, line 48: This is still racey, you don't know when this could be called relative to when Counters and Histograms are instantiated.

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/ExemplarTest.java
```diff
@@ -0,0 +1,402 @@
...
+      return new Exemplar(increment, timestamp, TRACE_ID, traceId, SPAN_ID, spanId);
+    }
+
+    @Override
+    public Exemplar sample(double value, double bucketFrom, double bucketTo, Exemplar previous) {
+      return new Exemplar(value, timestamp, TRACE_ID, traceId, SPAN_ID, spanId);
+    }
+  }
+
+  private void assertOldFormat(String line) throws IOException {
```
brian-brazil, line 373: Old and New here is a bit confusing, I'd suggest using their actual names.

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/ExemplarTest.java
```diff
@@ -0,0 +1,402 @@
...
+
+  private void assertNewFormat(String line) throws IOException {
+    StringWriter writer = new StringWriter();
+    TextFormat.writeOpenMetrics100(writer, registry.metricFamilySamples());
+    String metrics = writer.toString();
+    Assert.assertTrue("Line not found. Expected metric:\n" + line + "Actual metrics:\n" + metrics,
+        metrics.contains("\n" + line));
+  }
+
+  private String timestampString() {
```
brian-brazil, line 389: Why are you creating a copy of this function in the tests?

simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java
fstab: true

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/ExemplarTest.java
fstab: Removed the copy and used the version from `TextFormat`.

README.md
fstab: Removed the `TRACE_ID` String from `Exemplar`.

simpleclient/src/main/java/io/prometheus/client/exemplars/ExemplarConfig.java
brian-brazil: > I added a line to the JavaDoc saying that this must be called before a Counter or Histogram is created.

That's not possible for users to do, as metrics are usually defined statically and thus users have no control over it.

> However, in order to implement this we would need to make access to the exemplar sampler thread safe, which might severely slow down multi-threaded use of the metrics.

If you only want to read the current state of the sampler, that should always be fast as you don't need mutexes. It's only when mixing reads and writes that things can get a tad slow.

README.md
brian-brazil: If someone wanted an explicitly empty exemplar, they could pass an empty map.

simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java
brian-brazil: > Does that address your concerns?

No, the problem was never maven. It was some other Java build tool that was reading the poms.

README.md
brian-brazil: Did you forget to push? I'm not seeing a change here.

brian-brazil (commented):
I think you forgot to push.

fstab:
Sorry, I removed the reference to `Exemplar.SPAN_ID` from all code occurrences, but missed the one in `README.md`.

> If you only want to read the current state of the sampler, that should always be fast as you don't need mutexes. It's only when mixing reads and writes that things can get a tad slow.

Ok, then I'll refactor this such that you can change the exemplar sampler for existing metrics.

fstab:
I pushed another update. Now changes in `ExemplarConfig` will be picked up by existing Counters and Histograms.
I also ran all benchmarks again, as this was still on my todo list.

Regarding the variables in `pom.xml` files I think it's fair to require users to use Maven or the `./mvnw` script to build the project from source. I'm not aware of an alternative build tool that uses the `pom.xml` files. However, if you think there are users out there using such a tool I am happy to replace the variables with the version numbers.

README.md
```diff
@@ -291,6 +296,123 @@ class YourClass {
...
+    .name("my_latency")
+    .help("help")
+    .withExemplarSampler(new MyExemplarSampler())
+    ...
+    .register();
+```
+
+### Per Observation Exemplars
+
+You can explicitly create an exemplar for an individual observation. This takes precedence over the exemplar sampler configured with the metric.
```
brian-brazil, line 392: s/create/provide/

integration_tests/exemplars_otel_agent/src/test/resources/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM openjdk:11-jre
+RUN mkdir -p /app/
+ADD https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v1.1.0/opentelemetry-javaagent-all.jar /app/
```
brian-brazil, line 3: 1.2.0?

integration_tests/exemplars_otel/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+    <otel.version>1.2.0</otel.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.prometheus</groupId>
+      <artifactId>simpleclient</artifactId>
+      <version>${project.version}</version>
+      <exclusions>
+        <!-- uncomment the following to verify manually that simpleclient still works if the tracers are excluded -->
```
brian-brazil, line 25: Could this be another integration test?

simpleclient/src/main/java/io/prometheus/client/Counter.java
```diff
@@ -73,41 +81,83 @@
...
+     */
+    public Builder withExemplars() {
+      this.exemplarsEnabled = TRUE;
+      return this;
+    }
+
+    /**
+     * Disable exemplars.
+     * <p>
+     * This will disable exemplars for this counter even if exemplars are enabled by default in {@link ExemplarConfig}.
```
brian-brazil, line 133: This reads as though it prevents all use of exemplars, whereas it's only the automatic ones that it stops. This could be clearer.

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/ExemplarTest.java
```diff
@@ -0,0 +1,391 @@
...
+    Counter noLabelsDefaultExemplar = Counter.build()
+        .name("no_labels_default_exemplar")
+        .help("help")
+        .register(registry);
+    noLabelsDefaultExemplar.inc(3);
+    assertOpenMetrics100Format("no_labels_default_exemplar_total 3.0 # " + defaultExemplarLabels + " 3.0 " + timestampString() + "\n");
+    assert004Format("no_labels_default_exemplar_total 3.0\n");
+    noLabelsDefaultExemplar.inc(2);
+    // The TestExemplarSampler always produces a new Exemplar.
+    // The Exemplar with value 3.0 should be replaced with an Exemplar with value 5.0.
```
brian-brazil, line 60: This comment isn't right, the exemplar value is 2. It's the counter value that is 5.

brian-brazil (commented):
This looks roughly right.

> However, if you think there are users out there using such a tool I am happy to replace the variables with the version numbers.

As I said it's a complaint I got years ago, for which I forget the exact details. It's your call.

benchmark/README.md
```diff
@@ -23,97 +24,96 @@ Accordingly, in terms of client instrumentation performance I suggest the follow
...
-    i.p.b.CounterBenchmark.codahaleCounterIncBenchmark                    avgt        5   11.554 ± 0.251  ns/op
-    i.p.b.CounterBenchmark.codahaleMeterMarkBenchmark                     avgt        5   75.305 ± 7.147  ns/op
-    i.p.b.CounterBenchmark.prometheusCounterChildIncBenchmark             avgt        5   13.249 ± 0.029  ns/op
-    i.p.b.CounterBenchmark.prometheusCounterIncBenchmark                  avgt        5  127.397 ± 4.072  ns/op
-    i.p.b.CounterBenchmark.prometheusSimpleCounterChildIncBenchmark       avgt        5   12.989 ± 0.285  ns/op
-    i.p.b.CounterBenchmark.prometheusSimpleCounterIncBenchmark            avgt        5   54.822 ± 7.994  ns/op
-    i.p.b.CounterBenchmark.prometheusSimpleCounterNoLabelsIncBenchmark    avgt        5   13.131 ± 1.661  ns/op
+    i.p.c.b.CounterBenchmark.codahaleCounterIncBenchmark                    avgt        5   8.063 ± 0.440  ns/op
+    i.p.c.b.CounterBenchmark.codahaleMeterMarkBenchmark                     avgt        5  38.625 ± 2.111  ns/op
+    i.p.c.b.CounterBenchmark.prometheusCounterChildIncBenchmark             avgt        5   9.869 ± 0.252  ns/op
```
brian-brazil, line 33: Incidentally, we can remove these old Java client benchmarks. They've not been relevant in many years.

README.md
fstab: changed

integration_tests/exemplars_otel_agent/src/test/resources/Dockerfile
fstab: Seems that the agent release lacks a few days behind the library release. Agent is still 1.1.0.

integration_tests/exemplars_otel/pom.xml
fstab: Good idea, I turned this into an integration test. Instead of excluding dependencies in the `pom.xml`, the new implementation of `ExemplarsOtelIT` copies the individual JARs into the Docker image so that they can be removed in the test without rebuilding the Maven project.

While I was at it, I also renamed the `tracer_...` sub-modules to `simpleclient_tracer_...`.

simpleclient/src/main/java/io/prometheus/client/Counter.java
fstab: Changed the wording.

simpleclient_common/src/test/java/io/prometheus/client/exporter/common/ExemplarTest.java
fstab: fixed

benchmark/README.md
fstab: Do you mean remove from `README.md`, or remove the implementations as well? Are there benchmarks that are still relevant apart from the `ExemplarsBenchmark`?

brian-brazil: The implementations for the old java client. The benchmarks are still relevant, that client isn't however.

fstab: Ok, I removed the old client from the benchmarks.

fstab:
I removed the obsolete benchmarks, and I bumped the OpenTelemetry agent version as 1.2.0 was released today. What do you think?

OTEL_EXEMPLARS.md
```diff
@@ -0,0 +1,87 @@
...
+      </exclusion>
+    </exclusions>
+  </dependency>
+</dependencies>
+```
+
+### Disable OpenTelemetry Exemplars at Runtime
+
+If your application uses OpenTelemetry tracing, but you want to disable OpenTelemetry at runtime without changing code,
+start your application with the `otelExemplars` system property:
```
brian-brazil, line 76: This should probably be namespaced a bit more, so that it's tied to this client. As-is I'd presume from the name that it was an OTel internal.

OTEL_EXEMPLARS.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+java -DotelExemplars=inactive -jar my-application.jar
+```
+
+Alternatively, you can set the environment variable `OTEL_EXEMPLARS=inactive`:
```
brian-brazil, line 82: I'd recommend sticking with one way of doing things to avoid confusion. System properties are the way to do this in Java.

simpleclient_tracer/simpleclient_tracer_otel/src/main/java/io/prometheus/client/exemplars/tracer/otel/OpenTelemetrySpanContextSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+public class OpenTelemetrySpanContextSupplier implements SpanContextSupplier {
+
+  public static boolean isAvailable() {
+    try {
+      if ("inactive".equalsIgnoreCase(System.getProperties().getProperty("otelExemplars", System.getenv("OTEL_EXEMPLARS")))) {
+        return false;
+      }
+      OpenTelemetrySpanContextSupplier test = new OpenTelemetrySpanContextSupplier();
+      test.getSpanId();
+      test.getSpanId();
```
brian-brazil, line 17: Should this be getTraceId?

simpleclient_tracer/simpleclient_tracer_otel_agent/src/main/java/io/prometheus/client/exemplars/tracer/otel_agent/OpenTelemetryAgentSpanContextSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+public class OpenTelemetryAgentSpanContextSupplier implements SpanContextSupplier {
+
+  public static boolean isAvailable() {
+    try {
+      if ("inactive".equalsIgnoreCase(System.getProperties().getProperty("otelExemplars", System.getenv("OTEL_EXEMPLARS")))) {
+        return false;
+      }
+      OpenTelemetryAgentSpanContextSupplier test = new OpenTelemetryAgentSpanContextSupplier();
+      test.getSpanId();
+      test.getSpanId();
```
brian-brazil, line 22: Same here.

brian-brazil (commented):
Gave it a deeper review, and found a few nits.

OTEL_EXEMPLARS.md
```diff
@@ -0,0 +1,87 @@
...
+curl -H 'Accept: application/openmetrics-text; version=1.0.0; charset=utf-8' http://localhost:8080/metrics
+```
+
+You should see metrics with Exemplars, for example in the `request_duration_histogram` metric:
+
+```
+request_duration_histogram_bucket{path="/god-of-fire",le="0.004"} 4.0 # {trace_id="043cd631811e373e4180a678c06b128e",span_id="cd122e457d2ca5b0"} 0.0033 1618261159.027
+```
+
+Note that this is an example application for a unit test, so durations don't represent real durations, and some example metrics might not make sense in the real world.
```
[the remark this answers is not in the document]
fstab: changed

OTEL_EXEMPLARS.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Disabling OpenTelemetry Exemplars
+
+If you use OpenTelemetry tracing but don't want Exemplars, you can disable OpenTelemetries in multiple ways.
```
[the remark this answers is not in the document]
fstab: changed

OTEL_EXEMPLARS.md
fstab: Removed the environment variable.

OTEL_EXEMPLARS.md
fstab: Removed the environment variable.

simpleclient_tracer/simpleclient_tracer_otel/src/main/java/io/prometheus/client/exemplars/tracer/otel/OpenTelemetrySpanContextSupplier.java
fstab: Yes, good point. Fixed.

simpleclient_tracer/simpleclient_tracer_otel_agent/src/main/java/io/prometheus/client/exemplars/tracer/otel_agent/OpenTelemetryAgentSpanContextSupplier.java
fstab: Fixed.

fstab:
Thanks a lot for reviewing again. I pushed another update. What do you think?

brian-brazil:
:+1:

fstab:
Thanks a lot for your great review!

Issue #649, opened 2021-04-01
documented for production metric names are changed by 0.10.0 client


ndimitry:
We use Prometheus to collect and store metrics (so, a big thanks for the effort maintaining and enhancing it).

At the moment a lot of names are already documented and used for years so for us this is not a desirable behaviour to have "_total" being added to each counter (this breaks both documentation and precreated Grafana dashboards).

Is there some way to keep names unchanged? Or maybe some flag could be introduced to keep compatibility? If not there should (I'm sure) be an idea behind that, in case so, please, explain it to me.

Thanks in advance for the reply!

fstab:
Hi ndimitry, the idea is to become compatible with the [OpenMetrics](https://openmetrics.io/) standard. However, it might be good to introduce an "allow invalid counter names" flag in the next release so that people who cannot update their Grafana dashboards are not stuck forever with the previous release.

ndimitry:
Hi @fstab , thank you for an explanation!
I think I will create a pull request to make it possible to allow invalid counter names.
Hopefully we will finally move to the new standard too. Do you know if there's a chance that the draft will change and "_total" suffix will be (for example) replaced by other suffix? I'm asking that to try to avoid cases when we adopt to the new standard and then run into the case when we need to further adupt to the newer changes.

Issue #640, opened 2021-03-17
0.10.0 introduced a breaking change on metric names without an option to disable it


niwinz:
On my testing environment I have detected that many counters are suffixed with `_total` and added a new metric with `_created` suffix. After reading the release notes, I realized of the **breaking change**. Maybe some more emphasis on the BREAKING changed would be awesome.

In our case we don't want change our metric names, so we will not be able to upgrade to the latest versions :(

I think this kind of breaking changes should be introduced gradually, making them optional in any case.
A simple configuration flag on the registry will solve the issue.

fstab:
Thanks for your feedback. It's too late to change this now, but the good news for the future is: `client_java` 0.10.0 implements the [OpenMetrics](https://openmetrics.io/) standard, so this will remain pretty stable in the future. So if you manage to update your metric names it will very likely be the last time you have to do this.

niwinz:
I think you still have the option to add a flag for disable this new behavior for people on the same situation as me. As far as I know, prometheus does not support reniming existing metrics (maybe I'm wrong).

In the worse case and without an option to disable this new behavior, we probably will need to go to fork and vendor the 0.9.0 version, because changing the name right now is not an option.

hirunatan:
Our problem is that we apparently cannot change the name of existing metrics. And If we create new metrics, we lose all previous data and everything starts from zero.

If there is a way of renaming metrics without losing data it would also be ok.

BoopBoopBeepBoop:
Echoing the discontent on this one ... perhaps our fault for not picking up the breaking change in the release notes, but we just spent more than a little while trying to figure out why our dashboards were missing seemingly arbitrary subsets of metrics. Pretty significant breaking change anywhere that someone has a query wired to one of these metrics, I think.

PR #653, opened 2021-04-19
allowing to keep invalid counter names (with no _total suffix)

ndimitry:
In some cases keeping broken counter names (with no _total suffix) will be really helpful. For example when old names are documented and are in use with a lot of dashboards being precreated for them. Or in case it is sugnificant not to loose connection to old data when counters are renamed.

Possibly this might be a grace option allowing some grace period (for a year or two) for the teams to move to the new names. For us this is quite an effort shared between several development teams. And we are afraid that OpenMetrics format might change too because it is still an RFC draft. When OpenMetrics are becoming an RFC this will be steady argument for us to move to this format.

Links to the discussions https://github.com/prometheus/client_java/issues/640, https://github.com/prometheus/client_java/issues/649

### DIFF
--- a/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
+++ b/simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
@@ -25,6 +25,14 @@ public class TextFormat {
   public final static String CONTENT_TYPE_OPENMETRICS_100 = "application/openmetrics-text; version=1.0.0; charset=utf-8";
 
   /**
+   * A property to be passed via the command line (-D...="true") to allow old-style pre OpenMetrics counter names,
+   * in other words invalid counter names (not ending with the suffix "_total")
+   *
+   * @since 0.10.1
+   */
+  private final static String ALLOW_INVALID_COUNTER_NAMES = "io.prometheus.allowInvalidCounterNames";
+
+  /**
    * Return the content type that should be used for a given Accept HTTP header.
    *
    * @since 0.10.0
@@ -64,6 +72,10 @@ public class TextFormat {
    * Write out the text version 0.0.4 of the given MetricFamilySamples.
    */
   public static void write004(Writer writer, Enumeration<Collector.MetricFamilySamples> mfs) throws IOException {
+    // true in case ALLOW_INVALID_COUNTER_NAMES property is present and is true (ignoring case),
+    // false in all other cases (including missing value)
+    boolean allowInvalidCounterNames
+            = Boolean.parseBoolean(System.getProperties().getProperty(ALLOW_INVALID_COUNTER_NAMES));
     Map<String, Collector.MetricFamilySamples> omFamilies = new TreeMap<String, Collector.MetricFamilySamples>();
     /* See http://prometheus.io/docs/instrumenting/exposition_formats/
      * for the output format specification. */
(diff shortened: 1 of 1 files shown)


ndimitry:
@fstab please, review

fstab:
Hi, I finally got back to this, and I have a few points:

* The property only affects the text format 004, not the OpenMetrics format. This is good, because in OpenMetrics a counter without the `_total` prefix would be invalid. However, as far as I know there is no way to make a new Prometheus server release use the old format, as the `Accept` header is hard-coded here https://github.com/prometheus/prometheus/blob/03b354d4d9386e4b3bfbcd45da4bb58b182051a5/scrape/scrape.go#L686-L686. As soon as you switch to a new version of Prometheus you will have the `_total` suffix again.
* The change removes `_total` from all counters, even if there is a counter explicitly named `_total`.

Instead of starting the monitored application with `-Dio.prometheus.allowInvalidCounterNames=true` you could achieve the same effect by adding the following to your Prometheus scrape config:

```yaml
   metric_relabel_configs:
    - source_labels: [__name__]
      regex:  '(.*)_total'
      target_label: __name__
      replacement: '${1}'
```

Given the new issues this PR would introduce (result depends on the Prometheus version, and it is no longer possible to name a counter `..._total` I think it might be better to use a relabel config as a temporary fix until dashboards are adapted.

ndimitry:
@fstab thank you for the possible workaround description!

I guess, let's have this pull request closed then...

travisspencer:
I agree with @franck102. This should be reopened. Upgrading to 0.12 was required to get Jakarta servlet support (cf. #647), but this breaking change came in at the same time. Now, we're blocked -- we can't upgrade Jetty to v. 11 because that requires Jakarta not javax servlets, but upgrading this lib to work with the new Jetty version brings in this breaking change.

What I don't understand is how the resource representation changes if the client keeps asking for `text/plain; version=0.0.4`. If my client keeps negotiating that format, why does it get anything different? Seems it's not really `0.0.4` format any more (or the version is unstable).

Help in resolving this conundrum is much appreciated.

travisspencer:
@fstab , can you or one of the other maintainers of this project comment on my previous input? WDYT?

fstab:
Setting `io.prometheus.allowInvalidCounterNames=true` will remove the `_total` suffix from all counters, even if there is a counter explicitly named `_total`. So if you now have an issue with `_total` suffixes being added to all counters, with this PR you might run into an issue because `_total` suffixes are removed from all counters.

travisspencer:
That sounds like it could still produce a breaking change, @fstab , if I have counters named `..._total` already. I will check this, but would like to know which property I need to set or how I can use this library to make it work as it did in the previous version. If that's not doable, can the Jakarta support be back-ported?

fstab:
No, I am talking about this PR here: If we merge this PR, then we'll get a property `io.prometheus.allowInvalidCounterNames` that would remove all `_total` suffixes from counters. That's why I don't want to re-open this.

travisspencer:
Ah, I see your point, @fstab. I think then what should be reopened is #640. (I should have made my comment there and not on this PR, it seems.) Would you agree with reopening that issue?

JnRouvignac:
I agree that this PR may not be the right way to restore the old behaviour with respect to counter metric names.

I could start working on a new PR that adds a `io.prometheus.allowInvalidCounterNames` system property to restore the older behaviour.
Would you agree with merging such a PR @fstab if it satisfied you?

JnRouvignac:
I have noticed #791 which adds `PROMETHEUS_DISABLE_CREATED_SERIES`, so I think the answer is yes.
I'll look into doing a PR to allow smoother migration.
Thanks.

PR #658, opened 2021-05-02
Add matrix build of JDK versions

SuperQ:
Test build on various OpenJDK versions.

Signed-off-by: SuperQ <superq@gmail.com>


SuperQ:
Something was broken with the CircleCI webhook. I removed the integration and re-added. Looks like it's building again.

fstab:
FYI: I added TestContainers as part of my OpenTelemetry exemplars branch to test compatibility with different Java versions (including Java 6, which causes the most problems):

https://github.com/prometheus/client_java/blob/103eb5961d765b64ce145a50f9e24b84d5d61a21/integration_tests/java_versions/src/test/java/io/prometheus/smoketest/JavaVersionsIT.java#L35-L43

It's not merged yet.

These are two different things: The TestContainers tests make sure that the binary `client_java` library is compatible with different Java versions, while your change makes sure that you can build the library from source with different versions.

It would be perfectly fine if we require Java 11 to build `client_java` but release a library that is compatible with Java 6.

fstab:
TestContainers use Docker to run tests against Docker containers. In order to make them work with CircleCI, we need to switch to machine images:

https://github.com/prometheus/client_java/blob/103eb5961d765b64ce145a50f9e24b84d5d61a21/.circleci/config.yml#L1-L18

Not sure if we can convert your PR to use machine images.

fstab:
I like the idea to have a clear definition which Java versions can be used to build `client_java`, and I think it's a great idea to have automated tests for that. It would be perfect if we could find a way to do this and still support Docker-based `TestContainers`.

SuperQ:
We can run the `TestContainers` with `setup_remote_docker`. We can start with any base image like `cimg/base`.

So basically something like this:

```yaml
jobs:
  test_containers;
    steps:
    - checkout
    - setup_remote_docker
    - run: docker docker docker
```

fstab:
I am hoping to get the OpenTelemetry exemplars branch done this week. I'll try the `setup_remote_docker` then.

fstab:
Note: `client_java` now has a Jetty Jakarta Servlet example (https://github.com/prometheus/client_java/blob/master/simpleclient_servlet_jakarta/src/test/java/io/prometheus/client/exporter/ExampleBenchmark.java) which requires Java 11 to build.

PR #659, opened 2021-05-06
Add "jvm_compilation_time_total" counter (CompilationMXBean.getTotalCompilationTime)

rainerjung:
The subject says it all. It is one of the few missing metrics from the standard platform MBeans.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/CompilationExports.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/CompilationExports.java
@@ -1,0 +1,49 @@
+package io.prometheus.client.hotspot;
+
+import io.prometheus.client.Collector;
+import io.prometheus.client.CounterMetricFamily;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.CompilationMXBean;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Exports metrics about JVM compilation.
+ * <p>
+ * Example usage:
+ * <pre>
+ * {@code
+ *   new CompilationExports().register();
+ * }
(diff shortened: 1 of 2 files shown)


fstab:
Hi thanks a lot for your PRs. I saw them, and I will get back to this as soon as I found time to test this with different Java versions (IBM's J9, etc.).

fstab (changes requested):
Thanks a lot for your PRs. I have a couple of comments on this one:

Please wrap the code in `addCompilationMetrics()` with
```java
if (compBean != null && compBean.isCompilationTimeMonitoringSupported()) {
    // ...
}
```
to avoid `NullPointerException` and `UnsupportedOperationException`.

Prometheus metrics should have their base unit in their name, so it would be good to rename it to `jvm_compilation_time_seconds_total`.

The new metric should be added to `DefaultExports.register()` (btw when you touch this function you could sort the Exports alphabetically because it's currently hard to check if one is missing).

~~It could be handy to add the compiler (as returned by `compBean.getName()`) as a label, so that the metric looks like `jvm_compilation_time_seconds_total{compiler="j9jit29"}`~~

Update on the last point: it would be more in line with Prometheus conventions to add the compiler to an info metric instead of a label in `jvm_compilation_time_seconds_total`.

dhoard:
@rainerjung are you planning on implementing the suggestions? ... if not, I'll submit a PR based on the your work and the suggestions.

dhoard:
@rainerjung new PR https://github.com/prometheus/client_java/pull/797 create.

Since `getTotalCompilationTime()` returns milliseconds, I chose to expose the value as is.

dhoard:
@rainerjung please close this PR since I have picked up the work. Thanks!

PR #667, opened 2021-06-11
Add Counter.Child no-arg constructor

snago:
Fix backwards incompatible change introduced in the "exemplar commit", b76e1e95d1c8ce3141584e2a2d260519c3a6eb19

After upgrading to 0.11.0 the following no longer compiled:
```java
    private final Counter completedTaskCounter =
            Counter.build()
                   .name("foo_completed_tasks")
                   .help("The number of tasks completed.")
                   .create()
                   .setChild(new Counter.Child() {
                       @Override
                       public double get() {
                           return completedTasks.longValue();
                       }
                   });
```

The exemplars commit added a new constructor to `Counter.Child` which meant that the implicit no-arg constructor was removed.

As a work-around I've changed to `new Counter.Child(null, null)`, but this fix might save someone else from compilation errors after upgrading.

Signed-off-by: Robin Karlsson <snago86@gmail.com>


fstab (approved):
Good point, thanks for the PR.

PR #671, opened 2021-07-06
Add Exemplar.toString

anuraaga:
Useful for example, in test assertions, see https://github.com/open-telemetry/opentelemetry-java/pull/3353/files#diff-10589ce2b6c8942193e382b96c566ac748a995454f2d7527b2bb7e762cd520f1R422

@fstab

/cc @jseureth

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java
+++ b/simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java
@@ -187,8 +187,10 @@ public class Exemplar {
   public String toString() {
     StringBuilder sb = new StringBuilder("Exemplar{value=");
     sb.append(value);
-    sb.append(", ts=");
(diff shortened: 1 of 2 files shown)


fstab:
Thanks for the PR. It would be useful to produce the same format as `TextFormat.writeOpenMetrics100()`, so that the `toString()` representation doesn't differ from the OpenMetrics format. What do you think?

anuraaga:
Thanks @fstab - I took a look in but it seems quite difficult since `TextFormat` has some fairly complicated private methods such as `omWriteTimestamp`. Not to mention that `TextFormat` is not in `simpleclient`. So it's an interesting idea but I think unless we were to move at least some `TextFormat` logic into `simpleclient` it doesn't seem practical, it's essentially a circular dependency.

anuraaga:
Very old PR that I should probably close :D

Issue #678, opened 2021-08-04
Few issues around of OSGi-specific headers in manifests


bigmarvin:
I'm glad to see most artifacts of this project are packed as bundles, while I find some little glitches as listed below.

1. The artifacts of name `simpleclient_tracer_*` are not packed as bundles while other bundles like `simpleclient` itself import packages from them, which fails the startup of OSGi runtime.
2. The resolution of all imported packages is mandatory by default, but I doubt if every client would use every imported package every time. It's better to make them optional so the runtime knows which version of packages to look for when necessary and yet no startup is blocked.

fstab:
Thanks for bringing this up. I've never looked into OSGI. Could you let me know a simple way to set up an OSGI project so that I could try this? Maybe I can turn it into an automated test.

bigmarvin:
Here are some resources I planned to explore when available. Wish them help!

https://felix.apache.org/documentation/subprojects.html

bigmarvin:
Thanks for addressing the 1st part! Let me try to illustrate the 2nd part as following.

The primary advantage of OSGi would be enabling same class of different versions co-work properly in one JVM. It's achieved by building the topology of versioned packages provided/required by bundles (called wiring).

During the start-up of OSGi runtime, it would check if this topology can be built or not. If some package required by some bundle is mandatory while no bundle can provide it, the start-up would fail. That's what I ran into in the first place. If the required package is optional, on the other hand, the start-up won't fail at this phase at least.

Optional resolution is handy when, say some classes under some target package integrate with many partners or so (e.g. some middleware integrating with all kinds of JMS implementation). However, as some client of this target package, it would only integrate with one or few partners in real world. If the packages of all partners are mandatorily required, the client must also include all package providers in the runtime for wiring to work but effectively most of them would be never used. If the resolution is optional, such massive inclusion wouldn't be necessary.

The drawback would be, however, if the packages of some partner **in use** are absent, e.g. not provided by any bundle, the start-up would still fail when the client bundle tries to load some class using that partner in the target package. Though postponed from wiring phase to activating phase, this error is expected and can be fixed by including the provider of these missing classes (usually some artifact of the partner). OSGi runtime would always use the info of import-package instructions to locate qualified packages regardless of its resolution, mandatory or optional.

[...]

PR #679, opened 2021-08-04
[issues/678] update pom.xml accordingly

bigmarvin:
To address https://github.com/prometheus/client_java/issues/678 --

1. make the artifacts of name `simpleclient_tracer_*` packed as bundles
2. make the resolution of imported packages optional by default

@fstab and @tomwilkie, may I ask you maintainers to take a look?

If possible, I'm also requesting the delivery of this fix in 0.11.x asap, as the absence of must-import packages prevents the upgrade of `simpleclient` in OSGi runtime.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,11 @@
                     <artifactId>maven-enforcer-plugin</artifactId>
                     <version>1.4.1</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.felix</groupId>
+                    <artifactId>maven-bundle-plugin</artifactId>
+                    <version>2.4.0</version>
+                </plugin>
             </plugins>
         </pluginManagement>
         <plugins>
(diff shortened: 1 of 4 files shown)


fstab:
Oh, I fixed part of #678, and it seems I have overlooked that there is a PR for that. The current status on `master` is: I added `<packaging>bundle</packaging>`, but I'm unsure what `*;resolution:=optional` does, so I didn't add that. I guess some dependencies are not optional. Might it break things if we declare everything optional? I'm not an OSGI user, so I could use some help explaining the effect of `*;resolution:=optional`.

bigmarvin:
> Oh, I fixed part of #678, and it seems I have overlooked that there is a PR for that. The current status on `master` is: I added `<packaging>bundle</packaging>`, but I'm unsure what `*;resolution:=optional` does, so I didn't add that. I guess some dependencies are not optional. Might it break things if we declare everything optional? I'm not an OSGI user, so I could use some help explaining the effect of `*;resolution:=optional`.

Great, thanks for the efforts on this! Let's continue the discussion of 2nd part in #678.

PR #681, opened 2021-08-06
Prevent name conflicts and rename jvm_classes_loaded

fstab:
The `ClassLoadingExports` collector provides a `Gauge` named `jvm_classes_loaded` and a `Counter` named `jvm_classes_loaded_total`. This is an illegal name conflict in OpenMetrics, and this is also the root cause for https://github.com/prometheus/jmx_exporter/issues/621.

This PR renames `jvm_classes_loaded` to `jvm_classes_currently_loaded`.

**This is a breaking change. Prometheus queries referencing `jvm_classes_loaded` need to be adapted to the new name `jvm_classes_currently_loaded`.**

Moreover, the PR adds a check in `CollectorRegistry.register(Collector)` to prevent collectors with such name conflicts from being registered.

The check will be performed if the `Collector` implements `Collector.Describable`, or if `CollectorRegistry.autoDescribe` is `true` (it's `true` by default for the `CollectorRegistry.defaultRegistry` and `false` by default for custom registries).

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -48,16 +48,29 @@ public class CollectorRegistry {
    */
   public void register(Collector m) {
     List<String> names = collectorNames(m);
+    assertNoDuplicateNames(m, names);
     synchronized (namesCollectorsLock) {
       for (String name : names) {
         if (namesToCollectors.containsKey(name)) {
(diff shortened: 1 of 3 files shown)


fmmr:
starting solr (and other apps) with prometheus and calling DefaultExports.initialize() now fails: https://github.com/prometheus/client_java/issues/697

fmmr:
sorry -. wrong link in last comment - https://github.com/prometheus/client_java/issues/699

PR #686, opened 2021-08-27
Added SSL support to HTTPServer

dhoard:
Added SSL support to **HTTPServer** when using **HTTPServer.Builder**. Creation of the **SSLContext** is an external concern and should be addressed by the calling code.


dhoard:
I'll fix the PR. Please wait.

PR #687, opened 2021-08-30
Added BasicAuthenticator

dhoard:
Added BasicAuthenticator to allow other projects that use HTTPServer a common implementation.


dhoard:
@fstab please review.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/BasicAuthenticator.java
```diff
@@ -0,0 +1,88 @@
+package io.prometheus.client.exporter;
+
+/**
+ * Basic HTTP authenticator
+ */
+public class BasicAuthenticator extends com.sun.net.httpserver.BasicAuthenticator {
```
fstab, line 6: It would be good to give this class a different name so that users can't confuse `io.prometheus.client.exporter.BasicAuthenticator` and `com.sun.net.httpserver.BasicAuthenticator`.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/BasicAuthenticator.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    @Override
+    public boolean checkCredentials(String username, String password) {
+        if ((this.username == null) || (this.username.trim().length() == 0)) {
+            return false;
+        }
+
+        if ((this.password == null) || (this.password.trim().length() == 0)) {
+            return false;
+        }
```
fstab, line 33: `this.username` and `this.password` cannot be `null` or empty, so there's no need to check this here.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/BasicAuthenticator.java
```diff
@@ -0,0 +1,88 @@
...
+            return this;
+        }
+
+        /**
+         * Build the authenticator
+         * @return
+         */
+        public BasicAuthenticator build() {;
+            assertNotNullNotEmpty(this.username, "username is null or empty");
+            assertNotNullNotEmpty(this.password, "password is null or empty");
```
fstab, line 72: What about moving this check to the constructor and making the constructor public? Then users could choose whether they want to use the `Builder` or not.

fstab (changes requested):
Thanks a lot! I added a few minor review comments.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/BasicAuthenticator.java
dhoard: Agreed.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/BasicAuthenticator.java
dhoard: I will refactor to simplify the code.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/BasicAuthenticator.java
dhoard: Thoughts on renaming to `HTTPServerBasicAuthenticator`?

dhoard: Renamed to `HTTPServerBasicAuthenticator`

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServerBasicAuthenticator.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Basic HTTP authenticator
+ */
+public class HTTPServerBasicAuthenticator extends com.sun.net.httpserver.BasicAuthenticator {
+
+    private static final String REALM = "/";
+
+    private String username;
+    private String password;
```
fstab, line 11: Nitpick: You could make these final to make it clear that `HTTPServerBasicAuthenticator` is an immutable class.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServerBasicAuthenticator.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Check username and password
+     * @param username
+     * @param password
+     * @return
+     */
+    @Override
+    public boolean checkCredentials(String username, String password) {
+        return (this.username.equals(username) && this.password.equals(password));
```
fstab, line 36: Nitpick: Parentheses not needed.

fstab (changes requested):
Thanks a lot. I have two little nitpicks, but apart from that it's looking good. Thanks for your contribution.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServerBasicAuthenticator.java
dhoard: Changed.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServerBasicAuthenticator.java
dhoard: Changed.

Issue #685, opened 2021-08-16
 WARNING: sendResponseHeaders: being invoked with a content length for a HEAD request


yakirgb:
Hi,
I'm using JAR: `jmx_prometheus_httpserver-0.16.1-jar-with-dependencies.jar` to monitor HDFS NameNodes 
Every 60 seconds i'm getting this warning:
```
Aug 16 12:34:59 nn00103 java[22008]: Aug 16, 2021 12:34:59 PM sun.net.httpserver.ExchangeImpl sendResponseHeaders
Aug 16 12:34:59 nn00103 java[22008]: WARNING: sendResponseHeaders: being invoked with a content length for a HEAD request
````
The exporter is working without any problem, only printing warning. 
Thank you,Yakir

dhoard:
Created PR ( https://github.com/prometheus/client_java/pull/688 ) to resolve the issue. @fstab please review.

fstab:
Sorry for the delay. Could you let me know which Java version you are using? I cannot reproduce this.

To me it looks like the warning is wrong, according to to [the HTTP/1.1 Spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13) there must be a `Content-length` header.
> The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, **in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.**

dhoard:
I'm running Azul's OpenJDK version 11...

openjdk version "11.0.12" 2021-07-20 LTS
OpenJDK Runtime Environment Zulu11.50+19-CA (build 11.0.12+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.50+19-CA (build 11.0.12+7-LTS, mixed \`mode)\`

\-

The warning is in the OpenJDK 11 though OpenJDK 13 codebase, so can't explain why you're not seeing it...

https://github.com/openjdk/jdk11u/blob/63fc945c1d4cdc893d89796804acfe0ce9341860/src/jdk.httpserver/share/classes/sun/net/httpserver/ExchangeImpl.java#L238

https://github.com/openjdk/jdk17/blob/a37254c79fa5973465d90f4b52ab88fe68016c9f/src/jdk.httpserver/share/classes/sun/net/httpserver/ExchangeImpl.java#L238

\-

The change was made based Java bug #6886723 and the proposed workaround to remove the warning...

[...]

PR #688, opened 2021-09-01
Added code to correctly handle HEAD request

dhoard:

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -84,6 +84,14 @@ public class HTTPServer implements Closeable {
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
 
+            // Correct HEAD behavior per Java bug #6886723
+            if (t.getRequestMethod().equals("HEAD")) {
+                t.getRequestBody().close();
+                t.getResponseHeaders().add("Transfer-encoding", "chunked");
+                t.sendResponseHeaders(200, -1);
+                return;
+            }
+
             String contextPath = t.getHttpContext().getPath();
             ByteArrayOutputStream response = this.response.get();
             response.reset();
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -83,15 +83,6 @@ public class HTTPServer implements Closeable {
         @Override
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
-
-            // Correct HEAD behavior per Java bug #6886723
-            if (t.getRequestMethod().equals("HEAD")) {
-                t.getRequestBody().close();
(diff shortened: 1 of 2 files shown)


dhoard:
cc @fstab @tomwilkie

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -84,6 +84,14 @@ protected ByteArrayOutputStream initialValue()
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
 
+            // Correct HEAD behavior per Java bug #6886723
+            if (t.getRequestMethod().equals("HEAD")) {
+                t.getRequestBody().close();
+                t.getResponseHeaders().add("Transfer-encoding", "chunked");
+                t.sendResponseHeaders(200, -1);
```
fstab, line 91: According to [the HTTP/1.1 Spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13)
> The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, **in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.**

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -84,6 +84,14 @@ protected ByteArrayOutputStream initialValue()
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
 
+            // Correct HEAD behavior per Java bug #6886723
+            if (t.getRequestMethod().equals("HEAD")) {
+                t.getRequestBody().close();
+                t.getResponseHeaders().add("Transfer-encoding", "chunked");
```
fstab, line 90: The `Content-type` header should be there as well. According to [the HTTP/1.1 Spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4)
> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. **The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request.**

fstab (commented):
Thanks a lot for the PR. I tried to reproduce the bug, but I could not find a single Java version where the `HEAD` request doesn't work. Even with Azul's Java 6 it works. Moreover, the current implementation handles headers correctly, so if I run
```
curl -i --head -H 'Accept: application/openmetrics-text; version=1.0.0; charset=utf-8' http://localhost:9000/metrics
```
I get the correct `Content-type` and `Content-length`.

Are you aware of any Java version where `HEAD` requests don't work? I would only fix it if it is a real problem.

fstab:
I just saw that #685 is linked here, I commented there as well.

dhoard:
I updated the PR. The issue is around the generation of the warning message. I have updated https://github.com/prometheus/client_java/issues/685 with specifics.

fstab (approved):
Ok, I see now what was the issue. Yes, it makes sense to fix this. The API of `sun.net.httpserver.HttpExchangeImpl.sendResponseHeaders()` is really strange, thanks for looking into it.

yakirgb:
@fstab any option to cut new release?

PR #689, opened 2021-09-07
Implement filtering of async requests.

### LINKED ISSUE #676: Add support for async requests in MetricsFilter
It seems to me that [MetricsFilter](https://github.com/prometheus/client_java/blob/master/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java) currently does not support async requests, it reports very low times which I guess are only limited to the time the initial request is answered.

lapo-luchini:
This fixes #676 for me… but there are no tests as I have no idea how to properly test an async servlet responses with Mockito.

Doesn't break any of the existing tests, though:

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running io.prometheus.client.exporter.MetricsServletTest
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.548 sec - in io.prometheus.client.
exporter.MetricsServletTest
Running io.prometheus.client.filter.MetricsFilterTest
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.232 sec - in io.prometheus.client.
filter.MetricsFilterTest

Results :

Tests run: 11, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
@@ -3,6 +3,8 @@ package io.prometheus.client.filter;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Histogram;
 
+import javax.servlet.AsyncEvent;
+import javax.servlet.AsyncListener;
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
 import javax.servlet.FilterConfig;
@@ -162,7 +164,7 @@ public class MetricsFilter implements Filter {
     }
 
     @Override
-    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+    public void doFilter(ServletRequest servletRequest, final ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
         if (!(servletRequest instanceof HttpServletRequest)) {
             filterChain.doFilter(servletRequest, servletResponse);
             return;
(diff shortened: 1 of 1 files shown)


simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
```diff
@@ -172,17 +174,37 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+        boolean isAsync = false;
         try {
             filterChain.doFilter(servletRequest, servletResponse);
+            isAsync = servletRequest.isAsyncStarted();
+            if (isAsync) {
+                servletRequest.getAsyncContext().addListener(new AsyncListener() {
+                    volatile boolean done = false;
+                    private void meter() {
+                        if (!done) {
+                            done = true;
```
fstab, line 192: This isn't thread safe, because there's a _check-then-act_ race condition. I don't think thread safety is needed here, because my understanding is that `meter()` will only be called once. If it's only called once, please remove `done`. If there are cases where it is called multiple times in different threads, fix the race condition.

fstab (changes requested):
This is a nice one, thanks a lot!

Unfortunately, I have moved the `Servlet` filter implementation in the meantime, all common functionality is now in `io.prometheus.client.servlet.common.filter.Filter`, and there are two implementations using it: One for `javax` Servlets, and one for `jakarta` Servlets.

It will require a bit of boiler-plate code to port your solution to the current implementation, but I think it's worthwhile doing it. Basically you would need a common implementation of the `AsyncListener` in `simpleclient_servlet_common`, and an implementation of `jakarta.servlet.AsyncListener` and `javax.servlet.AsyncListener` in `simpleclient_servlet` and `simpleclient_servlet_jakarta`. This is similar to what I did with `MetricsServlet` and `MetricsFilter`.

Could you port your solution to the current `master` branch? I would really appreciate that.

simpleclient_servlet/src/main/java/io/prometheus/client/filter/MetricsFilter.java
lapo-luchini: Makes sense… I can't find anything definitive [on the 3.0 specs](https://download.oracle.com/otndocs/jcp/servlet-3.0-fr-oth-JSpec/) that only one of the three methods will ever be called, so I'll rather err on the safe side by fixing the race.

PR #690, opened 2021-09-07
Add an InstrumentedAppender constructor that allows passing a CollectorRegistry

MatthewDolan:
I needed the functionality to create an `InstrumentedAppender` for logback which registers the metrics with a custom registry (other than the default registry). The code only supported the assumed default registry. This pull request adds a second public constructor for supplying a registry.

This pull request also abstracts the tests and runs them twice. Once for the default registry and once for a supplied registry.

### DIFF
--- a/simpleclient_logback/src/test/java/io/prometheus/client/logback/InstrumentedAppenderTest.java
+++ b/simpleclient_logback/src/test/java/io/prometheus/client/logback/InstrumentedAppenderTest.java
@@ -12,88 +12,86 @@ import org.junit.Before;
 import org.junit.Test;
 
 public class InstrumentedAppenderTest {
-  public abstract static class Base {
-    CollectorRegistry registry;
-    InstrumentedAppender appender;
-    private ILoggingEvent event;
+  private CollectorRegistry registry;
+  private InstrumentedAppender appender;
+  private InstrumentedAppender defaultAppender;
+  private ILoggingEvent event;
 
-    @Before
-    public void setUp() throws Exception {
-      this.appender.start();
+  @Before
+  public void setUp() throws Exception {
+    registry = new CollectorRegistry();
 
-      this.event = mock(ILoggingEvent.class);
-    }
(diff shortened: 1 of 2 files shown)


simpleclient_logback/src/test/java/io/prometheus/client/logback/InstrumentedAppenderTest.java
```diff
@@ -12,65 +12,88 @@
...
-            new String[]{"level"}, new String[]{level}).intValue();
+  public static class InstanceTest extends Base {
+    @Before
+    public void setUp() throws Exception {
+      this.registry = new CollectorRegistry();
+
+      this.appender = new InstrumentedAppender(this.registry);
+
+      super.setUp();
+    }
```
fstab, line 97: Thanks for the PR, the changes to the production code are looking good.

However, in the test code I think that the inner classes `DefaultTest` and `InstanceTest` are not executed in the Maven build. If I change an `assertEquals()` to make a test fail, the Maven build is still green. In my opinion there's no need for the `DefaultTest`, you could make `InstrumentedAppenderTest` use a custom registry and remove the inner classes.

fstab (changes requested):

simpleclient_logback/src/test/java/io/prometheus/client/logback/InstrumentedAppenderTest.java
MatthewDolan: Thanks for the review. Good catch.

I reverted the change and did something more simple/less fancy. I added two appenders (one default and one for a custom registry) into the test and then just appended to both of them and checked both of them. That should work.

Happy to revert even this if you would prefer the simplicity of testing only with the custom registry. Just wanted to push it up as an alternative option.

Tested locally by changing an `assertEquals(...)` this time.

fstab (approved):
Looks good, thanks a lot.

PR #695, opened 2021-09-15
Added cleaner SSL support to HTTPServer

dhoard:
Added cleaner SSL support to `HTTPServer`.

This will allow downstream projects to set up an `SSLContext` and use it with the `HTTPServer.Builder` via the `withSSLContext(SSLContext sslContext)` method to support SSL.

Added public static helper method `HTTPServer.createSSLContext(String sslContextType, String keyStoreType, String keyStorePath, String keyStorePassword)` to create an `SSLContext` for the most common use-cases.


fstab:
Hi, thanks a lot for the PR, and sorry for the delay. I have this on the top of my TODO list, and I am planning to get back to it this week.

fstab:
Here are a few initial thoughts: First of all, thank you very much for digging into this, SSL is really an ugly part of Java's API.

I think it's worthwhile to step back a bit and look at how SSL should be configured from a user's perspective, for example in `jmx_exporter`.

## Compatibility with node_exporter

My first thought was, let's make SSL configuration similar to `node_exporter`, so that user's don't need to learn different ways how to configure our exporters. In `node_exporter` you have a configuration like this:

```yaml
tls_server_config:
  cert_file: node_exporter.crt
  key_file: node_exporter.key
```

It would be great if there was an easy way to use these `crt` and `key` files in the Java configuration as well. However, I did not find an obvious way how to do this.

I will research this a bit more, because it would be great if we could provide a consistent way of configuring SSL across different exporters.

## Compatibility with Java's default behavior

If we cannot follow the `node_exporter`'s convention, it would be good to follow Java's conventions. My understanding is that by default you can configure Java's SSL with the following system properties:

```
-Djavax.net.ssl.keyStore=/path/to/my/keystore.jks
-Djavax.net.ssl.keyStorePassword=my_secret_password
```
It turns out, if you use the default `HttpsConfigurator` and the default `SSLContext` this seems to work. Replace line `359` with

```java
((HttpsServer) httpServer).setHttpsConfigurator(new HttpsConfigurator(SSLContext.getDefault()));
```

Then create a key store with

```
keytool -genkeypair -keyalg RSA -alias selfsigned -keystore keystore.jks -storepass my_secret_password -dname "CN=localhost, OU=Developers, O=NA, L=Ufa, C=RB"
```

Then start the Server with `-Djavax.net.ssl.keyStore=/path/to/my/keystore.jks -Djavax.net.ssl.keyStorePassword=my_secret_password`.

As this works I'm wondering: Do we need the custom `SSLContext` and `HttpsConfigurator`? I don't have an opinion yet, I'm just trying to understand the implications.

fstab:
Hi @dhoard, thanks for the update.

Regarding security: I don't think that system properties are necessarily insecure, because if you want to avoid command line parameters you can set them programmatically as well with `System.setProperty()`.

However, it's true that if SSL is configured via system properties, the configuration will apply to the rest of the application as well. Maybe we should offer both ways, a simple `withSSL()` that uses the default system properties, and an advanced `withHttpsConfigurator(HttpsConfigurator)` for custom configuration.

I'm wondering how much of `createHttpsConfigurator()` and `createSSLContext()` should actually be part of `client_java`. I tried to create an `HttpsConfigurator` without overriding `configure()` linke this:

```java
new HttpsConfigurator(sslContext);
```

It works (at least the test is green when I do this), so I'm wondering if overriding `configure()` is really needed.

Regarding the `createSSLContext()` method, I find it hard to decide which settings should be configurable via parameters, and which settings should be hard-coded. For example, it is likely that `sslContextType` is always `SSL` for most users, however, it is configurable. On the other hand the algorithm name might not be `SunX509` for some users (I found a tutorial with `IbmISeriesX509`), however it is hard-coded.

I'm not sure whether there is a good solution for an SSL configuration API. I'm wondering if we should remove `createHttpsConfigurator()` and `createSSLContext()`, and just provide a `withHttpsConfigurator(HttpsConfigurator)` method, and let the users create the `HttpsConfigurator` themselves. This would avoid mixing SSL configuration with the `client_java` API (I guess there are many special cases where whatever API we provide does not fit), and it would give users full flexibility.

What do you think?

dhoard:
@fstab all good points. 

I think we should add the `withHttpsConfigurator()` method and remove the `withSSL()` from the `HTTPServer.Builder()` class. 

Then, I think we should removed the `createSSLContext()` method from the `HTTPServer.Builder()` class and introduce a new class `KeyStoreSSLContext` to encapsulate the logic to use a keystore. 

This should allow creating a 'PEMCertificateSSLContext` so that a `client_java` use can use either a keystore or PEM certificate.

fstab:
The question is how much value our own `KeyStoreSSLContext` would provide over the built-in `SSLContext`. If there are no good reason for introducing our own abstraction we could just work with `SSLContext` directly.

fstab:
I feel we should stick with the built-in `HttpsConfigurator` for now and leave the rest to the user. If we implement our own API, we will either end up with something that's too simplistic for many users, or with something that's as complicated as `HttpsConfigurator` itself. If it turns out that many users configure `HttpsConfigurator` in the same way, we can still add an API later based on how users are actually using it.

dhoard:
@fstab changes made per our converstation. Please review.

dhoard:
@fstab Thoughts on merging this PR?

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -200,6 +202,7 @@ static ThreadFactory defaultThreadFactory(boolean daemon) {
         private Predicate<String> sampleNameFilter;
         private Supplier<Predicate<String>> sampleNameFilterSupplier;
         private Authenticator authenticator;
+        private HttpsConfigurator configurator;
```
fstab, line 205: Nitpick: Rename to `httpsConfigurator` so that it's clear what this is about.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -308,11 +319,13 @@ public HTTPServer build() throws IOException {
                 assertNull(sampleNameFilterSupplier, "cannot configure 'sampleNameFilter' and 'sampleNameFilterSupplier' at the same time");
                 sampleNameFilterSupplier = SampleNameFilterSupplier.of(sampleNameFilter);
             }
+
             if (httpServer != null) {
                 assertZero(port, "cannot configure 'httpServer' and 'port' at the same time");
                 assertNull(hostname, "cannot configure 'httpServer' and 'hostname' at the same time");
                 assertNull(inetAddress, "cannot configure 'httpServer' and 'inetAddress' at the same time");
                 assertNull(inetSocketAddress, "cannot configure 'httpServer' and 'inetSocketAddress' at the same time");
+                assertNull(configurator, "cannot configure 'httpServer' and 'configurator' at the same time");
```
fstab, line 328: Nitpick: Replace `'configurator'` with `'httpsConfigurator'` in the error message so that it's clear that this is related to the `withHttpsConfigurator()` call.

fstab (changes requested):
Thanks a lot and sorry for the delay. It's looking good, I just commented two nitpicks.
Please also rebase it to the current `master`.
I'm happy to merge and release it so we can use it in `jmx_exporter`.

Thanks you for your patience!

fstab (approved):

fstab:
Thanks a lot! Would you mind rebasing the branch to the current master? Automatic rebase fails because of merge conflicts. This branch has 10 commits, and I think for later reference it would be easier to squash them into one and rebase them on the current master.

BTW I just published a release so that people don't have the vulnerable log4j dependency in their dependency anymore. However, that does not mean that we need to wait for long until we create a new release with the SSL support. I'd say as soon as we want to merge code in `jmx_exporter` that needs the SSL support, we create a new `client_java` release and use it in `jmx_exporter`.

fstab:
Thanks a lot!

fstab:
It's released with [0.14.0](https://github.com/prometheus/client_java/releases/tag/parent-0.14.0) :tada:.

PR #698, opened 2021-09-16
Added Base64 implementation to simpleclient_common to remove jax-api …

dhoard:
Added Base64 implementation to simpleclient_common to remove jax-api dependency. Abstracted Base64 implementation to be more Java 8 like/make it easier to replace if required.

### DIFF
--- a/simpleclient_pushgateway/pom.xml
+++ b/simpleclient_pushgateway/pom.xml
@@ -48,6 +48,7 @@
             <groupId>javax.xml.bind</groupId>
             <artifactId>jaxb-api</artifactId>
             <version>2.3.0</version>
+            <scope>provided</scope>
         </dependency>
         <!-- Test Dependencies Follow -->
         <dependency>
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/Base64.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/Base64.java
@@ -1,0 +1,48 @@
+package io.prometheus.client.exporter;
+
+import javax.xml.bind.DatatypeConverter;
+
+/**
+ * This class delegates to either javax.xml.bind.DatatypeConverter (for Java < 8) or java.util.Base64 (Java 8+)
+ * to perform Base64 encoding of a String.
+ *
+ * This code requires Java 8+ for compilation.
+ */
+public class Base64 {
+
+    private static final boolean HAS_JAVA_UTIL_BASE64 = hasJavaUtilBase64();
+
+    private static boolean hasJavaUtilBase64() {
+        try {
+            Class.forName("java.util.Base64");
+            // Java 8+
+            return true;
+        } catch (Throwable t) {
+            // Java < 8
+            return false;
+        }
+    }
+
+    private Base64() {}
+
+    /**
+     * Encodes a byte[] to a String using Base64.
+     *
+     * Passing a null argument will cause a NullPointerException to be thrown.
+     *
+     * @param src
+     * @return String in Base64 encoding
+     */
+    @SuppressWarnings("all")
+    public static String encodeToString(byte[] src) {
+        if (src == null) {
+            throw new NullPointerException();
+        }
+
+        if (HAS_JAVA_UTIL_BASE64) {
+            return java.util.Base64.getEncoder().encodeToString(src);
+        } else {
+            return DatatypeConverter.printBase64Binary(src);
+        }
+    }
+}
--- a/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/BasicAuthHttpConnectionFactory.java
+++ b/simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/BasicAuthHttpConnectionFactory.java
@@ -3,7 +3,6 @@ package io.prometheus.client.exporter;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
-import javax.xml.bind.DatatypeConverter;
 
 public class BasicAuthHttpConnectionFactory implements HttpConnectionFactory {
     private final HttpConnectionFactory originConnectionFactory;
@@ -28,7 +27,7 @@ public class BasicAuthHttpConnectionFactory implements HttpConnectionFactory {
     private String encode(String user, String password) {
         try {
             byte[] credentialsBytes = (user + ":" + password).getBytes("UTF-8");
-            String encoded = DatatypeConverter.printBase64Binary(credentialsBytes);
+            String encoded = Base64.encodeToString(credentialsBytes);
             return String.format("Basic %s", encoded);
         } catch (UnsupportedEncodingException e) {
             throw new IllegalArgumentException(e);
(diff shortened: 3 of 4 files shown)


dhoard:
@fstab please review.

fstab (changes requested):
The code is looking good, however, I would like to move it out of the `simpleclient_common` module. That module is used by a lot of other projects and I would like to keep it's scope as small as possible.

As far as I can see the only usage in production code is in `simpleclient_pushgateway`. Could you move it there?

The other usage is in a test in `simpleclient_httpserver`, but I guess it does not harm to keep `javax.xml.bind` as a test dependency in that module.

dhoard:
@fstab Thanks for the review! My thoughts around putting it into `simpleclient_common`...

1. I am hoping we can get HTTP authentication in `jmx_exporter` soon, which would require a `Base64` implementation for Java 6/7 compatibility.

2. Other users of `client_java` may also want to implement HTTP authentication when using `HTTPServer` and therefore also need it for Java 6/7 compatibility.

By putting into `simpleclient_pushgateway` we essential push the issue to downstream projects to solve the `Base64` issue to maintain Java 6/7 compatibility when using `simpleclient_httpserver`. A common implementation seems like it could be beneficial.

I'm concerned that if/when we add HTTP authentication to `jmx_exporter` this will be deemed something that needs to be addressed by `client_java/simple_httpserver` which means either keeping two versions of the `Base64` code in `client_java`... one in `simpleclient_pushgateway` and one in `simpleclient_httpserver` or refactoring back to a common implementation in `simpleclient_common`.

Thoughts?

wilkinsona:
Instead of maintaining an implementation of Base64 encoding, I wonder if it would be better to use `java.util.Base64` when it's available and fall back to `DatatypeConverter` when it is not?

dhoard:
@wilkinsona I thought of that approach, but not sure it would solve the issue. The goal is not to have a dependency on JAX APIs whatsoever.

For Java 8+, it would be easy, a `Base64` class to check for/use `java.util.Base64` using delegation via reflection (Not sure we want to introduce reflection in the process.)

For Java 6/7, there is nothing to fall back to unless we include a `Base64` implementation.

Am I missing something?

wilkinsona:
`DatatypeConverter` is part of the JDK in [Java 6](https://docs.oracle.com/javase/6/docs/api/javax/xml/bind/DatatypeConverter.html) (and 7 and 8). It moved into a separate module in Java 9 and was removed entirely in Java 11. For Java 6 and 7, you can delegate to `DatatypeConverter`. For Java 8+, you can delegate to `java.util.Base64`. As long as you can compile against Java 8 while producing Java 6-compatible byte code, you can do so without reflection.

dhoard:
Good points. I'll see what I do when I get some time.

dhoard:
@wilkinsona I have implemented the code as you described. Works as expected.

@fstab I still feel this should be in `client_java/simpleclient_common` since it will be a benefit to both `client_java` and any projects dependent on `client_java`. (I have moved it back in the latest PR for review)

simpleclient_common/pom.xml
```diff
@@ -45,5 +45,11 @@
             <version>4.13.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.3.0</version>
+            <scope>compile</scope>
```
wilkinsona, line 52: I suspect this is only needed for compilation with Java 11 and later. Perhaps it should be marked as optional or could even be removed entirely?

dhoard: Java 11 is required for compilation, so scope = `compile` is required.

wilkinsona: In that case, I think it should be marked as optional or provided as it shouldn't be needed at runtime.

fstab (changes requested):
Thanks a lot for the discussion! I thought about this, and I am hesitant to add a dependency or a Base64 implementation to `simpleclient_common`.

`simpleclient_common` is used way beyond the `client_java` ecosystem: It's in Micrometer, which ships with Spring Boot, Quarkus, and others. And it's also part of the Java implementation of the upcoming OpenTelemetry metrics standard.

That means lots of applications out there have `simpleclient_common` as an indirect dependency, even if they aren't aware and don't use the `client_java` library. We should be very conservative with what we add to that module. What if somebody implements an incompatible jaxb-api version 3 in the future?

> By putting into `simpleclient_pushgateway` we essential push the issue to downstream projects to solve the `Base64` issue to maintain Java 6/7 compatibility when using `simpleclient_httpserver`.

This is correct, but as 3rd party projects using `client_java` are unlikely to support Java 6 they can just use `java.util.Base64` and won't have an issue.

The main project where this is relevant is the `jmx_exporter`, but given how widely `simpleclient_common` is used I'd rather push the issue downstream to `jmx_exporter` and risk a bit of copy-and-paste between `simpleclient_pushgateway` and the `jmx_exporter` than putting a Base64 implementation in `simpleclient_common`.

dhoard:
@fstab @wilkinsona changes made. please review.

simpleclient_pushgateway/pom.xml
```diff
@@ -62,5 +57,11 @@
            <version>3.9.17</version>
            <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.3.0</version>
+            <scope>provided</scope>
+        </dependency>
```
fstab, line 65: Nitpick: This is not a test dependency, so it would be good to move it up above the comment `<!-- Test Dependencies Follow -->`.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/Base64.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * This class delegates to either javax.xml.bind.DatatypeConverter (for Java < 8) or java.util.Base64 (Java 8+)
+ * to perform Base64 encoding of a String.
+ *
+ * This code requires Java 8+ for compilation.
+ */
+public class Base64 {
+
+    private static boolean hasJavaUtilBase64;
```
fstab, line 13: Nitpick: As this is supposed to never change, it should be `final` for clarity.

fstab (changes requested):
Thanks a lot! I added two more nitpicks, apart from that it's ready to be rebased, squashed and merged :tada:.

simpleclient_pushgateway/pom.xml
dhoard: I'll make the change.

simpleclient_pushgateway/src/main/java/io/prometheus/client/exporter/Base64.java
dhoard: Because `hasJavaUtilBase64` is initialized in a static block, it can't be final.

wilkinsona: It can be if you do something like this instead:

```java
private static final boolean HAS_JAVA_UTIL_BASE64 = base64IsPresent();

private static boolean base64IsPresent() {
    try {
        Class.forName("java.util.Base64");
        return true;
    }
    catch (Throwable ex) {
        return false;
    }
}
```

fstab (approved):
Thanks a lot! I can confirm that it works with Zulu's Java 6 as well as OpenJDK 11 and 17.

Issue #703, opened 2021-09-26
Use of ThreadLocal storage in HTTPServer seems resulting in inefficient memory allocation


gvisco:
### Disclaimer
I might be missing something in the design of the class (I'm sorry if this is the case), and yet I think it's worth checking and possibly suggest alternative implementations.

### TL;DR
The class HTTPServer in `client_java/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java` seems to allocate more memory than what is actually needed, and for too long. 

### Description
The class relies, to handle the HTTP requests, on a ByteArrayOutputStream in ThreadLocal storage, that is:
- as big as the serialized reply, and allocated 5 times, once per thread in the pool
- never re-used, since it is overwritten for each incoming request
- never de-allocated or cleaned up after the reply has been delivered

Basically it looks like if the class is implemented to avoid the instantiation of new ByteArrayOutputStream objects and, to achieve this, it pays the maximum cost in terms of memory allocation. 

### Code walkthrough
The following code is from the HTTPServer class, as it is currently on the `master` branch, version `0.12.1-SNAPSHOT`.
The class has a member `response` of type LocalByteArray, which is a ThreadLocal ByteArrayOutputStream:

```java
private static class LocalByteArray extends ThreadLocal<ByteArrayOutputStream> {
    @Override
    protected ByteArrayOutputStream initialValue()
    {
        return new ByteArrayOutputStream(1 << 20);
    }
}
```

This is how the class handles the incoming HTTP requests, I've added the comments. 

```java
public void handle(HttpExchange t) throws IOException {
    String query = t.getRequestURI().getRawQuery();
    String contextPath = t.getHttpContext().getPath();
    // take, from the thread local storage, the output stream containing the previous response
    // this is the only reference in the code to 'this.response`
    ByteArrayOutputStream response = this.response.get();
    // throw away the previous response
    response.reset();
    // wrap the output stream with a writer
    OutputStreamWriter osw = new OutputStreamWriter(response, Charset.forName("UTF-8"));
    if ("/-/healthy".equals(contextPath)) {
        osw.write(HEALTHY_RESPONSE);
    } else {
        String contentType = TextFormat.chooseContentType(t.getRequestHeaders().getFirst("Accept"));
        t.getResponseHeaders().set("Content-Type", contentType);
        Predicate<String> filter = sampleNameFilterSupplier == null ? null : sampleNameFilterSupplier.get();
[...]

PR #704, opened 2021-09-30
Refactored micro-optimization. Converted code to use a ThreadPoolExec…

dhoard:
Refactored micro-optimization. Converted code to use a ThreadPoolExecutor.

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -18,11 +18,14 @@ import java.nio.charset.Charset;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.zip.GZIPOutputStream;
 
@@ -54,20 +57,11 @@ public class HTTPServer implements Closeable {
         }
     }
 
-    private static class LocalByteArray extends ThreadLocal<ByteArrayOutputStream> {
-        @Override
-        protected ByteArrayOutputStream initialValue()
-        {
-            return new ByteArrayOutputStream(1 << 20);
-        }
-    }
-
     /**
      * Handles Metrics collections from the given registry.
      */
     public static class HTTPMetricHandler implements HttpHandler {
         private final CollectorRegistry registry;
-        private final LocalByteArray response = new LocalByteArray();
         private final Supplier<Predicate<String>> sampleNameFilterSupplier;
         private final static String HEALTHY_RESPONSE = "Exporter is Healthy.";
 
@@ -84,8 +78,7 @@ public class HTTPServer implements Closeable {
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
             String contextPath = t.getHttpContext().getPath();
-            ByteArrayOutputStream response = this.response.get();
-            response.reset();
+            ByteArrayOutputStream response = new ByteArrayOutputStream(1 << 20);
             OutputStreamWriter osw = new OutputStreamWriter(response, Charset.forName("UTF-8"));
             if ("/-/healthy".equals(contextPath)) {
                 osw.write(HEALTHY_RESPONSE);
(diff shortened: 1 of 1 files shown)


dhoard:
@fstab I don't _think_ this is related to my changes. If fails the same way when I build (`mvn clean install`) using `prometheus/client_java` `master`.

simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
```diff
@@ -84,8 +78,7 @@ protected ByteArrayOutputStream initialValue()
         public void handle(HttpExchange t) throws IOException {
             String query = t.getRequestURI().getRawQuery();
             String contextPath = t.getHttpContext().getPath();
-            ByteArrayOutputStream response = this.response.get();
-            response.reset();
+            ByteArrayOutputStream response = new ByteArrayOutputStream(1 << 20);
```
gvisco, line 81: If I get correctly this comment from @fstab (https://github.com/prometheus/client_java/issues/703#issuecomment-930470387)

>If the initial 1M capacity is not enough, the LocalByteArray will increase in size and copy the data to a larger array. Keeping it will have the effect that this happens only once, because after that it remains in memory with the new capacity.

the previous implementation had the advantage to avoid enlarging the array for each request if the data is larger than 1M. 

If I may suggest a possible approach to keep this feature, we could use an `AtomicInteger` to keep track of the maximum size of the serialized requests so far, and initialize the `ByteArrayOutputStream` immediately to the required size (maybe adding an extra 1K, to avoid expanding the array if the array grows of few bytes). 
This would also save us some byte if the request is smaller than 1M.

dhoard: `LocalByteArray` (i.e. the `ByteArrayOutputStream`) will always grow to the maximum number of bytes that have been written using the object. Being `ThreadLocal` vs method local doesn't change the behavior.

gvisco: I'm sorry @dhoard , my comment was not clear (or I might have misinterpreted @fstab comment). Let me try to explain better my point.

The `ByteArrayOutputStream` is initialized to have, internally, a byte array with initial capacity of 1M. If the response written in the `ByteArrayOutputStream` is larger than the capacity, then the underlying byte array is automatically enlarged to fit the data (typically doubling the size, from 1M to 2M). This enlargement does not come completely for free, since a new array needs to be allocated and the data must be copied from one array to the other.

The previous implementation reduced the number of times that the enlargement was performed, by keeping in memory, and reusing, the instance of `ByteArrayOutputStream` that was already been enlarged.
For example, if you had to write responses of 1.5M, each instance of `ByteArrayOutputStream` was enlarged only once for the entire life of the application.

After the proposed change, instead, every time a request is processed a new `ByteArrayOutputStream` is created with initial capacity equals to 1M. If the serialized response is bigger than the 1M capacity, therefore, the enlargement need to be performed for each request.

Assuming that this additional processing may be a problem - but I would leave to @fstab to comment on whether this is important or not - I was suggesting a way to avoid the array enlargement to happen too frequently, which is to keep track of the required capacity and use this information to initialize the newly created `ByteArrayOutputStream`.

dhoard:
This code has issues...

- doesn't handle `NamedDaemonThreadFactory` correctly... the thread number will always increase on every thread creation
- will incur a processing lag when increasing the thread pool from 1. (i.e. block the calling client)

I will have to look at this some more...

dhoard:
After looking at the code and the above issues... closing this PR since it doesn't provide any real benefits.

dhoard:
@gvisco all analysis is valuable.

Maybe it's worthwhile to make the code configurable so that the end-user could choose whether to use an instance variable vs thread-local `ByteArrayOutputStream`.

Any changes in `client_java` (regardless of approach) would require a new version(s) of exporters that use `client_java`. (`jmx_exporter` is most likely the most common use-case.)

I feel we should just leave the fixed thread pool in place as is. I don't see any value in implementing code that is the same as `Executors.newFixedThreadPool()`

fstab:
Hi, thanks a lot everyone for the valuable discussion. If you look into the implementation of `FixedThreadPoolExecutor` it creates a `ThreadPoolExecutor` like this:
```java
ThreadFactory threadFactory = NamedDaemonThreadFactory.defaultThreadFactory(daemon);
executorService = new ThreadPoolExecutor(5, 5,
              0L, TimeUnit.MILLISECONDS,
              new LinkedBlockingQueue<Runnable>(),
              threadFactory);
```
What about replacing it with this (the only difference is that the initial number of threads is `1` instead of `5`)?
```java
ThreadFactory threadFactory = NamedDaemonThreadFactory.defaultThreadFactory(daemon);
executorService = new ThreadPoolExecutor(1, 5,
              0L, TimeUnit.MILLISECONDS,
              new LinkedBlockingQueue<Runnable>(),
              threadFactory);
```
As one thread is enough for most users, this would prevent the unnecessary creation of the other 4 threads.

PR #710, opened 2021-10-18
Fix the situation where metric collection stops when NPE occours

codings-dan:
We use `Prometheus` to monitor `Alluxio` system metrics. When some metrics are empty for various reasons, the system no longer monitors the metrics. We can catch this exception, so as not to affect the collection and monitoring of other metrics.

### DIFF
--- a/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
+++ b/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java
@@ -9,6 +9,7 @@ import com.codahale.metrics.MetricFilter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.Timer;
+import io.prometheus.client.Predicate;
 import io.prometheus.client.dropwizard.samplebuilder.SampleBuilder;
 import io.prometheus.client.dropwizard.samplebuilder.DefaultSampleBuilder;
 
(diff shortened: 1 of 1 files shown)


codings-dan:
I think I probably understood your idea, but there may be a difference. We can't rewrite `collect(Predicate<String> sampleNameFilter)`, because the situation of reporting a null pointer exception requires a  method similar to `collect(Predicate<T extends Metric> sampleNameFilter)` , this may need to change the code of `Collector.class`. Or, did I not understand your suggestion
Looking forward for your response, thx!

fstab:
I was thinking you add a method like this to `DropwizardExports`:

```java
@Override
public List<MetricFamilySamples> collect(Predicate<String> nameFilter) {
    // TODO: Collect only metrics where nameFilter.test(name) is true
}
```

With such a method, `DropwizardExports` will be capable of filtering metrics by name before they are collected. Then you could use that to exclude the erroneous `Alluxio` metrics.

dhoard:
@codings-dan I feel a combination of @fstab 's proposals would be ideal.

1. Implement the `SampleNameFilter`...

```
@Override
public List<MetricFamilySamples> collect(Predicate<String> nameFilter) {
    // TODO: Collect only metrics where nameFilter.test(name) is true
}
```

2. catch/throw an unchecked `RuntimeException` when performing the `collect`...

Note: Using a `RuntimeException` since the implementation of the `Metric` could potentially be performing some work, etc... (really depends on the implementation of the `Metric`, which we don't know from this code's point of view.)

-----

Possible/example code (Not tested)

```
import static io.prometheus.client.SampleNameFilter.ALLOW_ALL;
```

```
    @Override
        return collect(null);
    }

    @Override
    public List<MetricFamilySamples> collect(Predicate<String> nameFilter) {
        nameFilter = nameFilter == null ? ALLOW_ALL : nameFilter;
        Map<String, MetricFamilySamples> mfSamplesMap = new HashMap<String, MetricFamilySamples>();
        String type = null;
        String name = null;
        try {
            type = "Gauge";
            for (SortedMap.Entry<String, Gauge> entry : registry.getGauges(metricFilter).entrySet()) {
                name = entry.getKey();
                if (nameFilter.test(name)) {
                    addToMap(mfSamplesMap, fromGauge(name, entry.getValue()));
                }
            }
            type = "Counter";
            for (SortedMap.Entry<String, Counter> entry : registry.getCounters(metricFilter).entrySet()) {
                name = entry.getKey();
                if (nameFilter.test(name)) {
                    addToMap(mfSamplesMap, fromCounter(name, entry.getValue()));
                }
            }
            type = "Histogram";
            for (SortedMap.Entry<String, Histogram> entry : registry.getHistograms(metricFilter).entrySet()) {
                name = entry.getKey();
                if (nameFilter.test(name)) {
                    addToMap(mfSamplesMap, fromHistogram(name, entry.getValue()));
                }
            }
            type = "Timer";
            for (SortedMap.Entry<String, Timer> entry : registry.getTimers(metricFilter).entrySet()) {
                name = entry.getKey();
                if (nameFilter.test(name)) {
                    addToMap(mfSamplesMap, fromTimer(name, entry.getValue()));
                }
            }
            type = "Meter";
            for (SortedMap.Entry<String, Meter> entry : registry.getMeters(metricFilter).entrySet()) {
                name = entry.getKey();
                if (nameFilter.test(name)) {
                    addToMap(mfSamplesMap, fromMeter(name, entry.getValue()));
                }
            }
        } catch (Exception e) {
            throw new RuntimeException("Exception processing " + type + " " + name, e);
        }

        return new ArrayList<MetricFamilySamples>(mfSamplesMap.values());
    }
 ```

codings-dan:
@dhoard Thanks for reviewing the code and making suggestions, I'll try to fix this again based on your suggestion

Issue #607, opened 2020-11-28
Split package in simpleclient.servlet and simpleclient.pushgateway


Alexey1Gavrilov:
Modules `simpleclient.servlet` and `simpleclient.pushgateway` are exporting classes from the same package `io.prometheus.client.exporter` causing troubles when building Java modular applications due to a split package error:

`java: the unnamed module reads package io.prometheus.client.exporter from both simpleclient.servlet and simpleclient.pushgateway`

To reproduce the problem create a `module-info.java` like this:

```
module test {
  requires simpleclient.pushgateway;
  requires simpleclient.servlet;
}
```

brian-brazil:
I'm not familiar with this, can you explain in exactly what systems this is a problem?

Renaming these would likely cause widespread breakage to existing code, so a split is not likely.

Alexey1Gavrilov:
Basically you cannot build a modular Java application having a module which imports classes from both `simpleclient.pushgateway` and `simpleclient.servlet` modules

Here a repo reproducing the problem:  https://github.com/Alexey1Gavrilov/java-module-issues/tree/master/promethues
The compilation fails with the following error:

```
> mvn compile
...
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] the unnamed module reads package io.prometheus.client.exporter from both simpleclient.servlet and simpleclient.pushgateway
[ERROR] module simpleclient.pushgateway reads package io.prometheus.client.exporter from both simpleclient.servlet and simpleclient.pushgateway
[ERROR] module simpleclient.servlet reads package io.prometheus.client.exporter from both simpleclient.pushgateway and simpleclient.servlet
[ERROR] module simpleclient reads package io.prometheus.client.exporter from both simpleclient.servlet and simpleclient.pushgateway
[ERROR] /Users/agavrilov/Work/git/java-module-issues/promethues/src/main/java/module-info.java:[1,1] module javamodule.promethues reads package io.prometheus.client.exporter from both simpleclient.servlet and simpleclient.pushgateway
[INFO] 5 errors
```

A possible workaround for a modular application is to avoid importing from both in the one application module, ie have a two separate submodules requiring one of `simpleclient.pushgateway` or `simpleclient.servlet`.

Generally speaking splitting packages across .jar files should be avoided.
[This SO question](https://stackoverflow.com/a/51828900/3537858) is a good source of info.  
I think it makes sense to rename the package in a major version update. Should not cause much troubles to the users.

[...]

PR #712, opened 2021-10-29
Add automatic module names for compatibility with Jigsaw

apmasell:
This adds entries to the JAR manifest files to allow consumption of Prometheus
by the Java module system (JPMS aka Jigsaw) introduced in Java 9. Since these
are added only to the manifests, rather than the `module-info.class`, they will
not change behaviour for JVMs that do not use the module system. Some modules
are not compliant with the requirement that a package be present in only one
module, so they are excluded.

### DIFF
--- a/simpleclient_servlet/pom.xml
+++ b/simpleclient_servlet/pom.xml
@@ -84,5 +84,5 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-
+    <!--This cannot have an automatic module name as its packages conflicts with the simpleclient module-->
 </project>
--- a/simpleclient/pom.xml
+++ b/simpleclient/pom.xml
@@ -55,4 +55,19 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>io.prometheus.client</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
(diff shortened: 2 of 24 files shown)


simpleclient_servlet/pom.xml
```diff
@@ -84,5 +84,5 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-
+    <!--This cannot have an automatic module name as its packages conflicts with the simpleclient module-->
```
fstab, line 87: I guess this is directly related to my question here: https://github.com/prometheus/client_java/issues/607#issuecomment-952310842.

As I have never used Jigsaw, could you please explain why we cannot expose the packages `io.prometheus.client.exporter` and `io.prometheus.client.filter` here so that users can use the `MetricsServlet` and the `MetricsFilter`? Sorry if this is a stupid question, I have no experience with Java modules.

fstab (commented):
Thanks a lot! I added a question related to https://github.com/prometheus/client_java/issues/607#issuecomment-952310842.

simpleclient_servlet/pom.xml
apmasell: Yeah, so it's a bit of a backwards compatibility nightmare. There are two kinds of modules: one with `module-info.class` and automatic ones.

If you create your own `module-info` then you can control what you export, but you can't create classes in any packages from modules that you import, so your `Adapter` class would fail to compile.

For automatic modules, because they're backwards compatible, you have to export all the classes you have and only one module can export a package.

Oracle claimed it was always best practice so have each package provided by only one JAR, but that was never taken seriously. Now with the new module system, it's law. The logic is that Java's "friend" permission-level meant that only classes in the same package could access a certain class or method or field or whatever. Since you could always just create a class in the same package, it had no real security enforcement. This restriction is meant to make friend permissions be useful.

apmasell: Upon closer inspection: 
`simpleclient_httpserver` and `simpleclient_pushgateway` both have classes in the `io.prometheus.client.exporter`. That means no module can import both of them at the same time. I can't imagine a use case where both are necessary, so it can be ignored for now, but, if these packages are going to be refactored in the future, they should be renamed.

PR #726, opened 2021-12-12
log4j version upgrade for CVE-2021-44228

### LINKED ISSUE #725: log4j version upgrade for CVE-2021-44228
Based on : https://github.com/advisories/GHSA-jfh8-c2jp-5v3q and https://www.lunasec.io/docs/blog/log4j-zero-day/
Log4j versions prior to 2.15.0 are subject to a remote code execution vulnerability via the ldap JNDI parser.
Current version: [2.1 ](https://github.com/prometheus/client_java/blob/master/simpleclient_log4j2/pom.xml#L44)

yakirgb:
Fixes #725
Upgrade log4j-core from 2.1 to 2.15.0

### DIFF
--- a/simpleclient_log4j2/pom.xml
+++ b/simpleclient_log4j2/pom.xml
@@ -41,7 +41,7 @@
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-core</artifactId>
-            <version>2.1</version>
+            <version>2.15.0</version>
         </dependency>
 
         <dependency>


fstab (approved):
Thanks, it's merged.

However: `simpleclient_log4j2` gets metrics out of log4j2, but it does not log with log4j2. So `simpleclient_log4j2` is not affected by the CVE.

Anyway, as people will likely scan their dependencies for old log4j2 versions, I will release this soon so that `simpleclient_log4j2` does not pop up as a false alert.

desjardd1:
Hello is there a place to see if 2.16 has been deployed. 2.15 is still vulnerable (critical). Thanks

fstab:
Hi, I just updated to 2.16. It will be in the next release, which will come soon as we also want to release the SSL support for the `HTTPServer`.

That being said: With rel 0.13.0 we marked the `log4j` dependency in `simpleclient_log4j2` as `provided`. That means `simpleclient_log4j2` does not ship with `log4j`. Instead, it will use whatever `log4j` version is provided by the application being monitored.

If you monitor an application that still uses `log4j` 2.14.1 you will be vulnerable even if you use the current `simpleclient_log4j2` for monitoring. If you monitor an application that is up-to-date with `log4j` 2.16.0 `simpleclient_log4j2` will use 2.16.0 even though the current release was built with 2.15.0.

fstab:
I just released [0.14.0](https://github.com/prometheus/client_java/releases/tag/parent-0.14.0) with a `log4j` update to 2.16.0.

As said above, at runtime `simpleclient_log4j2` uses the `log4j` version that ships with the monitored application, so it is more important to make sure that the application you want to monitor ships with an up-to-date `log4j`. However, the dependency triggers some security scanners so I released the update.

Issue #732, opened 2021-12-15
Both 'Content-Length: 0' and 'Transfer-Encoding: chunked' found when 'Accept-encoding: gzip' header not sent


ShaharTrasmitSecurity:
According to the http 1.1 spec:

> A sender MUST NOT send a Content-Length header field in any message
> that contains a Transfer-Encoding header field. (https://tools.ietf.org/html/rfc7230#section-3.3.2).

When testing prometheus metrics exposition, we are creating HTTPServer (in `io.prometheus.client.exporter.HTTPServer`) to expose metrics to, and sending HTTP requests to query the exposed metrics

When no metrics matches the query and an empty response should return, the following exception occur:

```
java.lang.IllegalArgumentException: Both 'Content-Length: 0' and 'Transfer-Encoding: chunked' found
```

After debugging, i found that when gzip encoding is not allowed (through `Accept-encoding` header), the response will always be encoded as chunked, and will contain the Content-Length: 0 header which conflicts with the HTTP 1.1 spec

** The issue occur only when the content length is 0

dhoard:
@ShaharTrasmitSecurity I suspect it's in `HTTPMetricHandler` in this code...

```
            } else {
                String contentType = TextFormat.chooseContentType(t.getRequestHeaders().getFirst("Accept"));
                t.getResponseHeaders().set("Content-Type", contentType);
                Predicate<String> filter = sampleNameFilterSupplier == null ? null : sampleNameFilterSupplier.get();
                filter = SampleNameFilter.restrictToNamesEqualTo(filter, parseQuery(query));
                if (filter == null) {
                    TextFormat.writeFormat(contentType, osw, registry.metricFamilySamples());
                } else {
                    TextFormat.writeFormat(contentType, osw, registry.filteredMetricFamilySamples(filter));
                }
            }
```

A JUnit test will need to be written in `TestHTTPServer` to confirm.

dhoard:
@ShaharTrasmitSecurity @fstab I have an initial branch of code https://github.com/dhoard/client_java/tree/issue_732 that appears to resolve the issue, but requires refactoring of `TestHttpServer` to be able to get HTTP response headers.

Implementing https://github.com/prometheus/client_java/issues/700 would make this type of test easier.

dhoard:
@ShaharTrasmitSecurity PR https://github.com/prometheus/client_java/pull/738 has been merged to resolve the issue in the next release.

PR #738, opened 2021-12-21
Changes to HTTPMetricHandler to fix Issue #732

dhoard:
Changes to HTTPServer and HTTPMetricHandler to resolve getting Content-Length=0 when Transfer-Encoding=chunked. Refactored TestHTTPTest for cleaner testing.

### DIFF
--- a/simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/HttpResponse.java
+++ b/simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/HttpResponse.java
@@ -1,0 +1,109 @@
+package io.prometheus.client.exporter;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Class to perform HTTP testing
+ */
+class HttpResponse {
+
+    private int responseCode;
+    private Map<String, List<String>> headers;
+    private long contentLength = -1;
+    private String body;
+
+    /**
+     * Constructor
+     *
+     * @param responseCode
+     * @param headers
+     * @param contentLength
+     * @param body
+     */
+    HttpResponse(int responseCode, Map<String, List<String>> headers, long contentLength, String body) throws IOException {
+        this.responseCode = responseCode;
+        this.body = body;
+        this.contentLength = contentLength;
+        this.headers = new HashMap<String, List<String>>();
+
+        Set<Map.Entry<String, List<String>>> headerSet = headers.entrySet();
+        for (String header : headers.keySet()) {
+            if (header != null) {
+                List<String> values = headers.get(header);
+                this.headers.put(header.toLowerCase(), values);
+            }
+        }
+
+        if (getHeader("content-length") != null && getHeader("transfer-encoding") != null) {
+            throw new IOException("Invalid HTTP response, should only contain Connect-Length or Transfer-Encoding");
+        }
+    }
+
+    /**
+     * Method to get the HTTP response code
+     *
+     * @return int
+     */
+    public int getResponseCode() {
+        return this.responseCode;
+    }
+
+    /**
+     * Method to get a list of HTTP response headers values
+     *
+     * @param name
+     * @return List<String>
+     */
+    public List<String> getHeaderList(String name) {
+        return headers.get(name.toLowerCase());
+    }
+
+    /**
+     * Method to get the first HTTP response header value
+     *
+     * @param name
+     * @return String
+     */
+    public String getHeader(String name) {
+        String value = null;
+
+        List<String> valueList = getHeaderList(name);
+        if (valueList != null && (valueList.size() >= 0)) {
+            value = valueList.get(0);
+        }
+
+        return value;
+    }
+
+    /**
+     * Method to get the first HTTP response header value as a Long.
+     * Returns null of the header doesn't exist
+     *
+     * @param name
+     * @return Long
+     */
+    public Long getHeaderAsLong(String name) {
+        String value = getHeader(name);
+        if (value != null) {
+            try {
+                return Long.valueOf(value);
+            } catch (Exception e) {
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * Method to get the HTTP response body
+     *
+     * @return String
+     */
+    public String getBody() {
+        return body;
+    }
+}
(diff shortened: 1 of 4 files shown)


simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
```diff
@@ -115,316 +93,412 @@ public void init() throws IOException {
...
+      HttpResponse httpResponse = httpRequest.execute();
+
+      Assert.assertTrue(
+              (httpResponse.getHeader("content-length") == null)
+                      ^ (httpResponse.getHeader("transfer-encoding") == null));
+
+      String body = httpResponse.getBody();
+      assertThat(body).doesNotContain("a 0.0");
+      assertThat(body).doesNotContain("b 0.0");
+      assertThat(body).doesNotContain("c 0.0");
```
fstab, line 252: Can be replaced with `assertThat(body).isEmpty()`.

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
```diff
@@ -550,11 +680,29 @@ public static SSLContext createSSLContext(String sslContextType, String keyStore
...
+  }
+
+  /**
+   * Creates an HttpsConfiguration
+   * 
    * @param sslContext
    * @return HttpsConfigurator
    */
-  private static HttpsConfigurator createHttpsConfigurator(SSLContext sslContext) {
+  private final static HttpsConfigurator createHttpsConfigurator(SSLContext sslContext) {
```
fstab, line 705: `final` should be removed

fstab (changes requested):
Thaks a lot for the PR! I like the idea of refactoring the `request()` methods to make the test clearer.

However, it would be good if after the refactoring the test had less lines of code, because the new `HttpRequest` / `HttpResponse` classes allowed expressing the test more concisely. Currently, the test grew from 576 lines to 724 lines even though all the `request()` methods are removed. With a bit of refactoring we could reduce the boiler plate:

```java
Assert.assertTrue(
        (httpResponse.getHeader("content-length") == null)
        ^ (httpResponse.getHeader("transfer-encoding") == null));
```

This is copied 16 times. It would be good to have one test to make sure that the `content-length` and `transfer-encoding` headers are correct, and not to copy-and-paste this to all tests.


```java
HttpRequest httpRequest = new HttpRequest.Builder()
        .withURL("http://localhost:" + httpServer.getPort() + "/metrics?x")
        .build();

HttpResponse httpResponse = httpRequest.execute();
String body = httpRequest.execute().getBody();
```

The relevant part here is the `?x` URL parameter. This was easier to see in the previous single line `String body = request(s, "?x").body;`. Maybe we can add a helper method `request(HTTPServer, String)` that returns the initialized `Builder`? Then we could cover the simple case with `request(httpServer, "?x").build().execute().getBody()`. To reduce this even further, we could just move the `Builder` methods directly into the `HttpRequest`, then you could write `request(httpServer, "?x").execute().getBody()`.

For more complex tests we could still write something like

```
request(httpServer, "?x")
    .withHeader(...)
    .withAuthorization(...)
    .execute().getBody();
```

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
dhoard: Agreed.

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
dhoard: Agreed.

dhoard:
@fstab I have refactored the code to _mostly_ align with your suggestions.

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/HttpResponse.java
```diff
@@ -0,0 +1,112 @@
...
+                this.headers.put(header.toLowerCase(), values);
+            }
+        }
+
+        // HTTP 1.0 and 1.1 allow both headers to be null if "Connection: close"
+        if ((getHeader("content-length") != null) || (getHeader("transfer-encoding") != null)) {
+            if (false == ((getHeader("content-length") == null) ^ (getHeader("transfer-encoding") == null))) {
+                throw new IOException("Invalid HTTP response, should only contain Connect-Length or Transfer-Encoding");
+            }
+        }
```
fstab, line 46: Can be simplified to
```java
if (getHeader("content-length") != null && getHeader("transfer-encoding") != null) {
    throw new IOException("Invalid HTTP response, should only contain Connect-Length or Transfer-Encoding");
}
```

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
```diff
@@ -115,316 +107,269 @@ public void init() throws IOException {
...
+
     try {
-      String response = request(s, "?name[]=a&name[]=b").body;
-      assertThat(response).doesNotContain("a 0.0");
-      assertThat(response).contains("b 0.0");
-      assertThat(response).doesNotContain("c 0.0");
+      String body = createHttpRequestBuilder(httpServer, "/metrics?name[]=a&name[]=b").build().execute().getBody();
+      assertThat(body).doesNotContain("a 0.0");
+      assertThat(body).doesNotContain("b 0.0");
+      assertThat(body).doesNotContain("c 0.0");
```
fstab, line 206: Can be simplified to
```java
assertThat(httpResponse.getBody()).isEmpty();
```

fstab (changes requested):
It's looking great, thanks a lot! I added two little nitpick remarks, apart from that it's ready to be merged.

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/TestHTTPServer.java
dhoard: Agreed. Simplified the code to

```
HttpResponse httpResponse = createHttpRequestBuilder(httpServer, "/metrics?name[]=a&name[]=b").build().execute();
assertThat(httpResponse.getBody()).isEmpty();
```

simpleclient_httpserver/src/test/java/io/prometheus/client/exporter/HttpResponse.java
dhoard: Agreed.

fstab:
Forgot to push?

dhoard:
> Forgot to push?

Yep. pushed. Appologies.

fstab:
Thanks a lot!

PR #747, opened 2021-12-28
Deprecated HTTPServer constructors. Refactored TestHTTPServer to use …

dhoard:
Deprecated `HTTPServer` constructors. Refactored `TestHTTPServer` to use `HTTPServer.Builder`


fstab:
Thanks! This would be for rel 1.0.0, so I'll put this on hold until we have the rel 1.0.0 branch.

PR #749, opened 2022-01-04
Bump log4j version to 2.17.1 to solve CVE-2021-44832

Shoothzj:
Bumps log4j-core from 2.17.0 to 2.17.1


fstab (approved):
Thanks!

Just a note: The CVE does not apply here. It's still good to keep the dependency up-to-date, but with the current `simpleclient_log4j2` release you will still be safe.

Issue #654, opened 2021-04-21
Memory Leak ?


Laichj:
bash-4.4# jmap -histo 12 | head -50

 num     #instances         #bytes  class name
----------------------------------------------
   1:        397712     1597164184  [D
   2:        483607      295217552  [C
   3:         21586      243952824  [B
   4:        397700       15908000  io.prometheus.client.CKMSQuantiles
   5:        399649       12788768  java.util.LinkedList
   6:        469368       11264832  java.lang.String
   7:        261265        8360480  java.util.concurrent.ConcurrentHashMap$Node1
   8:         26458        7742712  [I
   9:        238620        7635840  io.prometheus.client.DoubleAdder
  10:        166874        5378600  [Ljava.lang.String;
  11:         50353        4025008  [Ljava.lang.Object;
  12:        159382        3825168  java.util.Arrays$ArrayList
  13:         38119        3354472  java.lang.reflect.Method
  14:         79540        3181600  [Lio.prometheus.client.CKMSQuantiles;
  15:         79540        3181600  io.prometheus.client.TimeWindowQuantiles
  16:         79540        2545280  io.prometheus.client.Summary$Child
  17:         19583        2162248  java.lang.Class
  18:           688        1995592  [Ljava.util.concurrent.ConcurrentHashMap$Node;
  19:         79541        1908976  [Lio.prometheus.client.CKMSQuantiles$Quantile;
  20:         18809        1482992  [Ljava.util.HashMap$Node;
  21:         79540        1272640  io.prometheus.client.Counter$Child
  22:         56075        1251976  [Ljava.lang.Class;
  23:         29710        1188400  java.util.LinkedHashMap$Entry
  24:         22828        1095744  org.aspectj.weaver.reflect.ShadowMatchImpl
  25:         32926        1053632  java.util.HashMap$Node
  26:         17473         838704  java.util.HashMap
  27:         13394         750064  java.util.LinkedHashMap
  28:         22828         730496  org.aspectj.weaver.patterns.ExposedState
  29:         44205         707280  java.lang.Object
  30:         25162         603888  java.util.ArrayList

fstab:
Little background of Quantiles: Calculating quantiles takes a lot of memory. In order to calculate the precise 0.95 quantile of a series of observations, you would need to keep a sorted array of all observations in memory, and then pick the one at the _array.size()*0.95th_ position.

[...]

Issue #550, opened 2020-05-11
Memory leak in CKMSQuantiles.java


grksharma999:
This issue is more related to Memory leak in CKMSQuantiles.java? #422
I am getting this issue on a CKMSQuantiles$Quantile

brian-brazil:
Can you provide more details?

grksharma999:
My application is running on Tomcat JVM. With increasing in requests/second to my application there is a memory leak in the JVM the memory is not getting released. Per my analysis the CollectorRegistry class usage causing the CKMSQuantiles$Quantile objects are not released.

The Heap dump in regular intervals show the leak suspect of 95% on the class CKMSQuantiles$Quantile.

brian-brazil:
Can you share graphs and code?

grksharma999:
Unfortunately I cannot share them as per the policies of my organization. But I observed it is pretty similar to the graphs of the issue #422

grksharma999:
Unfortunately I cannot share them as per the policies of my organization. But I observed it is pretty similar to the graphs of the issue #422

brian-brazil:
That wasn't an issue with client_java. I'm afraid without data, there's not much we can do.

brian-brazil:
Lacking information there's not much we can do here, and it's likely this code will be replaced at some point anyway.

Issue #422, opened 2018-09-10
Memory leak in CKMSQuantiles.java?


mdedetrich:
At work we have noticed that one of our applications have started leaking memory. After diagnosing with VisualVM over a period of time, I have noticed that the heap size of `io.prometheus.client.CKMSQuantiles$Item` is gradually increasing over time without it ever lowering. 

Using VisualVM, here is a screenshot of the first snaphot that we did
<img width="769" alt="screen shot 2018-09-10 at 16 38 05" src="https://user-images.githubusercontent.com/2337269/45304249-0014b600-b518-11e8-9ea0-632f959e47f3.png">
And here is the second, around an hour later
![screen shot 2018-09-10 at 16 39 28](https://user-images.githubusercontent.com/2337269/45304327-25a1bf80-b518-11e8-82aa-17e9341bcdcc.png)
As you can see, the amounts of `int`'s have also increased which seem to correspond to the members of the Quantiles `Item` class.

Although its expected that the number of items in the quantile increase, in our case the memory is never freed and our applications actually restart due to OOM roughly once a day. Our bucket size by default is 10 minutes, so it should be freeing up the memory shortly after that point.

We are btw using https://developer.lightbend.com/docs/cinnamon/current/plugins/prometheus/prometheus.html which uses client_java under the hood, you can also see the default configuration settings.

brian-brazil:
Hmm, that's not good. Did you wait at least 20m before taking the first snapshot? Does the growth continue similarly at about 3MB per hour?

mdedetrich:
I waited one hour before doing the first snapshot, and yes it does seem to continuously grow at a consistent rate.

I will double check this now as I am starting up some new nodes

mdedetrich:
I waited one hour before doing the first snapshot, and yes it does seem to continuously grow at a consistent rate.

I will double check this now as I am starting up some new nodes, will give an update tomorrow (it also doesn't appear to be entirely deterministic so I am clarifying this)

mdedetrich:
@brian-brazil I am doing another run now, unfortunately it didn't work last night because K8s terminated our pod halfway through.

BTW, the leak is more than 3mb per hour. 3mb is just the CKMSQuantiles$Item class, but the int's are also increasing (since they are contained within the class), in the above screenshot this is roughly 19 megs (for a total of ~22 megs per hour). 22 * 24 = 528 and our pods have 1.5 gigs of memory allocated which does roughly coincide with how often they seem to be shutting down due to OOM

[...]

PR #755, opened 2022-01-22
Improve CKMSQuantiles and address memory leak

DieBauer:
CKMSQuantiles is copied from an implementation in [2012](https://github.com/umbrant/QuantileEstimation), where it states that a ‘[HACK](https://github.com/umbrant/QuantileEstimation/blob/master/src/main/java/com/umbrant/quantile/QuantileEstimationCKMS.java#L95)’ was done, admitting a space leak. This is observed by https://github.com/umbrant/QuantileEstimation/issues/2, but never addressed.
This leak has been noticed several times in Prometheus context (#422, #550, #654).
By correctly applying the algorithm from the paper we fix the leak.

I have added unit-tests to show that the class's behaviour is correct.
I have also added a benchmark in the benchmark module showing the difference with the old (moved to the benchmark module) and current implementation.

According to my benchmarks, is in the new implementation a `get` of a quantile that has ‘seen’ 1 million elements 440 times faster.
Inserting 1 million elements is 3.5 times faster.

The amount of samples needed to keep its accuracy (within error bounds) is 80 times less than the previous implementation, and according to manual testing more or less constant, while the old implementation grows with the number of observed items (hence the space leak comment). 

```
New
Q(0,50, 0,050) was 500461,000 (off by 0,000)
Q(0,90, 0,010) was 895809,000 (off by 0,004)
Q(0,95, 0,005) was 947982,000 (off by 0,002)
Q(0,99, 0,001) was 989100,000 (off by 0,001)
Time (ms): 282
# of samples: 41

Old
Q(0,50, 0,050) was 500400,000 (off by 0,000)
Q(0,90, 0,010) was 900874,000 (off by 0,001)
Q(0,95, 0,005) was 950874,000 (off by 0,001)
Q(0,99, 0,001) was 990974,000 (off by 0,001)
Time (ms): 441
# of samples: 3246
```

While going through the CKMS paper and the Java implementation I have added remarks and snippets
from the paper, to clarify why certain choices are made.

edit: added benchmark results.

### Benchmark results
```
   Benchmark                                             (value)  Mode  Cnt    Score   Error  Units
   CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark       10000  avgt    4    0,476 ± 0,011  ms/op
   CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark      100000  avgt    4    4,794 ± 0,167  ms/op
   CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark     1000000  avgt    4   49,373 ± 2,321  ms/op
   CKMSQuantileBenchmark.ckmsQuantileOldInsertBenchmark    10000  avgt    4    0,398 ± 0,023  ms/op
   CKMSQuantileBenchmark.ckmsQuantileOldInsertBenchmark   100000  avgt    4    6,265 ± 0,253  ms/op
   CKMSQuantileBenchmark.ckmsQuantileOldInsertBenchmark  1000000  avgt    4  184,418 ± 8,570  ms/op
 

   Benchmark                                       Mode  Cnt    Score    Error  Units
   CKMSQuantileBenchmark.ckmsQuantileGetBenchmark  avgt    8  292,048 ± 35,153  ns/op
   CKMSQuantileBenchmark.ckmsQuantileOldGetBenchmark  avgt    4  128559,874 ± 1801,818  ns/op
```


fstab:
Thanks a lot for putting in the effort and reading through the `CKMSQuantiles` implementation. This is certainly not an easy piece of code.

The PR contains some refactoring and some changes in functionality. I'm trying to understand the functional changes.

The "hack" you mention in the original implementation is this:

```java
        // NOTE: according to CKMS, this should be count, not size, but this
        // leads
        // to error larger than the error bounds. Leaving it like this is
        // essentially a HACK, and blows up memory, but does "work".
        // int size = count;
        int size = sample.size();
        double minError = size + 1;
```
Your PR changes it to this:
```java
        int n = count;
        double minError = count;
```
So basically you enabled the line `int size = count;` that the original author had commented out.

The other functional changes seem to be in `insertBatch()`, so I assume the original code has a problem there leading to the "too large errors" that the original author mentioned.

I understand that you aligned the implementation more closely with the pseudo-code in the paper, and wrote tests to prove that this fixes the "too large errors" issue.

I'm wondering: Can you pinpoint what exactly is wrong in the original implementation? Maybe it's not possible to say that, but it would be awesome to know what the error was and how the change fixes it.

DieBauer:
Yes I think the PR can be split in 3 parts. 
1 is the actual fix of adhering to the algorithm. Then addition of comments and some refactoring of variable names, or unused variables. And third the addition of a Benchmark. I'm not sure where to keep the old implementation or that having this as a commit in the git history is enough to remove the 'old' part in general. Let me know what you think is best? 

Then for the actual issue of using the `sample.size()` in compress, is as far as I can tell the following, this becomes hand-wavy though:

The compress method iterates over the `LinkedList` iterator returned by `sample.listIterator()`, and according to some error bound checks removes elements. However, calling remove on this iterator changes the underlying LinkedList. 
Having the calculation depending on a 'moving target' causes problems in merging items that should belong together. 

Example: 
Let's say we have an already compressed sample.size() of 500, and a new batch size is added, extra 500 elements. After the insertBatch method we have a sample.size() of 1000. 

Given that we have a uniform distribution of [1, 500] and that we are interested in the p99 within 1%. 

We start compressing elements by looping over the sample.listIterator(). We know that all elements in sample are ordered on value. 
So the first sample we encounter has the lowest value. 
We are going to check if we can merge the first and the second element?
We calculate the 'allowableError' for this rank (which is 1, since it's the index of this iterator) and return a value based on the length of the iterator. (0.99 * 1000)
If the answer is positive (see code for actual check, there's also a delta in there which has the same flaw in this implementation we are discussing and makes the situation worse) we sum the `g` values, and adjust the iterator. Right now the length of the iterator is 1 smaller. so 999 values. 

I don't have this written down in math, but you can convince yourself that having this moving `sample.size()` adjusts the boundary check of the merging of the, now first, and second item. 

We jump right into the merging, have rank 1, and calculate a value based on the length of the iterator (0.99 * 999). 

At some point it is just not possible to match a given item with an error bound on this 'moving' sample.size().

(note that there is also a `delta` in the item which depends on the 'width' and thus indirectly on the sample.size()).
In the old implementation this would mean that the two items `I{val=996,000, g=1, del=421}` and `I{val=996,000, g=1, del=401}` could never be merged, and thus are stuck forever in the sample LinkedList. 

All of this is fixed if you do not take the rank in the sample into account, but the rank that that sample represents and not the sample size but the fixed 'observed values' size.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -35,86 +36,158 @@
...
+     * This is a different index than the bufferCount,
+     * because flushing is also done on `quantile.get()`,
+     * but we need to compress, regardless of reaching the bufferCount, to limit space usage.
      */
     private int compressIdx = 0;
 
     /**
-     * Current list of sampled items, maintained in sorted order with error
-     * bounds.
+     * The amount of values observed when to compress, equal to 1/2ε
```
fstab, line 108: I found `equal to 1/2ε` confusing, because it's always 500 independent of ε. I guess it's better to remove that from the comment.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -35,86 +36,158 @@
...
      */
-    protected LinkedList<Item> sample;
+    private final int insert_threshold;
 
     /**
      * Buffers incoming items to be inserted in batch.
+     * Incoming items are buffered into blocks of size 1/2ε
```
fstab, line 114: As above, I would remove the second line from the comment because the buffer size is always 500 independent of ε.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -185,30 +267,37 @@ private boolean insertBatch() {
...
-        for (int i = start; i < bufferCount; i++) {
+        // Keep track of the current rank by adding the g of each item. See also discussion in https://issues.apache.org/jira/browse/HBASE-14324
+        // Paper Section 3.1 on true rank: let r_i = Sum_{j=1}^{i−1} g_j
+        int currentRank = item.g;
+
+        for (int i = start; i < bufferIdx; i++) {
+            // item to be inserted
             double v = buffer[i];
+            // find the item in the samples that is bigger than our v.
             while (it.nextIndex() < sample.size() && item.value < v) {
```
fstab, line 282: Not your code, but I guess `it.nextIndex() < sample.size()` could be replaced with `it.hasNext()`.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -217,46 +306,83 @@ private boolean insertBatch() {
...
+         * is a sampled item from the data stream and two additional
+         * values are kept
+         */
         public final double value;
+        /**
+         * g_i is the difference between the lowest
+         * possible rank of item i and the lowest possible rank of item
+         * i − 1
+         */
         public int g;
```
fstab, line 380: Again not your code, but `g` should be `final` as well.

fstab (changes requested):
Thanks a lot again for the fix the explanation. I took the time to read the paper, and I think your changes are correct.

I have a couple of minor code review comments, but none of them is related to the algorithm itself.

Please also remove `CKMSQuantilesOld`, no need to keep that.

DieBauer:
> Thanks a lot again for the fix the explanation. I took the time to read the paper, and I think your changes are correct.
> 
> I have a couple of minor code review comments, but none of them is related to the algorithm itself.
> 
> Please also remove `CKMSQuantilesOld`, no need to keep that.

Thanks for taking the time and effort to review this! I'll address the comments. 

Let me know what the git commit strategy is (squash/rebase, ..)

fstab:
I like rebase and squash, so that we have a linear history with a single commit per PR.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -35,86 +36,158 @@
...
+ * Output(φ):
+ *    r0:=0;
+ *    for i := 1 to s do
+ *        ri:=ri−1+gi−1;
+ *        if(ri +gi +∆i >φn+f(φn,n)/2)
+ *            print(vi−1); break;
+ * </pre>
  */
-class CKMSQuantiles {
+public final class CKMSQuantiles { /* public for benchmark-module */
```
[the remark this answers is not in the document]
DieBauer: Unfortunately, the benchmark is in a subpackage of `client`: `io.prometheus.client.benchmark` so not accessible when not public.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
```diff
@@ -35,86 +36,158 @@
...
 
+    /**
+     * Set up the CKMS Quantiles. Can have 0 or more targeted quantiles defined.
+     * @param quantiles The targeted quantiles, can be empty.
+     */
     public CKMSQuantiles(Quantile[] quantiles) {
-        this.quantiles = quantiles;
+        // hard-coded epsilon of 0.1% to determine the batch size, and default epsilon in case of empty quantiles
+        double pointOnePercent = 0.001;
+        if (quantiles.length == 0) { // we need at least one for this algorithm to work
```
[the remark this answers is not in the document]
DieBauer: Agreed, I mainly have added this because in the comments of summary it states: 
```java
  final List<Quantile> quantiles; // Can be empty, but can never be null.
```
https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Summary.java#L83

I'm not sure what changes when this restriction is applied to users of Summaries that don't have specific quantiles applied to the summary.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
DieBauer: Agreed, I experimented with a dynamic epsilon, however, this is tricky as having a 'wide' error would mean that this buffer is filled very quickly. e.g. epsilon of 0.1 gives a size of 1/(0.2) = 5. So then you would add logic stating, at least 500 elems, at most 10000 or something arbitrary like that. 

I will keep it fixed at 500, and add a comment that this is an arbitrary decision.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
DieBauer: agree!

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
DieBauer: This `g` is written to in the compress method. `next.g += prev.g;`. I will remove the public from the members in `Quantile` and `Item` though.

simpleclient/src/main/java/io/prometheus/client/CKMSQuantiles.java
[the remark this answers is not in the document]
DieBauer: For now I'll move the benchmark up. Let me know how you want to proceed.

fstab:
Thank you!

fstab:
FYI I found a little bug in the million entries test after I merged the PR:

```java
Collections.shuffle(Arrays.asList(shuffle), rand);
```
`Arrays.asList(shuffle)` returns `Collection<int[]>` not `Collection<Integer>`, so the array was never shuffled.

With an actual random array it turns out that the corner case `Quantile(0.0, 0.01)` doesn't work.

I fixed the test and changed the constructor of `Quantile` to make sure the target quantile is strictly `> 0.0` and `< 1.0`, because allowing `0.0` and `1.0` doesn't work.

fstab:
I found another bug, so I reverted the PR on `master` and moved it to a `ckms` branch.

The bug is: If you take `testGetWithAMillionElements()` but remove all quantiles except for the `0.95` quantile (remove them both from the test setup at the beginning of the test as well as from the assertions at the end of the test) then the test fails:

```
java.lang.AssertionError: quantile=0.95, actual=855053.0, lowerBound=929999.9999999999
```

I'm not sure at the moment what causes this.

fstab:
If you find what's going wrong, feel free to create a PR for a fix against the `ckms` branch.

DieBauer:
I have looked into this. In the case of only 1 quantile, the `insertThreshold` is too high (hard coded 500). 
This can be mitigated by setting the threshold, and thus compressionIdx, (as is written in the paper) to 1/(2*epsilon) = 25, then the test passes. 
We could implement this optimization in case of 1 targeted quantile.
Having 2 or more quantiles seems to mitigate this in general. 

I have run the same unit-test suite against the 'old' implementation and multiple tests fail, also the 'edge'-case you mentioned. 

I have a fix for the quantiles of 1.0 and 0.0. These should be allowed. The compress function was too eager and should not have compressed the minimum observed value.
I'll create a PR to address this to the ckms branch.

DieBauer:
I benchmarked the length of the buffer (1 / 2 e) given an epsilon. 

Adding a million elements gives this. 
```
Benchmark                                          (epsilon)  Mode  Cnt    Score    Error  Units
CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark       0.01  avgt    4  151,359 ±  7,421  ms/op
CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark      0.001  avgt    4  174,215 ±  3,788  ms/op
CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark     0.0001  avgt    4  238,256 ± 20,575  ms/op
```
So with an epsilon of 1 percent, the buffersize is 50, while at 0.01% this is 5000. There the time it takes to sort the buffer  kicks in. 

The arrays.sort benchmark looks like this:
```
Benchmark                   (size)  Mode  Cnt      Score      Error  Units
CKMSQuantileBenchmark.sort      50  avgt    4    182,980 ±    1,044  ns/op
CKMSQuantileBenchmark.sort     500  avgt    4   1988,166 ±  788,125  ns/op
CKMSQuantileBenchmark.sort    5000  avgt    4  30943,722 ± 5770,539  ns/op
```

Arrays.sort grows faster with a larger number of elements than 10 * a smaller number of elements. 

Now in the insert benchmark, there is also the hidden cost of a smaller epsilon, thus the samples.size() grows faster to gain the precision required. 

I think it's best to have a buffer size dependent on the epsilon, to cater for the case where an epsilon is way bigger than the hard-coded one, and max it at 500, to not have too much performance degradation.

fstab:
Shouldn't the algorithm be correct independent of the `insertThreshold`? The threshold just says how often `compress()` is called. If the algorithm is correct we should get correct results independent of how often we call `compress()`.

DieBauer:
I'll try to see if I can get a better answer. The insertThreshold is indeed for compressing, when to compress or not depends on the amount of observed values `n` in `allowableErrors`. `n` is updated every `insertBatch`, I believe that when the batches are inserted more often, and thus the count updated, that the error margins are smaller, or closer to reality than with batches updated less often.
So indirectly, does the compress depend on the batch size.

DieBauer:
I need to look into this statement in section 6.2: Space Usage for Targeted Quantiles

> For the random input, the bounds
gave guarantees that held with 70% probability to find
quantiles that our algorithm found with absolute certainty.
For the “hard” input, which attempts to force the worst-case
space usage, the probability for the randomized algorithm
improved to around 95%, still far short of the low failure
rates demanded by network managers. Hence, we do not
report further on the results obtained by random sampling
for the remainder of this section

fstab:
Hi, looks like we were both working on it at the same time. I found this as well, and a couple of other issues (subtle rounding errors in the error function, which sounds small, but once a sample is at a wrong position this affects future inserts; doing compress from right to left instead of left to right; etc).

I also added much more fine grained tests that go through the list of samples after each insert() and compress() and make sure that the invariants defined in the paper are satisfied. That way I found some issues that are hard to see if you just test the end result.

I also realized that if you read that code, you need to have the paper next to you. So I removed comments that are just paragraphs from the paper, because people will read them in the paper anyway and not in the Javadoc. I also renamed a few things to be aligned with the names in the paper, like `allowableError() -> f()` or `count -> n`.

I pushed a commit where my tests are green, but I guess it would be good to have another look. And I did not run any benchmarks.

BTW regarding the 0.0 and 1.0 quantiles: I did not look at this, but the question is what happens if a user just specifies a 0.0 quantile to track the minimum or a 1.0 quantile to track the maximum, and no other quantiles?

DieBauer:
Alright, great to hear! I can imagine the amount that has to go in to testing this properly, which I skipped because of the 'ballpark' figures I got from the high-level tests. 

Let me know where I can have a look at your changes! 

Regarding only defining a quantile with 0 or 1, yes these are indeed edge cases of min and max in the value stream. 
I'm not sure if there's anything special that happens, of course there is overhead of going through the compress and insert.

fstab:
I thought about the `0` and `1` quantiles again, and it turned out that it was simpler to support this than I thought. I just pushed an update allowing `0` and `1`. Thanks a lot for pointing out that this would be useful.

fstab:
I merged it once more, thanks a lot for your contribution!

DieBauer:
Nice. 

I did benchmarking on the current implementation, insertion is constant (within error bounds) while in the old implementation it was linear. 
```
     OLD
     Benchmark                                          (value)  Mode  Cnt    Score    Error  Units
     CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark    10000  avgt    4    1,104 ±  0,013  ms/op
     CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark   100000  avgt    4   15,577 ±  2,418  ms/op
     CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark  1000000  avgt    4  366,650 ± 48,997  ms/op
     NEW
     Benchmark                                          (value)  Mode  Cnt    Score    Error  Units
     CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark    10000  avgt    4    1,549 ±  0,360  ms/op
     CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark   100000  avgt    4   16,395 ±  1,138  ms/op
     CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark  1000000  avgt    4  168,386 ± 35,664  ms/op
``` 

Lookups are 1 or 2 μs for p95 with 1 million elements. 
```
     Benchmark                                       (value)  Mode  Cnt     Score     Error  Units
     CKMSQuantileBenchmark.ckmsQuantileGetBenchmark    10000  avgt    4  1009,883 ± 127,939  ns/op
     CKMSQuantileBenchmark.ckmsQuantileGetBenchmark   100000  avgt    4  1484,729 ± 136,455  ns/op
     CKMSQuantileBenchmark.ckmsQuantileGetBenchmark  1000000  avgt    4  2245,291 ± 238,582  ns/op
```

This is because the sample size also increases with the number of elements.
see these results
```
testInsertBatch(batchSize=128, compressInterval=128, totalNumber=10000)
10000 samples: 77
testInsertBatch(batchSize=128, compressInterval=128, totalNumber=100000)
100000 samples: 92
testInsertBatch(batchSize=128, compressInterval=128, totalNumber=1000000)
1000000 samples: 123
```

So it is expected that looking up the same percentile in a larger list takes more time. 


One small improvement I see is in the `f` method, you account for the double precision rounding, and then do a `Math.floor` and then cast to int. This is more expensive than just casting to int. And since we can be sure that `u`, `v`, `q` and `r` are positive the result will be positive, so in the end the result is not changed.

This change is worthwhile since `f` is being called in many places (insert, lookup, compress). 

Here's a benchmark result when removing the math.floor call. 

```
CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark    10000  avgt    4    1,119 ± 0,068  ms/op
CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark   100000  avgt    4   11,775 ± 0,698  ms/op
CKMSQuantileBenchmark.ckmsQuantileInsertBenchmark  1000000  avgt    4  133,567 ± 5,849  ms/op
```

And the f method itself: Count = 10000 and r = 9500
```
Benchmark                               Mode  Cnt   Score   Error  Units
CKMSQuantileBenchmark.ckmsQuantileF  avgt    4  37,188 ± 6,681  ns/op
CKMSQuantileBenchmark.ckmsQuantileF  avgt    4  20,500 ± 1,054  ns/op <= Without floor()
```

I'll create a PR for that.

fstab:
It's released :tada:. Thanks again for this awesome PR!

fstab:
https://github.com/prometheus/client_java/releases/tag/parent-0.15.0

PR #756, opened 2022-01-25
Refactored code to allow the use of a custom ExecutorService

dhoard:
Refactored code to allow the use of a custom ExecutorService.

Code was refactored since part of the building logic was in `HTTPServer.Builder` and part of the logic was in a `HTTPServer` constructor, resulting in no good way to introduce the `ExecutorService`. With the refactor, all `HTTPServer` constructors now use the `HTTPServer.Builder` for a single point of construction.

Added `TestHTTPServerBuilder` to validate all of the possible builder configurations.

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -199,6 +199,7 @@ public class HTTPServer implements Closeable {
         private InetAddress inetAddress = null;
         private InetSocketAddress inetSocketAddress = null;
         private HttpServer httpServer = null;
+        private ExecutorService executorService = null;
         private CollectorRegistry registry = CollectorRegistry.defaultRegistry;
         private boolean daemon = false;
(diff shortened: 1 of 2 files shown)


dhoard:
@fstab please review.

fstab:
Thanks a lot for the PR. There are a lot of changes in this one, and I feel it would be easier to split this into independent PRs that can be discussed separately.

In general, people should use the `Builder` to create the `HTTPServer` and not call the constructor directly. We have some public constructors that we need to keep for backwards compatibility, but new configuration (like SSL config) will be implemented in the `Builder` and not in the constructor.

So in order to make the `ExecutorService` configurable, it would be sufficient to add the `withExecutorService()` method to the `Builder` and set the executor service in `Builder.build()`.

I know this PR is a couple of months old. Are you still available to continue this? It would be good if you could strip it down to a smaller PR that just adds the `withExecutorService()`.

dhoard:
@fstab code has been refactored. Please review.

fstab (approved):

fstab:
Thanks a lot!

PR #759, opened 2022-01-31
Improve shortcomings in the first iteration of CKMS

DieBauer:
After discussion in #755 a next iteration:
- Introduce a dynamic batch size with an upper limit of 500. It looks like smaller batch size gives better results (precision) for single quantiles. 
- Skip merging the first item of the samples, to always have the real value of p0. 
- Enable the quantile to be 0 or 1, and short-circuit the get method to directly get the values for p0 and p100. 
- Fixes a bug that the pXX value was not within error bounds


fstab:
Fixed with #755.

PR #760, opened 2022-02-03
CMKSQuantiles: Remove call to Math.floor in f()

DieBauer:
Since we can be sure that the result is always positive, rounding down with casting to int is faster than using Math.floor while getting the same result.

Added the Benchmark to validate the change. 
```
Benchmark                            Mode  Cnt   Score   Error  Units
CKMSQuantileBenchmark.ckmsQuantileF  avgt    4  37,188 ± 6,681  ns/op
CKMSQuantileBenchmark.ckmsQuantileF  avgt    4  20,500 ± 1,054  ns/op <= Without floor()
```

This might not seem much, but f is being called almost all the time (insert, compress, get) and this 17 ns difference can add up to a significant amount.


fstab:
Thanks a lot!

PR #762, opened 2022-02-06
CKMSQuantiles: Extract upperbound calculation to be done once per get

DieBauer:
The upperbound is constant for a given quantile, and does only depend on the desired rank. 

By calculating this once we save on calculations of this "constant".

On an iterator size of 39, and 4 quantiles defined, the get is almost 5 times faster. 
```
Benchmark                                       (value)  Mode  Cnt    Score    Error  Units
CKMSQuantileBenchmark.ckmsQuantileGetBenchmark    10000  avgt    4  70,541 ±  5,932  ns/op *Inlined*
CKMSQuantileBenchmark.ckmsQuantileGetBenchmark    10000  avgt    4  340,909 ±  4,159  ns/op *current*
```

### DIFF
--- a/benchmarks/src/main/java/io/prometheus/client/CKMSQuantileBenchmark.java
+++ b/benchmarks/src/main/java/io/prometheus/client/CKMSQuantileBenchmark.java
@@ -97,6 +97,10 @@ public class CKMSQuantileBenchmark {
             for (Double l : shuffle) {
                 ckmsQuantiles.insert(l);
             }
+            // make sure we inserted all 'hanging' samples (count % 128)
+            ckmsQuantiles.get(0);
+            // compress everything so we have a similar samples size regardless of n.
+            ckmsQuantiles.compress();
             System.out.println("Sample size is: " + ckmsQuantiles.samples.size());
         }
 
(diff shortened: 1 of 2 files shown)


fstab (approved):
Awesome, thanks a lot!

fstab:
By the way, I am wondering if `LinkedList` is actually a good data type for the `samples` from a performance standpoint, as it creates a lot of small objects at different memory locations.

Using an array (like an actual `int[]` or `double[]`, not an array of objects) might help to keep data in a single cache line, which would be a lot faster.

Maybe there is a way to encode the samples like `samples[n-1] = value; samples[n-2] = g; samples[n-3] = delta; ...` and just operate on that array directly. For `insert()` we could just add a new value to `samples` and accept that `samples` is not sorted after insert. For `get()`, we need to sort first, but we do not need to sort and flush the `buffer`, as we operate on the `samples` array directly. `compress()` could be implemented in-place, as we just move samples up when we merge them.

Anyway, just an idea, I didn't try it.

DieBauer:
Agreed, there is probably room for improvement. I won't pursue that right now. With this effort we put in, we have fixed the space requirement, several bugs, and the lookup time of a quantile significantly. 
Thanks for all feedback and help!

PR #768, opened 2022-03-10
Fix conditional for jvm_memory_bytes_committed

SvenssonWeb:


SvenssonWeb:
Hello @fstab. Do you have any comments?

fstab:
Thanks for the reminder, and thanks for fixing. Yes, you are right, that was the wrong metric name.

Issue #342, opened 2018-02-15
.label() Method throws Exceptions


Tigraine:
Apparently if labels are missing or a null value is supplied to a label the prometheus client will throw a `IllegalArgumentException`.

Exceptions being thrown on the "hot" path with null values in labels are a problem as they either require me to add try/catch logic around my monitoring code or decorate the whole prometheus client with try/catch blocks.

IMO there are a ton of instances where you might have a label that is not compile-time supplied (I know in most examples it's the HTTP method being passed).
But for a lot of cases it could be something like a AWS Region-ID or a Cluster-ID or a machine name or endpoint name etc..
Not saying that any of these values being null is a good thing, but errors happen sometimes and especially when you are dealing with legacy systems these things are commonplace.
Having code break in very unexpected ways because prometheus complains about it's arguments is really really problematic. 

The offending code in the client is here:
https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java#L64-L68

Is there any policy on metrics? Because in my book metrics/logging should be totally transparent and never require any additional logic. If I can't call the prometheus client with absolute certainty that it won't break my code under ANY circumstances I'm hesitant to include it as I am opening myself up to another level of complexity (and all of this suddenly needs testing etc). 

Imo it's preferable to have missing labels be mis-reported as empty string and the client being 100% runtime safe than having a operation fail because some parameter was not correctly checked.

brian-brazil:
We require a string, and the problem is that the empty string is a valid value for a label value so then you couldn't distinguish.

Tigraine:
I know - but having monitoring code break applications is the even worse case imo. 
I'd propose something like a strict/lax mode on the builder where you can express intentionally that you want this null-checked.

My concern is that I want my team to easily and without requiring additional testing etc to be able to sprinkle prometheus metrics throughout the applications without having to think about it breaking something or not.

Tigraine:
I could also see something like this:

```
 new Counter.Builder()
                .name("queries_failed_total")
                .help("Number of queries that failed")
                .labelNames("method", "org", "reason")
                .nullHandler(Counter.TreatMissingLabelsAsEmptyString)
                .register();
```

brian-brazil:
If you want liberal sprinkling, then an extra line of code you have to add kinda goes against that. This is something that'd have to work by default.

Tigraine:
I'd love to send a PR to address this - but for that I'd need to know what you think about this.
[...]

PR #776, opened 2022-04-22
#342 - Implemented Label Value Sanitizer

Tigraine:
@brian-brazil as discussed (a few years ago 🤣 ) here https://github.com/prometheus/client_java/issues/342

This PR adds the ability to all SimpleCollector instances to manipulate labels before they are processed.
This allows users to extend a label to be runtime safe or do manipulation like removing credentials etc..

Example would be:

```
String clientName = getClientName(); // this label is not null-safe
counter.label(clientName).inc() <-- will throw IllegalArgumentException only during runtime
```

Usage:
```
metric = Gauge.build()
            .name("nonulllabels")
            .help("help")
            .labelNames("l")
            .labelValueSanitizer(Gauge.TransformNullLabelsToEmptyString())
            .register(registry);
```

Instead of having to sanitise all labels before calling the instrumentation code we can make the code safe at setup and not care about downstream mis-use.
Also since this is implemented through an interface additional use cases can be passed by consumers of the library.


simpleclient/src/main/java/io/prometheus/client/LabelValueSanitizer.java
```diff
@@ -0,0 +1,5 @@
+package io.prometheus.client;
+
+public interface LabelValueSanitizer {
```
fstab, line 3: As you say this can also be used for other use cases, not just "sanitization". So maybe the interface should have a more generic name covering all use cases, like `LabelValueTransformer`?

simpleclient/src/main/java/io/prometheus/client/LabelValueSanitizer.java
```diff
@@ -0,0 +1,5 @@
...
+    String[] sanitize(String... labelValue);
```
fstab, line 4: I think it would be better to sanitize only a single label value, not the whole array. For example, you could register a sanitizer like this:
```java
metric = Gauge.build()
        .name("nonulllabels")
        .help("help")
        .labelNames("path", "status)
        .labelValueSanitizer("path", myPathSanitizer)
        .register(registry);
```
and then the `myPathSanitizer` would only be called for the path label value, but not for an array of all label values.

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -174,12 +176,42 @@ protected SimpleCollector(Builder b) {
...
+        for (int i = 0; i < labelValue.length; i++) {
+          if (labelValue[i] == null) {
+            labelValue[i] = "";
+          }
+        }
+        return labelValue;
+      }
+    };
+  }
+
```
fstab, line 214: If the sanitizer takes only a single string as argument instead of an array of strings, a good handler for null values would be `String::valueOf`. As this already exists in the Java runtime, there's no need to provide our own implementation in `SimpleCollector`.

fstab (changes requested):
Thanks a lot for the PR. I like the idea of having a generic way to transform labels. I added a few thoughts as review comments. Let me know what you think.

fstab:
Regarding the specific case of `null` values: If we revisit this and _not_ throw an `IllegalArgumentException`, what do you think would be a good default value? The literal string `"null"` as returned by `String::valueOf`, or an empty value?

simpleclient/src/main/java/io/prometheus/client/LabelValueSanitizer.java
dhoard: I agree with @fstab ... `LabelValueTransformer` seems like it could be used more generically.

simpleclient/src/main/java/io/prometheus/client/LabelValueSanitizer.java
fstab: Depends on what use case you have in mind.

What I have in mind is that you have a specific label that you want to transform (for example stripping a unique user ID from an HTTP path). For that, it would be good to implement a specific `pathTransformer` and apply it to the `path` label. Advantage is: You don't need to magically know at which position in the array the `path` label is. And if `path` is passed as a label to multiple different metrics, you can reuse the transformer, even if the `path` is on different positions for each metric.

The array as parameter only makes sense if you want to do exactly the same transformation for all labels. This would work for replacing `null` with `"null"`, but for every other use case I think it's better if you know which label you want to transform.

Tigraine:
> Regarding the specific case of `null` values: If we revisit this and _not_ throw an `IllegalArgumentException`, what do you think would be a good default value? The literal string `"null"` as returned by `String::valueOf`, or an empty value?

Thanks for the feedback! I will update the PR to move to a Transformer as it makes a lot of sense.

I was thinking about the appropriate default value a bit and in our case 'NULL' would be a perfectly good default value - but that might not be for everyone as a default which is the main reason I came up with this PR and the concept of Sanitization.

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -174,12 +176,42 @@ protected SimpleCollector(Builder b) {
...
     }
   }
 
+  /**
+   * Default Sanitizer - Will not perform any manipulation on labels
+   */
+  static LabelValueSanitizer NoOpLabelValueSanitizer() {
+    return new LabelValueSanitizer() {
+      @Override
+      public String[] sanitize(String... labelValue) {
```
dhoard, line 192: Change `labelValue` to `labelValues` since it's an array.

simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java
```diff
@@ -174,12 +176,42 @@ protected SimpleCollector(Builder b) {
...
+    };
+  }
+
+  /**
+   * Transforms null labels to an empty string to guard against IllegalArgument runtime exceptions
+   */
+  static LabelValueSanitizer TransformNullLabelsToEmptyString() {
+    return new LabelValueSanitizer() {
+      @Override
+      public String[] sanitize(String... labelValue) {
```
dhoard, line 204: Change `labelValue` to `labelValues` since it's an array.

simpleclient/src/main/java/io/prometheus/client/LabelValueSanitizer.java
dhoard: There may be future use cases where the functionality could be used to transform `String` values other than labels, so I feel it would make more sense to rename the interface to `StringTransformer`.

PR #781, opened 2022-05-17
Support multi-line help texts

sofiedc:
Linefeeds are no longer escaped.
Additional lines are prefixed by '# '.

This way, metrics and their labels / values can be documented more in detail.


fstab:
Thanks a lot. Unfortunately I think this would break compatibility with [OpenMetrics](https://openmetrics.io/), see the [ABPF definition of the Text format](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf).

sofiedc:
Thanks for the input. We actually wanted to use the HELP text as a kind of API doc, e.g.
image

Is there any recommendation on how to expose such documentation? Should always be separated from the metrics page?

dhoard:
@sofiedc you should document externally in a wiki, Jira, etc.

sofiedc:
ok, thanks, I'll close the review

Issue #780, opened 2022-05-12
Dropwizard metrics are not exported in simpleclient_vertx or simpleclient_vertx4


yaronel:
Using [DropwizardExports](https://github.com/prometheus/client_java/blob/master/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java) doesn't work with neither [simpleclient_vertx](https://github.com/prometheus/client_java/blob/master/simpleclient_vertx/src/main/java/io/prometheus/client/vertx/MetricsHandler.java) nor [simpleclient_vertx4](https://github.com/prometheus/client_java/blob/master/simpleclient_vertx4/src/main/java/io/prometheus/client/vertx/MetricsHandler.java).

The Vertx handler calls [`CollectorRegistry.filteredMetricFamilySamples`](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L139) which returns an enumeration of collectors from `CollectorRegistry.namesToCollectors`. 

DropwizardExports implements `Collector.Describable` but it's [`describe`](https://github.com/prometheus/client_java/blob/master/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java#L198) method returns an empty `List<MetricFamilySamples>` so nothing is added to the `CollectorRegistry.namesToCollectors` [when it's registered](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L108). However, the empty list of names does get added to [`CollectorRegistry.collectorsToNames`](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L63) which is how I was able to go around this issue for the time being.

Right now the only way I was able to export the Dropwizard metrics was by calling [`CollectorRegistry.metricFamilySamples`](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L128) which uses the `CollectorRegistry.collectorsToNames`.

fstab:
:+1: thanks for reporting. I implemented a fix in #785, but didn't test it yet. Would you mind to have a look?

yaronel:
@fstab Hey, sorry I didn't get a notification so only seeing your comment now.
I checked it and it still doesn't work.
[...]

PR #785, opened 2022-05-19
Fix Describable returning an empty list

fstab:
As described in #780, if a `Collector` implements `Describable`, but the `describe()` method returns an empty list, the metrics from this `Collector` will never be collected.

This PR should fix this.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -169,10 +169,15 @@ public class CollectorRegistry {
         HashSet<Collector> collectors = new HashSet<Collector>();
         synchronized (namesCollectorsLock) {
           for (Map.Entry<Collector, List<String>> entry : collectorsToNames.entrySet()) {
-            for (String name : entry.getValue()) {
-              if (sampleNameFilter.test(name)) {
-                collectors.add(entry.getKey());
-                break;
+            List<String> names = entry.getValue();
+            if (names.isEmpty()) {
+              collectors.add(entry.getKey());
+            } else {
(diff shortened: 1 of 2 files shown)


simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
```diff
@@ -168,10 +168,12 @@ private Iterator<Collector> filteredCollectorIterator() {
       } else {
         HashSet<Collector> collectors = new HashSet<Collector>();
         synchronized (namesCollectorsLock) {
-          for (Map.Entry<String, Collector> entry : namesToCollectors.entrySet()) {
-            // Note that namesToCollectors contains keys for all combinations of suffixes (_total, _info, etc.).
-            if (sampleNameFilter.test(entry.getKey())) {
-              collectors.add(entry.getValue());
+          for (Map.Entry<Collector, List<String>> entry : collectorsToNames.entrySet()) {
+            for (String name : entry.getValue()) {
```
yaronel, line 172: Maybe I am missing something but it doesn't seem to fix the issue.
The problem is the `List<String>` of names is empty, so this inner loop is not going to run.
One way to make it work, though I don't know if this is the behavior you want, is something like this:
```java
          for (Map.Entry<Collector, List<String>> entry : collectorsToNames.entrySet()) {
            if (entry.getValue().isEmpty()) {
              collectors.add(entry.getKey());
            } else {
              for (String name : entry.getValue()) {
                if (sampleNameFilter.test(name)) {
                  collectors.add(entry.getKey());
                  break;
                }
              }
            }
          }
```

fstab:
Thanks a lot @yaronel. I fixed it and added a test to make sure it works.

PR #786, opened 2022-05-19
Reduce number of core threads in HTTPServer to one

fstab:
In a typical scenario with a Prometheus server scraping every 15 seconds, one thread should be enough in `HTTPServer`. Reduce the number of core threads from 5 to 1.

### DIFF
--- a/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
+++ b/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java
@@ -1,10 +1,15 @@
 package io.prometheus.client.exporter;
 
+import com.sun.net.httpserver.Authenticator;
+import com.sun.net.httpserver.HttpContext;
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
 import com.sun.net.httpserver.HttpsConfigurator;
 import com.sun.net.httpserver.HttpsServer;
 import io.prometheus.client.CollectorRegistry;
-import io.prometheus.client.SampleNameFilter;
 import io.prometheus.client.Predicate;
+import io.prometheus.client.SampleNameFilter;
 import io.prometheus.client.Supplier;
 import io.prometheus.client.exporter.common.TextFormat;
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
This PR violates the OpenMetrics spec and should be reverted. Concurrent
exposition must be supported, as it is very common such as a HA pair. This
PR will break users, failing scrapes and adding artifacts to their graphs.

fstab:
Thanks for keeping an eye on PRs! In this case I think there is just a misunderstanding with the weird terminology of `ThreadPoolExecutor`:
* `corePoolSize`: the number of threads to keep in the pool, even if they are idle
* `maximumPoolSize` the maximum number of threads to allow in the pool

This PR sets the `corePoolSize` to 1, but leaves the `maximumPoolSize` at 5. It's still possible to run 5 requests in parallel, but the executor won't keep 5 threads running if they're idle.

brian-brazil:
Thanks for the clarification, that makes perfect sense to do.

ctrlaltluc:
We have a setup using the Prometheus client including this change, and we started to experience some inconveniences. 

Our setup is Kafka on Kubernetes, running [prometheus/jmx_exporter](https://github.com/prometheus/jmx_exporter) as an agent attached to the Kafka process. Our scrape duration is about 12 seconds.
The Kafka brokers are fronted by Envoy reverse proxies, which have periodic health-checks ([with 1s timeouts](https://github.com/banzaicloud/koperator/blob/master/pkg/resources/envoy/configmap.go#L372)) calling the `/-/healthy` endpoint [provided by the JMX exporter](https://github.com/prometheus/client_java/blob/master/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java#L92-L93).

After upgrading the Prometheus client to include this fix, we started experiencing health-check network timeouts. Investigations showed us that the latency when calling `/-/healthy` increases from 10-60ms (with the core pool size of 5) to 10ms-15s (with the core pool size of 1). Since this change is the only one causing these timeouts, we attribute this increase to the overhead of creating a new thread, if the pool has no available one to serve `/-/healthy`.

We normally thought about increasing the timeout, which is one of the options beside rolling back the Prometheus Java client version to 0.16. 
I just wanted to raise this here, because the latency distribution tail increased from under 100ms to 15s seems pretty big.

fstab:
@ctrlaltluc thanks a lot for pointing this out. This was a bug. Apparently the description of `corePoolSize` and `maximumPoolSize` is misleading, as explained in [this Blog post](http://www.bigsoft.co.uk/blog/2009/11/27/rules-of-a-threadpoolexecutor-pool-size).

I pushed a fix, switching to a cached thread pool executor.

ctrlaltluc:
@fstab nice guy, Sun! 

[This](http://www.bigsoft.co.uk/blog/2009/11/27/rules-of-a-threadpoolexecutor-pool-size) explains the behavior of `/-/healthy` timing out on our side too. The health-checks were served, but when the core thread managed to get the task from the queue. 

Thanks!

PR #787, opened 2022-05-20
Filter metrics by suffix

fstab:
Allow filtering metrics by suffix. For example, this can be used to filter out `_created` metrics.

There are different ways to configure this, depending on how you expose metrics:

## Manual creation of the HTTPServer

```java
    HTTPServer httpServer = new HTTPServer.Builder()
        .withRegistry(registry)
        .withSampleNameFilter(new SampleNameFilter.Builder()
            .nameMustNotEndWith("_created")
            .build())
        .build();
```

See `TestHTTPServer.testCreatedMetricRemoved()`.

## Exporter Servlet

```xml
    <servlet>
        <servlet-name>prometheus-exporter</servlet-name>
        <servlet-class>io.prometheus.client.servlet.jakarta.exporter.MetricsServlet</servlet-class>
        <init-param>
            <param-name>name-must-not-end-with</param-name>
            <param-value>_created</param-value>
        </init-param>
    </servlet>
```

## jmx_exporter

We are working on a new configuration file format for `jmx_exporter`, see the `new-config` branch. The new config will have a section like this:

```yaml
metricFilter:
  nameMustNotEndWith:
    - _created
```

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/SampleNameFilter.java
+++ b/simpleclient/src/main/java/io/prometheus/client/SampleNameFilter.java
@@ -22,13 +22,17 @@ public class SampleNameFilter implements Predicate<String> {
     private final Collection<String> nameIsNotEqualTo;
     private final Collection<String> nameStartsWith;
     private final Collection<String> nameDoesNotStartWith;
+    private final Collection<String> nameEndsWith;
+    private final Collection<String> nameDoesNotEndWith;
 
     @Override
     public boolean test(String sampleName) {
         return matchesNameEqualTo(sampleName)
-                && !matchesNameNotEqualTo(sampleName)
+                && matchesNameNotEqualTo(sampleName)
                 && matchesNameStartsWith(sampleName)
-                && !matchesNameDoesNotStartWith(sampleName);
+                && matchesNameDoesNotStartWith(sampleName)
+                && matchesNameEndsWith(sampleName)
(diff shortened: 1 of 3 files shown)


brian-brazil:
> We are working on a new configuration file format for jmx_exporter, see the new-config branch. The new config will have a section like this:

This seems redundant, you can already configure the jmx exporter to drop samples by setting `name` to empty.

However you likely don't even need that - `_created` samples will generally only be there if there's mBeans exposing them and you've rules covering that with the JMX exporter, which wouldn't be typical at all..

fstab:
The main use case I had in mind for the metric filter was that some users want to filter out `jvm_threads_deadlocked` and `jvm_threads_deadlocked_monitor` because as the Javadoc says this might be an expensive operation under some circumstances. However, different users might want to filter different things, so I just exposed  generic filtering by prefix/suffix/exact match.

One aspect that might be confusing: The name filters are applied after scraping the JMX bean. If you want to prevent the JMX bean from being scraped, there's the `jmxBeanFilter` (previously called blacklist / whitelist).

Anyway: A full example is [test-config-new.yaml](https://github.com/prometheus/jmx_exporter/blob/new-config/collector/src/test/resources/test-config-new.yaml), but it's very early stages. I implemented a parser that's backwards-compatible with the current format. Next step is to write tests for different combinations of SSL config (both for the collector via jmxUrl and for the HTTPServer), but that's quite time consuming.

PR #789, opened 2022-06-01
Move servlet adapters to an internal package to avoid duplicating classes when building OSGi bundles

### LINKED ISSUE #771: simpleclient_servlet duplicates classes from simpleclient.jar starting in 0.12
Starting with release 0.12, the simpleclient_servlet jar now contains duplicates of most (all?) classes from the core client. The dependency from the servlet module on the main module still exists, so I can only assume this is unintentional.

adessaigne:
This fixes #771 

OSGi requires that any particular package is exposed by only one bundle (jar). With the `Adapter` class directly in the root `io.prometheus.simpleclient` it was copying all the other classes from this particular package and so copying the classes from `simpleclient`. I've chosen `io.prometheus.simpleclient.internal` as no package named internal is exported by default in OSGi. So, in OSGi environments, this class will only be visible by `MetricsServlet` and `MetricsFilter`

I've also included a fix of a typo that was preventing the adapter class to work as expected.

### DIFF
--- a/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
+++ b/simpleclient_servlet/src/main/java/io/prometheus/client/exporter/MetricsServlet.java
@@ -1,6 +1,6 @@
 package io.prometheus.client.exporter;
 
-import io.prometheus.client.Adapter;
+import io.prometheus.client.internal.Adapter;
 import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.Predicate;
 import io.prometheus.client.servlet.common.exporter.Exporter;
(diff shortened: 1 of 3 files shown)


fstab:
Thanks a lot for figuring this out. I am considering to start a 1.0 release branch. We could use this for breaking changes, like refactoring the module and package structure to make sure there's only one module for each package.

What do you think, if we start a 1.0 branch, how should it be structured?

adessaigne:
Hi Fabian, thanks for having a look at this pull request.

Even if I moved around the `Adapter` class I don't think that it can be considered as breaking change as no one should call this class directly.

As for the 1.0 branch I'm not familiar enough with the codebase to make  informed decision but I'll gladly help.

fstab:
Thanks, I merged it.

PR #790, opened 2022-06-01
Mark opentelemtry-api as optional to make it an optional dependency in OSGi

### LINKED ISSUE #678: Few issues around of OSGi-specific headers in manifests

adessaigne:
As the `io.opentelemetry:opentelemetry-api` has the provided scope, a user has to include this dependency to make it work.
As it's not retrieved by default, I marked it as optional in Maven to make it optional in OSGi environments.

This fixes #678

### DIFF
--- a/simpleclient_tracer/simpleclient_tracer_otel/pom.xml
+++ b/simpleclient_tracer/simpleclient_tracer_otel/pom.xml
@@ -23,6 +23,7 @@
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-api</artifactId>
       <scope>provided</scope>
+      <optional>true</optional>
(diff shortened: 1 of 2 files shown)


fstab:
Thanks a lot!

PR #791, opened 2022-06-05
Support `_created` time series suppression

### LINKED ISSUE #774: Add ability to turn off _created metrics
Clone of prometheus/client_python#672

> As per the [June 24th 2021 Prometheus Dev Summit](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#heading=h.omrxayfajpwh) there is consensus that all first party clients should have a parameter to turn off _created metrics.

mindw:
If `PROMETHEUS_DISABLE_CREATED_SERIES` environment variable is true suppress creation of `_created` OpenMetrics time series.

My Java skills are pretty much non-existant, all comments are welcome :)
Fixes #774 

When merging, please consider using `squash-merging`, thanks :)

CC @fstab @SuperQ 

Help needed:
- [x] ~Place to put documentation~
- [ ] Also support setting via property?


dhoard:
@mindw I took your code and refactored it in an effort to maintain existing performance (remove method calls.)

Please take a look: https://github.com/prometheus/client_java/pull/793

CC @fstab @SuperQ

mindw:
@dhoard thank you for suggested code changes - they where added to the PR with a minor twist.

Can you please suggest the best place to document this? would it be the `README.md`? JavaDoc?

It may not be a very good idea exposing any code as public unless its explicitly supposed to be part of the interface.

dhoard:
@mindw We should document the use of the environment variable in the `READMD.md`.

| It may not be a very good idea exposing any code as public unless its explicitly supposed to be part of the interface.

I agree. Typically, the code would use a static initialization block in `Collector` to get the environment variable and set a final `USE_CREATED` variable when the class loads. This is the ideal solution for a configure once scenario. The code change is easy... but I couldn't get the test code working.

mindw:
@dhoard is there anything left preventing this PR form being merged?

dhoard:
@mindw looks good to me. @fstab will need to review and merge.

README.md
```diff
@@ -74,6 +74,13 @@ There are canonical examples defined in the class definition Javadoc of the clie
 Documentation can be found at the [Java Client
 Github Project Page](http://prometheus.github.io/client_java).
 
+### Disabling `_created` metrics
+
+By default, counters, histograms, and summaries export an additional series
```
fstab, line 79: I know this is copy-and-paste from `client_python`. However, I would suggest replacing "By default, ..." with "In OpenMetrics format, ...".

README.md
```diff
@@ -74,6 +74,13 @@ There are canonical examples defined in the class definition Javadoc of the clie
...
+suffixed with `_created` and a value of the unix timestamp for when the metric
+was created. If this information is not helpful, it can be disabled by setting
+the environment variable `PROMETHEUS_DISABLE_CREATED_SERIES=True`.
```
fstab, line 82: I guess Python's `True` has a capital `T`, but as this is Java, I'd use `true` with a lower case `t` in the documentation to avoid confusion. The implementation is case-insensitive anyway, which is good, but i think people are more used to `true`.

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -150,44 +149,43 @@ public MetricFamilySamples filter(Predicate<String> sampleNameFilter) {
      * {@code # HELP}), and as this name <a href="https://github.com/prometheus/common/issues/319">must be unique</a>
      * we include the name without suffix here as well.
      */
-    public String[] getNames() {
+    public List<String> getNames() {
```
fstab, line 152: What about leaving `getNames()` as it is?

The `_created` names in `getNames()` are used to prevent illegal name clashes, like somebody registering a counter named `my_counter_total` but also a gauge named `my_counter_created`. These are a weird corner cases anyway, and I think it's strange to allow this when the `PROMETHEUS_DISABLE_CREATED_SERIES` environment variable is set.

Moreover, if we leave `getNames()` as it was, `PROMETHEUS_DISABLE_CREATED_SERIES` would just enable/disable the `_created` series. If we change `getNames()`, we will allow weird corner cases where an application breaks unless `PROMETHEUS_DISABLE_CREATED_SERIES` is set. I don't think this is the intention.

simpleclient/src/main/java/io/prometheus/client/Collector.java
```diff
@@ -396,4 +394,17 @@ public static String doubleToGoString(double d) {
     }
     return Double.toString(d);
   }
+
+  protected static final String DISABLE_CREATED_SERIES = "PROMETHEUS_DISABLE_CREATED_SERIES";
+  private static final List<String> TRUTHS = Arrays.asList("true", "1", "t");
+
+  protected static boolean USE_CREATED = getUseCreated();
```
fstab, line 401: * Please make it `final`. It's not supposed to change.
* If you make it `final` but still want to test it, you should not make it `static`.
* Rename to `CREATED_SERIES_DISABLED` because that makes it clearer that this is related to the `PROMETHEUS_DISABLE_CREATED_SERIES` environment variable.
* If you revert the changes in `getNames()`, you could move this one level down in the class hierarchy to `SimpleCollector`.

simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java
```diff
@@ -0,0 +1,92 @@
...
+        histogram.observe(100);
+        histogram.observe(200);
+        registry.register(histogram);
+
+        Summary noLabels = Summary.build().name("test_summary").help("test summary").register(registry);
+        noLabels.observe(2);
+
+        List<Collector.MetricFamilySamples> mfsList;
+        try {
+            env.setup();
```
fstab, line 49: If you make the flag `final`, you'll need to call `env.setup()` before initializing the metrics.

simpleclient/pom.xml
```diff
@@ -54,6 +54,24 @@
       <version>4.13.2</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.mockito</groupId>
+          <artifactId>mockito-inline</artifactId>
+          <version>3.12.4</version>
+          <scope>test</scope>
+      </dependency>
```
fstab, line 62: Not needed?

fstab (changes requested):
Thanks for the PR. I added a few comments, nothing major. Please have a look.

README.md
mindw: That may contradict both code and tests. The `_created` metrics seem to be emitted regardless of format (EA both 004 and 100 text formats. For example https://github.com/prometheus/client_java/blob/7c9fc397c21ebc9119f40131f075e1b2ed7b3079/simpleclient_common/src/test/java/io/prometheus/client/exporter/common/TextFormatTest.java#L57 .

README.md
mindw: For the record the code will accept any variation of `true` - even `tRuE`.

simpleclient/src/main/java/io/prometheus/client/Collector.java
mindw: According to https://github.com/prometheus/client_java/issues/774#issuecomment-1133955882 these are actually _not_ forbidden just discouraged. Sure, evil minded  integrators may abuse this by developing their code under `PROMETHEUS_DISABLE_CREATED_SERIES=true` but it will break on the first time it will _not_ be used.

simpleclient/pom.xml
mindw: Alas, it seems to be needed - https://github.com/webcompere/system-stubs#dependencies .

simpleclient/src/main/java/io/prometheus/client/Collector.java
mindw: > * Please make it `final`. It's not supposed to change.
> * If you make it `final` but still want to test it, you should not make it `static`.
Yes, it was final - but then I could not find a way to test it. It I remove the `static` every object will have its own copy of it which seems to be excessive. The above is a compromise.

> * Rename to `CREATED_SERIES_DISABLED` because that makes it clearer that this is related to the `PROMETHEUS_DISABLE_CREATED_SERIES` environment variable.
That will flip it's meaning and will change all code from ```
      if (USE_CREATED) {
```
to
```
      if (!CREATED_SERIES_DISABLED) {
```
Which is a double negative and is considered as a poor programming practice. For example - https://cleankotlin.nl/blog/double-negations#:~:text=Double%20negations%20are%20not%20not,can%20hide%20in%20your%20program.

> * If you revert the changes to `getNames()`, you could move this one level down in the class hierarchy to `SimpleCollector`.
Not sure of the benefit - let's resolve `getNames()` discussion first.

File: simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java (hunk 5 above)
Comment (mindw) [reply simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java:49]: How can I do that?

File: README.md (hunk 1 above)
Comment (fstab) [reply README.md:79]: Oh, sorry, I forgot that `client_java` adds the `_created` series to text format as well.

File: README.md (hunk 2 above)
Comment (fstab) [reply README.md]: Yes, having a case insensitive implementation (and an implementation compatible with Python's) is good.

File: simpleclient/src/main/java/io/prometheus/client/Collector.java (hunk 4 above)
Comment (fstab) [reply simpleclient/src/main/java/io/prometheus/client/Collector.java:401]: I think if the flag should not change at runtime it should be `final`, even if that makes it a bit harder to test. There are a few options how to write a test if it's `final`:
* Make it non-static
* Set the field via reflection (you can remove the `final` modifier via reflection)
* Write an integration test with Testcontainers test (then you could set the environment variable before starting the test application).

Regarding the name of the flag: It's always hard to maintain code if there are different names for the same concept. If we name the environment variable `PROMETHEUS_DISABLE_CREATED_SERIES` but the flag `USE_CREATED` somebody reading the code would always need to double-check if these mean the same thing. If we use a similar name, it's more obvious.

File: simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java (hunk 5 above)
Comment (fstab) [reply simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java:49]: That was referring to
> If you make it final but still want to test it, you should not make it static.

There are a few other options how to write a test though.

File: simpleclient/pom.xml (hunk 6 above)
Comment (fstab) [reply simpleclient/pom.xml:62]: I removed the dependency, and the test works just fine. I don't understand why it is needed.

Anyway, if the `system-stubs` dependency is hard to understand, it might be good not to use it. A few of the options how to make the code testable do not include modifying environment variables at runtime:
* Use reflection to set the flag
* Write an integration test with Testcontainers.

Comment (mindw) [reply simpleclient/pom.xml:62]: Yes, that is the expected output for JVM 11. 

Quoting from the link above:

> ⚠ WARNING: JDK Compatibility. From JDK16 onwards, there are deeper restrictons on the ability to use reflection. Previous versions of this library, and others in the space, encounter an Illegal Reflective Access warning, or even a runtime error such as java.lang.reflect.InaccessibleObjectException when trying to manipulate the Map behind the system's environment variables.

Consequently, this library now uses mockito-inline version 3.x to enable the interception of calls for reading environment variables. This requires consumers to both use a compatible version of Mockito AND be prepared for the inline implementation of Mockito mocks.

Note: Groovy users may need Mockito >= 4.5.0 for compatibility.

File: simpleclient/src/main/java/io/prometheus/client/Collector.java (hunk 4 above)
Comment (mindw) [reply simpleclient/src/main/java/io/prometheus/client/Collector.java:401]: > I think if the flag should not change at runtime it should be `final`, even if that makes it a bit harder to test. There are a few options how to write a test if it's `final`:
> 
> * Make it non-static
> * Set the field via reflection (you can remove the `final` modifier via reflection)
> * Write an integration test with Testcontainers test (then you could set the environment variable before starting the test application).

I'm not sure what I can add. The extra variable per object seems like too much a price to pay given the alternative. 

Can you please provide some guidance on how to use `Testcontainers` here? I'm unfamiliar with that framework.
Otherwise it seems like a good way on making the variable final (which something I wanted to do in the first place). It will also make the new dependency redundant.

> Regarding the name of the flag: It's always hard to maintain code if there are different names for the same concept. If we name the environment variable `PROMETHEUS_DISABLE_CREATED_SERIES` but the flag `USE_CREATED` somebody reading the code would always need to double-check if these mean the same thing. If we use a similar name, it's more obvious.

Thanks for pointing this rule out. We're talking about readability multiple places vs readability one place. For me the many wins - the condition should be read as "TRUE" instead of "NOT NOT". Finally, the this choice mirrors the python one. The environment variable is mentioned once in a function and is pretty easy to lookup in any decent IDE.

File: simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java (hunk 5 above)
Comment (mindw) [reply simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java:49]: As noted above - some guidance/example on how to use the framework would be appreciated.

Comment (fstab) [reply simpleclient/src/test/java/io/prometheus/client/DisableCreatedTest.java:49]: Have a look at `JavaVersionsIT`. I guess for your test you won't need different Java versions, but apart from that it should be similar.

File: simpleclient/src/main/java/io/prometheus/client/Collector.java (hunk 3 above)
Comment (mindw) [reply simpleclient/src/main/java/io/prometheus/client/Collector.java]: Fair enough - I've reverted the this change. Thanks for taking the time and effort explaining it!

**Comment by fstab** [comment]
@mindw I think I'll do a release this week, and it would be good to include this feature. What do you think about adding the feature now (and make the flag `final`), and continue with the test later in a separate PR? That way the feature would make it into the release, and we'd have more time to get the test done.

**Comment by mindw** [comment]
> What do you think about adding the feature now (and make the flag final), and continue with the test later in a separate PR? That way the feature would make it into the release, and we'd have more time to get the test done.

No problem - tests will wait ill the weekend

**Comment by fstab** [comment]
Thanks a lot @mindw. One last thing: I'm wondering whether `Collector` is the right place for reading the environment variable. Initially I was thinking to move it up to `SimpleCollector`, because it's only relevant for `Counter`, `Histogram`, and `Summary`, and not for other collectors.

However, thinking about it I feel that it might be worthwhile to move the environment variable parsing into its own class, like this:

```java
class EnvironmentVariables {

    private static final String DISABLE_CREATED_SERIES = "PROMETHEUS_DISABLE_CREATED_SERIES";
    private static final List<String> TRUTHS = Arrays.asList("true", "1", "t");
    private static final boolean createdSeriesEnabled = !getBoolean(DISABLE_CREATED_SERIES, true);

    static boolean isCreatedSeriesEnabled() {
        return createdSeriesEnabled;
    }

    private static boolean getBoolean(String envVarName, boolean defaultValue) {
        String stringValue = System.getenv(DISABLE_CREATED_SERIES);
        if (stringValue != null) {
            return !TRUTHS.contains(stringValue.toLowerCase());
        }
        return defaultValue;
    }
}
```

Then the `Histogram`, `Counter`, and `Summary` could use it like this:

```java
if (EnvironmentVariables.isCreatedSeriesEnabled()) {
    // add _created sample
}
```
I feel this would help with separation of concerns. What do you think?

mindw:
@fstab It is done :) It seems to me that both suggestions converged.
Cheers!

fstab:
:+1: thanks!

PR #795, opened 2022-06-16
Fixed code to correctly unregister a Collector that is not registered

dhoard:
Fixed code to correctly unregister a Collector that is not registered.


dhoard:
@fstab please review. Thanks!

fstab (approved):
Thanks a lot for finding this, and for fixing.

PR #798, opened 2022-07-05
Added CompilationMXBean compilation time metric

dhoard:
Added CompilationMXBean compilation time metric in milliseconds.
Reorded collector registration in `DefaultExports`.

@fstab @rainerjung

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -254,4 +254,22 @@ public class CollectorRegistry {
     return null;
   }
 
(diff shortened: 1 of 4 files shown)


fstab (commented):
Thanks a lot! It's looking good. The only thing I'm wondering is if we should convert it to seconds, and call it `jvm_compilation_time_seconds_total` instead. The [Prometheus best practices](https://prometheus.io/docs/practices/naming/#base-units) say that base units should be used, and the base unit for time is seconds.

dhoard:
@fstab code changed from milliseconds to seconds. please review.

fstab:
Thanks a lot @dhoard !

PR #799, opened 2022-07-07
Fixed code to handle the scenario where help is set to `null`

dhoard:
Fixed code to handle the scenario where help is set to `null`. Added JUnit tests to validate correct behavior.

A user may think they don't want/need help, and thus set help to `null`. The original code allows a `null` value when creating metrics (Gauge, etc..), but results in a `NullPointerException` in `TextWriter.writeEscapedHelp()` when writing the metric value.

If help is `null`, an `IllegalStateException` should be thrown since a non-null/non-empty value is required.

@fstab


dhoard:
Closing/refactoring since `help` is not required.

PR #802, opened 2022-07-08
SubstrateVM awareness

### LINKED ISSUE #801: Make hotspot collector SubstrateVM aware
This will allow to use it in GraalVM native image builds. The problem is with Thread metrics, because `SubstrateThreadMXBean`, which implements `com.sun.management.ThreadMXBean` in the SubstrateVM of GraalVM implements a number of methods throwing unrecoverable VM Error:
https://github.com/oracle/graal/blob/1221b291b4f1fbc3f05bd7f850d7591cecce85ca/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateThreadMXBean.java#L146

arixmkii:
Fixes #801 

Group items, which are not supported in GraalVM inside conditional branch, which checks that class name is not a known of GraalVM.

### DIFF
--- a/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/NativeImageChecker.java
+++ b/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/NativeImageChecker.java
@@ -1,0 +1,13 @@
+package io.prometheus.client.hotspot;
+
+/**
+ * Contains utilities to check if we are running inside or building for native image. Default behavior is to check
+ * if specific for graalvm runtime property is present. For additional optimizations it is possible to do add
+ * "--initialize-at-build-time=io.prometheus.client.hotspot.NativeImageChecker" to graalvm native image build command and
+ * the native image will be identified during build time.
+ */
+public final class NativeImageChecker {
+    static final boolean isGraalVmNativeImage = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
+
+    private NativeImageChecker() {}
+}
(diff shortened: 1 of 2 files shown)


fstab (changes requested):
Thanks a lot! I was wondering if there is a more standard way to detect if code is running in a GraalVM native image, and I came across GraalVM's [ImageInfo](https://github.com/oracle/graal/blob/master/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/ImageInfo.java).

The JavaDoc reads like you could do something like this:

```java
static final boolean isGraalVmNativeImage = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
```
Does that work? It would be more stable API then relying on the class name.

arixmkii:
I will check if this works, but I have some thoughts now on a bit different topic.

@fstab do you think it might become a feature to implement independently disabled properties with systemProps like reading `System.getProperty("prometheus.java.hostspot.thead-deadlocks.disabled")` and disabling this specific metrics if they introduces issues. This would allow to support not only SubstrateVM via system config, but any arbitrary runtime. I believe that https://www.graalvm.org/22.0/reference-manual/java-on-truffle/ is one of those, which might introduce specific requirements.

arixmkii:
Factored a utility class, which could be used for NativeImage checks. Decided to created a utility class to provide flexibility of build time initialization for this as I see no issues it could cause, but I believe that optimizing compiler in this case could manage to eliminate dead branches if conditional blocks. I haven't verified this behavior in resulted native image.

fstab (approved):

fstab:
Thanks a lot, I merged it.

To your question regarding disabling this specific metrics: You can do this using the sample name filter. However, configuration is not part of the core library, it's up to the exporter to configure this. Examples:

* `HTTPServer`: You can configure the filter using `HTTPServer.Builder.withSampleNameFilter()`.
* `MetricsServlet`: You can configure the filter using `ServletConfig` (web.xml, or other ways to configure servlets).
* `jmx_exporter`: Filters will be configurable in the new config YAML format we are currently working on.

PR #805, opened 2022-08-10
Add exemplars support for Micrometer Tracing

jonatan-ivanov:
This PR adds exemplars support for [Micrometer Tracing](https://github.com/micrometer-metrics/tracing).
Micrometer Tracing is the successor of [Spring Cloud Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth) in terms of providing an abstraction layer over tracing libraries (i.e.: Brave and OTel) and it provides distributed tracing support for the Spring portfolio (Spring Boot 3.x, Spring Framework 6.x, etc).

There is one thing missing from this PR where I would like to ask for some guidance from the maintainers: there is no static configuration. Since Micrometer Tracing needs a `Tracer` instance and it supports both Brave + OTel (Brave does not have static config) this might be problematic since users can use their own `Tracer` instance that might be different than the Prometheus client is aware of.

@fstab @tomwilkie


jonatan-ivanov:
Also, do you happen to know why the javadoc goal is failing? The error message does not seem to be very helpful.

integration_tests/it_exemplars_micrometer_tracing/src/main/java/io/prometheus/client/it/exemplars_micrometer_tracing/ExampleApplication.java
```diff
@@ -0,0 +1,53 @@
...
+ * Example application using Micrometer Tracing.
+ */
+public class ExampleApplication {
+
+  public static void main(String[] args) throws IOException, InterruptedException {
+    SimpleTracer tracer = new SimpleTracer();
+    ExemplarSampler exemplarSampler = new DefaultExemplarSampler(new MicrometerTracingSpanContextSupplier(tracer));
+    ExemplarConfig.setCounterExemplarSampler(exemplarSampler);
+    ExemplarConfig.setHistogramExemplarSampler(exemplarSampler);
+    ExemplarConfig.enableExemplars();
```
fstab, line 26: It should not be necessary to set the `ExemplarSampler` explicitly. The idea is that `client_java` supports Exemplars automatically out-of-the-box if a supported tracer is found.

The magic happens in the `Tracer.findSpanContextSupplier()` in the `simpleclient` module. It checks if `OpenTelemetrySpanContextSupplier.isAvailable()`, then if `OpenTelemetryAgentSpanContextSupplier.isAvailable()`.

Please copy-and-paste these lines and add a check if `MicrometerTracingSpanContextSupplier.isAvailable()`.

As you are the first contributor adding Exemplars support, please provide feedback. Here's some context on the current implementation. Please let me know if this makes sense, or if there is a better or simpler way to do it:

Requirements:

1. Exemplars work out-of-the-box if a supported tracer is detected. For example, if you attach the OpenTelemetry Java agent to an application using `client_java`, you will automatically get Exemplars without changing your code. In the same way, if you add Micrometer tracing to an application using `client_java` you should automatically get Exemplars without any code change.
2. It should not be possible to break an application by adding `simpleclient` as a dependency.

Achieving (2) was easy as long as `simpleclient` was just a metrics library without dependencies. However, to implement Exemplar support we had to add dependencies like `opentelemetry-api`. This can be problematic if an application ships with another version of `opentelemetry-api`.

We took two steps to avoid potential issues:
* The `opentelemetry-api` dependency is `provided` to make sure we use the tracer that ships with the application at runtime.
* If there are incompatible API changes between different `opentelemetry-api` versions there might be runtime errors when using different versions at runtime and at compile time. As a last resort, we allow users to exclude the `simpleclient_tracer_otel` dependency from `simpleclient` to get rid of OTel exemplars for good. That's why we catch the `NoClassDefFoundError` in `findSpanContextSupplier()`. There's even an integration test removing the dependency and making sure that `simpleclient` still works.

Anyway, if you have any thoughts on this let me know.

fstab (changes requested):
Sorry again for dropping the ball. Thanks a lot for adding Micrometer tracer support for Examplars.

I will take another look at the initialization, but I think if there's no static variable to get the tracer from users will need to set the tracer explicitly as in your example.

fstab:
Heads-up: I renamed `master` to `main`.

jonatan-ivanov:
javadoc: 🤦🏼 🤣 I'm sorry.

static initialization
I'm not 100% sure I understand this:
>Is there any chance to create a class that would be detected by Spring or CDI so that we can get the Micrometer tracer injected?

Spring and other CDI solutions will initialize their context after class loading (by creating non-static instances) so even if they inject something to a static context, that will happen (or can easily happen without control) after the java client's static initializers were executed. Also, since they can (and will) inject the registry, the global/static components are less useful in those environments (in fact they can cause trouble and make testing hard/impossible).

This is my honest feedback, I don't want to offend anyone or step on any toes (also, I'm biased): I think the Prometheus Java Client should focus on letting CDI/users configure things in non-static ways (providing builders with defaults and factory methods are completely ok). I think it could also use a `ServiceLoader` for static access. I assume that this ship has sailed (and might not be the scope of this PR) so I guess the options are:

1. Keep this as-is saying that users will not get static initializer support for Micrometer Tracing they should do this manually or their framework should do this for them.
(Right now, this is enough for me from Micrometer/Micrometer Tracing/Spring side.)

2. Only support OTel: this means handling two cases: 1. only OTel, 2. Micrometer Tracing (with OTel bridge) + OTel 3. Neither.

3. Support everything (this is basically reimplementing Spring Boot auto-configuration or other auto-configuration in other frameworks), the cases are:
   1. OTel only
   1. Micrometer Tracing (with OTel bridge) + OTel
   1. Micrometer Tracing (with Brave bridge) + Brave
   1. None of the above
   1. Lots of invalid cases like Micrometer Tracing with OTel bridge + Brave/something else/nothing (or Micrometer Tracing with OTel bridge and Brave bridge together)
   1. If Micrometer Tracing adds support for another Tracing library or Prometheus Client adds support for another Tracing library (i.e.: Brave) or both, this will increase quickly

PR #808, opened 2022-08-23
Make some fields public and add children()/labelNames() methods in SimpleCollector

tjiuming:
## Motivation:
 Make some inner fields/methods accessible to provide an optional `collect` way.

### DIFF
--- a/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
+++ b/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java
@@ -101,7 +101,7 @@ public class CollectorRegistry {
   /**
    * A snapshot of the current collectors.
    */
-  private Set<Collector> collectors() {
+  public Set<Collector> collectors() {
     synchronized (namesCollectorsLock) {
       return new HashSet<Collector>(collectorsToNames.keySet());
     }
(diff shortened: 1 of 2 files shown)


fstab:
Thanks a lot!

I'm always a bit hesitant to make API public, because that means we will not be able to change this API in the future without breaking changes.

Could you elaborate a bit on the use cases you have in mind? The description in the PR is currently very terse, and I'm not sure what problem this is trying to solve.

tjiuming:
@fstab It's a minor PR to https://github.com/prometheus/client_java/pull/788, and will not comes a lot of changes. For breaking changes, I believe it will not happen for a long time, and `Counter/Gauge/Summary/Histogram`'s Child is also PUBLIC

dhoard:
I feel making `Set<Collector>` in `CollectorRegistry` public is valid and would allow an alternative way to collect metrics.

Assuming that change is accepted, you can write your own custom versions of `Gauge`, etc. which have methods that allow access to the raw data. You will have to do some reflection based inspection / Object casting to use the methods... but that would be required if you expose more internals of `SimpleCollector`

Exposing more internals of `SimpleCollector` seems like a bad idea.

Not all collectors extend `SimpleClient` and it breaks the spirit of the contract for working with a `Collector` from a `CollectorRegistry` point of view.

tjiuming:
@fstab @dhoard Thanks for your reply and I'll close the PR, we can use UNSAFE/REFLECTION to access the internal fields to prevent generate toooooo many samples instances.

Issue #828, opened 2022-12-11
[flaky test] Flaky test in CustomMappingSampleBuilderTest


tt0suzy:
`io.prometheus.client.dropwizard.samplebuilder.CustomMappingSampleBuilderTest.test_WHEN_MoreToFormatInLabelsAndName_THEN_ShouldReturnCorrectSample`
This is a [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html). It can pass mvn test but when run using the tool [NonDex](https://github.com/TestingResearchIllinois/NonDex), it fails. NonDex is a tool that will introduce non-determinism in certain java collections. Here is the test result:
```
-------------------------------------------------------------------------------
Test set: io.prometheus.client.dropwizard.samplebuilder.CustomMappingSampleBuilderTest
-------------------------------------------------------------------------------
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.011 sec <<< FAILURE!
test_WHEN_MoreToFormatInLabelsAndName_THEN_ShouldReturnCorrectSample(io.prometheus.client.dropwizard.samplebuilder.CustomMappingSampleBuilderTest)  Time elapsed: 0.004 sec  <<< FAILURE!
java.lang.AssertionError: expected:<Name: app_okhttpclient_client_HttpClient_greatService LabelNames: [service, status] labelValues: [greatService_400, s_400] Value: 1.0 TimestampMs: null> but was:<Name: app_okhttpclient_client_HttpClient_greatService LabelNames: [status, service] labelValues: [s_400, greatService_400] Value: 1.0 TimestampMs: null>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at io.prometheus.client.dropwizard.samplebuilder.CustomMappingSampleBuilderTest.test_WHEN_MoreToFormatInLabelsAndName_THEN_ShouldReturnCorrectSample(CustomMappingSampleBuilderTest.java:129)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #829, opened 2022-12-11
fixed flaky test

tt0suzy:
Fixed flaky test in [828](https://github.com/prometheus/client_java/issues/828)

The cause of failure is because the test used a `HashMap` to store `labels`, however, as per [Oracle's documentation](https://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html):

> The HashMap class is roughly equivalent to Hashtable, except that it is unsynchronized and permits nulls. This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.

So the order of labels in the `Collector.MetricFamilySamples.Sample `object is non-deterministic. It can be fixed by replacing `HashMap` with `LinkedHashMap`, such that the order of entries is deterministic.


fstab:
Nice, thanks a lot!

Issue #664, opened 2021-05-21
tag format for Graphite client


liukun4515:
If we want send the tagged carbon format.
In the doc, https://graphite.readthedocs.io/en/latest/tags.html

We should use the ';' to split the metric name, tags values.

But the graphite client is error.

fstab:
Looks like the Graphite Tag Support was added with Graphite release 1.1.X (end of 2017) while the `simpleclient_graphite_bridge` was implemented beginning of 2016. So no wonder that this is missing.

I will put this on my todo list, but it will take some time because I need to find the time to set up Graphite and try it out. If you want to provide a pull request please create one.

liukun4515:
@fstab can i take this issue and file a pull request.

PR #833, opened 2022-12-22
update the tag format for grahite client

liukun4515:
Signed-off-by: liukun4515 <liukun@apache.org>

Closes https://github.com/prometheus/client_java/issues/664


liukun4515:
@fstab PTAL

fstab (approved):
Thanks a lot!

fstab (changes requested):
Sorry, I approved too early. There's a `GraphiteTest` that's failing after your change. I assume the test is wrong, as it expects output `labels.l.fo_o` but after your change the output is `labels;l=fo_o`. Could you adapt the test, and signoff? Thanks

liukun4515:
> Sorry, I approved too early. There's a `GraphiteTest` that's failing after your change. I assume the test is wrong, as it expects output `labels.l.fo_o` but after your change the output is `labels;l=fo_o`. Could you adapt the test, and signoff? Thanks

Thanks, the test has been resolved 
cc @fstab

fstab:
Thanks a lot!

PR #842, opened 2023-05-13
New data model

fstab:
**Context**

The current implementation of the simpleclient metric library makes some assumptions that are tightly coupled to OpenMetrics text format. For example, the implementation assumes that each `Sample` has exactly one `double value`.

With the advent of the new native histograms some of these core assumptions don't hold anymore: Native histograms require the Protobuf exposition format, and a Sample of a native histogram has a complex type as a value.

Rather than trying to fit native histograms into the existing model, we are proposing a new model with this PR.

**What This PR Does**

This PR adds three new Maven modules

* `prometheus-metrics-model`: A data model for immutable snapshots of Prometheus metrics. Metric libraries will produce these snapshots when the Prometheus server scrapes. In a future PR, we will add a metrics library, this PR is just the immutable snapshot model. The goal is that this model will not only be used by `client_java`, but will be used by 3rd party metrics library as well if they want to expose metrics in Prometheus format.
* `prometheus-exposition-formats`: Converts snapshots from the `prometheus-metrics-model` to various exposition formats: Prometheus text format, OpenMetricds text format, Prometheus Protobuf format. We have plans to add support for OpenMetrics Protobuf format and for OpenTelemetry Metrics Protobuf format in the future.
* `protobuf-shaded`: Dependencies to Google's Protobuf library may cause issues if an application uses another version of that library. To avoid issues, the `protobuf-shaded` module generates a relocated version of Google's Protobuf library under a package named `io.prometheus.com_google_protobuf_3_21_7`.

**How to Build**

You need to install the `protobuf-shaded` module to your local Maven repository, otherwise there will be compile errors in `prometheus-exposition-formats`:

```
cd protobuf-shaded
../mvnw install
cd ..
```

Once `protobuf-shaded` is available in your local Maven repository, you can just build `client_java` as usual (requires Java 11, higher or lower Java versions won't work):

```
./mvnw clean verify
```

The manual install of `protobuf-shaded` is necessary because we didn't release `protobuf-shaded` to Maven central yet. Once we released it, Maven will download it automatically and the manual step won't be necessary anymore.

**Next Step**

We are opening this PR now to get feedback on the data model. There's no need to merge it until we have implemented a metrics library using the new model.


fstab:
New data model has been merged to the [1.0.x](https://github.com/prometheus/client_java/tree/1.0.x) branch and released with the `1.0.0-alpha-x` [releases](https://github.com/prometheus/client_java/releases). Closing this PR.

PR #846, opened 2023-06-29
Fix "expected label name, got \"BCLOSE\"" error

wangkaish:
My prometheus(2.33.4) request metrics got "expected label name, got \"BCLOSE\"", I try update response content `{a="1", }` to `{a="1"}` then the problem fixed.


fstab:
Thanks for the PR. I don't think the trailing comma was the root cause of your issue.
* The data you see in your browser is not the data the Prometheus server scrapes. Your browser uses Text format, while the Prometheus server scrapes OpenMetrics format. To view OpenMetrics format (assuming your application exposes metrics on port 9000) run `curl -H 'Accept: application/openmetrics-text' localhost:8080/metrics`.
* Text format explicitly supports trailing commas, so if your Prometheus server falls back to text format for some reason it will ignore trailing commas.

Anyway, there have been a lot of discussions around this, and I think the current implementation won't change. However, there is work on a re-implementation of the metrics library which won't have the trailing commas, it's currently being developed on the `new-metrics-library` branch.

wangkaish:
Hi @fstab, You are right, the real reason is not more comma at the end of the value.

I updated Spring 2.5.1 to 2.7.11 and then Prometheus started prompting BCLOSE, The Tomcat versions used by these two Spring versions are 9.0.46 and 9.0.74, the response Content-Type is `text/plain; version=0.0.4;charset=utf-8` and `text/plain;version=0.0.4;charset=utf-8`, in Tomcat 9.0.47 0 an issue (https://bz.apache.org/bugzilla/show_bug.cgi?id=62912) is to remove the space in the Content-Type value, I don't know if this affects the Prometheus here.

PR #856, opened 2023-08-05
Changed to remove use of inefficient String method

dhoard:


simpleclient_common/src/main/java/io/prometheus/client/exporter/common/TextFormat.java
```diff
@@ -137,8 +137,9 @@ public static void write004(Writer writer, Enumeration<Collector.MetricFamilySam
   }
 
   private static void writeEscapedHelp(Writer writer, String s) throws IOException {
-    for (int i = 0; i < s.length(); i++) {
-      char c = s.charAt(i);
+    char[] characters = s.toCharArray();
+    for (int i = 0; i < characters.length; i++) {
+      char c = characters[i];
```
fstab, line 142: I'm not sure if this is more efficient than `s.charAt(i)`. The `charAt()` method does nothing more than returning `characters[i]`, and it does not require copying an array with `s.toCharArray()`.

dhoard: This was based on some YourKit profiling but could be a red herring.

I don't feel strongly either way - more than happy to create a PR to revert the change.

fstab: Interesting that this turned up during profiling. Let's leave the PR merged for now, but before we port this to the new [prometheus-metrics-exposition-formats](https://github.com/prometheus/client_java/blob/1.0.x/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/TextFormatUtil.java#L41-L41) module on the 1.0.x branch I'd like to write a small JMH micro benchmark just to be sure.

PR #859, opened 2023-09-04
Supporto for multi target pattern in java

ganzuoni:
To supporto prometheus multi target pattern it is necessary for the callectors to have access to the original http request receive by the endpoint. 
I've added a common wrapper to http and servlet endpoints requests passed to collectors.
Backward compatibility it is preserved.

@fstab

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -2,7 +2,7 @@
 title: "client_java"
 ---
 
-This is the documentation for the [Prometheus Java client library](https://github.com/prometheus/client_java) verion 1.0.0 and higher.
+This is the documentation for the [Prometheus Java client library](https://github.com/prometheus/client_java) version 1.0.0 and higher.
 
 The main new features of the 1.0.0 release are:
 


fstab:
Hi @ganzuoni, sorry for getting back to this so late, I must have missed this while we were busy preparing 1.0.0.

Anyway, I read [Understanding and using the multi-target exporter pattern](https://prometheus.io/docs/guides/multi-target-exporter/), and the doc says:

> This pattern is only used for certain exporters, such as the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).

In both cases (blackbox and SNMP) the goal of multi target requests is to have the exporter run a test against a target:

* For the Blackbox exporter there's an example of running an HTTP GET request to test if a system is up. This uses blackbox exporter's `http_2xx` module.
* For the SNMP exporter they say multi target requests can be used for querying network gear speaking SNMP.

I'm wondering if `client_java` should implement this. `client_java` does not have any built-in tests that it could run against a target.

Do you have any specific use case in mind?

ganzuoni:
Hi Fabian,
the SNMP exporter is a proxy (ignore protocol adapter function because is
not relevant) that allows a single installation to be referenced as
different exporters
In my use case I have a custom Collector that fetches JSON data from remote
storage appliances (Infinidat, Hitachi) and generates prometheus metrics.
In case of multiple identical appliances I have to duplicate remote target
definitions just to list different endpoints and there is the need to
reconfigure both the proxy and Prometheus in case of target adding/removal.
I don't understand exactly your concern about built-in tests...I've added
some tests

fstab:
Thanks @ganzuoni for the example with fetching data from JSON sources. That's a valid use case and it makes sense.

There has been a bit of refactoring with the 1.0.0 release. Would you port your PR to the current `main` branch?

Here's what changed:

* The method `CollectorRegistry.metricFamilySamples()` has been moved to `PrometheusRegistry.scrape()`. Also, the method changed it's return type: It returns an explicit data model rather than generic `MetricFamilySamples`, but that should not affect your PR.
* There is a common `PrometheusScrapeHandler` to handle HTTP scrape request. This is used by the `HTTPServer` and also by the `PrometheusMetricsServlet`. This is nice, because if you add support for the `target` query parameter there it will work for both exporters.
* We dropped Java 6 support and have Java 8 as baseline.
* Docs are not in `README.md` anymore, but in `./docs/content/`, see `./docs/README.md`.

As we are on Java 8 now, we can use interfaces with default methods :tada:. So instead of `supportsCollectorScrapingContext()` we can just add default methods to `Collector` and `MultiCollector`, something like:

```java
default MetricSnapshots collect(ScrapeReqeust request) {
    // ...
}

default MetricSnapshots collect(Predicate<String> includedNames, ScrapeReqeust request) {
    // ...
}
```

The default implementation should ignore the `request` parameter.

Little nitpick: I feel the name `CollectorScrapingContext` is too generic because context can be anything, so I'd go with a name like `ScrapeRequest`. There are two query parameters that are hanled in the exporter, which is `name[]` and `debug`. Maybe we can strip these and not put them in `ScrapeRequest`, because they are not supposed to be used by collectors.

What do you think?

PR #873, opened 2023-10-12
Support multi-target pattern

ganzuoni:
Porting of the same implementation to main branch


prometheus-metrics-exporter-common/src/main/java/io/prometheus/metrics/exporter/common/PrometheusHttpScrapeRequest.java
```diff
@@ -0,0 +1,24 @@
+package io.prometheus.metrics.exporter.common;
+
+import io.prometheus.metrics.model.registry.PrometheusScrapeRequest;
+
+public class PrometheusHttpScrapeRequest implements PrometheusScrapeRequest {
```
fstab, line 5: We could use `PrometheusHttpRequest` directly and remove `PrometheusHttpScrapeRequest` and `PrometheusScrapeRequest`.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/Collector.java
```diff
@@ -19,6 +19,15 @@ public interface Collector {
      */
     MetricSnapshot collect();
 
+    /**
+     * Provides Collector with the details of the request issued by Prometheus to allow multi-target pattern implementation
+     * Override to implement request dependent logic to provide MetricSnapshot
+     */
+	default MetricSnapshot collect(PrometheusScrapeRequest scrapeRequest) {
+		MetricSnapshot result = collect();
```
fstab, line 27: Can simplify to `return collect();`

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/Collector.java
```diff
@@ -33,6 +42,21 @@ default MetricSnapshot collect(Predicate<String> includedNames) {
         }
     }
 
+    /**
+     * Like {@link #collect(includedNames)}, but with support for multi-target pattern.
```
fstab, line 46: Should be `{@link #collect(Predicate)}`

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/ExtendedCollector.java
```diff
@@ -0,0 +1,30 @@
+package io.prometheus.metrics.model.registry;
+
+import io.prometheus.metrics.model.snapshots.MetricSnapshot;
+
+public abstract class ExtendedCollector implements Collector {
```
fstab, line 5: I think this can be removed and users just extend `Collector` directly.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/ExtendedMultiCollector.java
```diff
@@ -0,0 +1,25 @@
+package io.prometheus.metrics.model.registry;
+
+import io.prometheus.metrics.model.snapshots.MetricSnapshots;
+
+public abstract class ExtendedMultiCollector implements MultiCollector {
```
fstab, line 5: I think this can be removed and users just extend `MultiCollector` directly.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/MultiCollector.java
```diff
@@ -18,6 +18,16 @@ public interface MultiCollector {
      */
     MetricSnapshots collect();
 
+    /**
+     * Provides Collector with the details of the request issued by Prometheus to allow multi-target pattern implementation
+     * Override to implement request dependent logic to provide MetricSnapshot
+     */
+	default MetricSnapshots collect(PrometheusScrapeRequest scrapeRequest) {
+		MetricSnapshots result = collect();
```
fstab, line 26: Can be simplified to `return collect();`

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/MultiCollector.java
```diff
@@ -34,6 +44,23 @@ default MetricSnapshots collect(Predicate<String> includedNames) {
         return result.build();
     }
 
+    /**
+     * Like {@link #collect(includedNames)}, but with support for multi-target pattern.
```
fstab, line 48: Should be `{@link #collect(Predicate)}`

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/MultiCollector.java
```diff
@@ -34,6 +44,23 @@ default MetricSnapshots collect(Predicate<String> includedNames) {
...
+     */
+    default MetricSnapshots collect(Predicate<String> includedNames, PrometheusScrapeRequest scrapeRequest) {
+    	MetricSnapshots allSnapshots = collect(scrapeRequest);
+        MetricSnapshots.Builder result = MetricSnapshots.builder();
+        for (MetricSnapshot snapshot : allSnapshots) {
+            if (includedNames.test(snapshot.getMetadata().getPrometheusName())) {
+                result.metricSnapshot(snapshot);
+            }
+        }
+        return result.build();
```
fstab, line 60: Only the first line differs from the `collect()` method without the `scrapeRequest` parameter. What about creating a helper method for the common code?

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
```diff
@@ -72,6 +72,29 @@ public MetricSnapshots scrape() {
         return result.build();
     }
 
+    public MetricSnapshots scrape(PrometheusScrapeRequest scrapeRequest) {
```
fstab, line 75: There's a lot of duplicate code in `PrometheusRegistry` now. Maybe it makes sense to implement the existing `scrape()` methods like this:

```java
    public MetricSnapshots scrape() {
        return scrape((PrometheusScrapeRequest) null);
    }

    public MetricSnapshots scrape(Predicate<String> includedNames) {
        return scrape(includedNames, null);
    }
```

And refactor the implementations of the new `scrape()` methods to call `collect()` like this:

```java
scrapeRequest == null ? collector.collect() : collector.collect(scrapeRequest);
```

fstab (changes requested):
Hi @ganzuoni, thanks a lot for the PR! I added a few comments, nothing major, mostly refactoring.

BTW while reviewing it I found a few shortcomings in the current implementation and merged #875 and #876. Looks like multi target support is the first time that we look closely at using the `Collector` interface directly. Thanks a lot for coming up with that use case!

Please rebase to the current `main` so that you get the latest changes.

examples/example-exporter-httpserver/src/main/java/io/prometheus/metrics/examples/httpserver/SampleExtendedCollector.java
```diff
@@ -0,0 +1,35 @@
+package io.prometheus.metrics.examples.httpserver;
+
+import io.prometheus.metrics.core.metrics.Counter;
+import io.prometheus.metrics.model.registry.ExtendedCollector;
+import io.prometheus.metrics.model.registry.PrometheusScrapeRequest;
+import io.prometheus.metrics.model.snapshots.MetricSnapshot;
+
+public class SampleExtendedCollector extends ExtendedCollector {
```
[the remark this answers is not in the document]
ganzuoni: Agree

prometheus-metrics-exporter-common/src/main/java/io/prometheus/metrics/exporter/common/PrometheusHttpRequest.java
```diff
@@ -7,6 +7,11 @@
 
 public interface PrometheusHttpRequest {
 
+    /**
+     * Return the absolute path of a Http Request
+     */
+    String getRequestPath();
```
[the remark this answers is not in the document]
ganzuoni: Yes, I know, anyway adapter implementation takes care of what the wrapped returns as context, servlet path and pathInfo


    _public static class Request implements PrometheusHttpRequest {

        private final HttpServletRequest request;

        public Request(HttpServletRequest request) {
            this.request = request;
        }

        @Override
		public String getRequestPath() {
        	StringBuffer uri = new StringBuffer();
			String contextPath = request.getContextPath();
        	if (contextPath.startsWith("/")) {
        		uri.append(contextPath);
        	}
			String servletPath = request.getServletPath();
			if (servletPath.startsWith("/")) {
        		uri.append(servletPath);
			}
			String pathInfo = request.getPathInfo();
			if (pathInfo != null) {
				uri.append(pathInfo);
			}
			return uri.toString();
		}_

prometheus-metrics-exporter-common/src/main/java/io/prometheus/metrics/exporter/common/PrometheusHttpScrapeRequest.java
fstab: Ah, ok, I see. What about `interface PrometheusHttpRequest extends PrometheusScrapeRequest`? Then we could remove `PrometheusHttpScrapeRequest`.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/ExtendedMultiCollector.java
fstab: What about removing `ExtendedMultiCollector` and adding all methods to the `MultiCollector` interface? Same as you did for `Collector`?

fstab (changes requested):
Thanks for the update @ganzuoni, I re-opened three conversations, mostly with things I think can be removed :).

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/ExtendedMultiCollector.java
ganzuoni: Well, it is just to simplify specific implementation. Beware that collectMetricSnapshots is a protected method.
Anyway, default implementation of MultiCollector.collect(PrometheusScrapeRequest scrapeRequest) simply ignores the scrapeRequest, while ExtendeMultiCollector changes this behaviour forcing the implementation of collectMetricSnapshots

fstab: Thanks @ganzuoni. I understand the argument that by making `ExtendeMultiCollector` enforces an implementation of `collectMetricSnapshots()`. As always in software engineering, it's a trade-off. I think keeping the API simpler (no additional abstract base classes) is the better trade-off in that case.

* It's not clear that returning an empty result is always the best default behavior for `collect()`. For example, in some cases throwing an Exception to trigger an HTTP 500 response might be a good idea, because then the target shows as `failed` in the Prometheus server, and users see directly that there's something wrong. If we stick with the `MultiCollector` interface users must explicitly implement a `collect()` method, and I think this is better than providing a default implementation in the core API.
* `Collector` and `MultiCollector` are the same except for the return type. So, if it is a good idea to have an `ExtendeMultiCollector` in addition to `MultiCollector`, it will also be a good idea to have an `ExtendeCollector` in addition to `Collector`. If we really follow through with this, the core API gets two new base classes. I think it's not worth the complexity for the multi target use case.

fstab (changes requested):
Thanks a lot @ganzuoni. We are almost there, just one more comment on the `ExtendedMultiCollector` thread.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/ExtendedMultiCollector.java
ganzuoni: Agree

fstab (approved):
Awesome, thanks a lot!!!

Issue #871, opened 2023-10-07
JPMS support


solomax:
Hello,

Would it be possible to release version with `Automati-Module-Name`?

to avoid warnings like:

```
[WARNING] *********************************************************************************************************************************************************************************************
[WARNING] * Required filename-based automodules detected: [prometheus-metrics-core-1.0.0.jar, prometheus-metrics-model-1.0.0.jar]. Please don't publish this project to a public artifact repository! *
[WARNING] *********************************************************************************************************************************************************************************************
```

solomax:
Hello @fstab,

As far as I understand this can be implemented 2 ways:

**First Option:** special entry in `MANIFEST` of every module
 
here is an idea:

```diff
diff --git a/pom.xml b/pom.xml
index 2845c00d..2f468daa 100644
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <automatic-module-name>--module-name-need-to-be-overriden--</automatic-module-name>
     </properties>
 
     <distributionManagement>
@@ -192,6 +193,11 @@
                 <artifactId>maven-bundle-plugin</artifactId>
                 <version>2.4.0</version>
                 <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Automatic-Module-Name>${automatic-module-name}</Automatic-Module-Name>
+                    </instructions>
+                </configuration>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
diff --git a/prometheus-metrics-core/pom.xml b/prometheus-metrics-core/pom.xml
index 25007d1e..a31705c2 100644
--- a/prometheus-metrics-core/pom.xml
+++ b/prometheus-metrics-core/pom.xml
@@ -32,6 +32,9 @@
         </developer>
     </developers>
 
+    <properties>
+        <automatic-module-name>io.prometheus.metrics.core</automatic-module-name>
+    </properties>
     <dependencies>
         <dependency>
             <groupId>io.prometheus</groupId>
```

as you can see:
1. special property is defined in parent pom
2. later on it is set the valid name for this particular module (done for one module only)

**Another Option:** `module-info.java` need to be added to every module

IMO first option is faster and lightweight, with second one you will create _real_ modules  :)

fstab:
[...]

PR #874, opened 2023-10-15
Fixes issue #871: Automatic-Module-Names are added

solomax:


solomax:
@fstab hope this one is better :)

fstab (changes requested):
It's looking good, thanks you very much. Just one tiny thing: Please move the `<properties>` to the same place in all `pom.xml` files, between `<description>` and `<licenses>`:

```xml
    <name>...</name>
    <description>
        ...
    </description>

    <properties>
        all properties here
    </properties>

    <licenses>
        ...
    </licenses>
```

Sometimes we need to make changes to multiple `pom.xml` files, and I learned that this is a lot easier if the structure of the `pom.xml` files is the same. So I'm trying to keep the order of tags consistent.

solomax:
> It's looking good, thanks you very much. Just one tiny thing: Please move the `<properties>` to the same place in all `pom.xml` files, between `<description>` and `<licenses>`:

done!

fstab:
Awesome, thank you!

Issue #877, opened 2023-10-16
Add Exemplars support for the OpenTelemetry Exporter


fstab:
Exemplars are not implemented yet in the OpenTelemetry exporter. We should add support for that.

https://github.com/prometheus/client_java/blob/586c2f9e4c62553eb157c82b292377521375f554/prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/otelmodel/PrometheusData.java#L47C48-L47C48

PR #883, opened 2023-10-29
#887 otlp exporter exemplars

0x006EA1E5:
@fstab 
This PR adds support for exemplars in the OTLP Exporter
Addressed https://github.com/prometheus/client_java/issues/877

The change seems fairly trivial, but I wonder if we need more test coverage?

### DIFF
--- a/prometheus-metrics-exporter-opentelemetry/pom.xml
+++ b/prometheus-metrics-exporter-opentelemetry/pom.xml
@@ -79,6 +79,30 @@
             <version>4.13.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wiremock</groupId>
+            <artifactId>wiremock</artifactId>
+            <version>3.2.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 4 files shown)


0x006EA1E5:
Is there a reason why we are using JUnit 4, not 5?

fstab (changes requested):
Thanks a lot, it's looking great.

One tiny thing: There's an empty file `trace.proto` which I think is not needed.

Regarding tests: I like your test, it's a good way to test this. I know that test coverage of the OTel exporter is currently not great. For the Prometheus formats the `ExpositionFormatsTest` is much more complete, but it was also a lot of work :).

> Is there a reason why we are using JUnit 4, not 5?

No, it's just that nobody updated it. If you like to create a PR: I'd appreciate an update to JUnit 5 :)

0x006EA1E5:
> One tiny thing: There's an empty file trace.proto which I think is not needed.

Removed, sorry about that

0x006EA1E5:
@fstab Are there any more actions you need me to do here? Is it suitable to me merged as is?

fstab (approved):
Looks great, thanks a lot!

Issue #887, opened 2023-11-01
Counts in ClassicHistogramBuckets cannot be negative throws when using SimpleclientCollector bridge


turchenkoalex:
If I use non-integer numbers in buckets for a classic histogram, an error is thrown when scraping metrics

Sample code to reproduce the error
```java
package io.prometheus.metrics.simpleclient.bridge;

import io.prometheus.client.CollectorRegistry;
import io.prometheus.client.Histogram;
import io.prometheus.metrics.model.registry.PrometheusRegistry;

public class Test {
    public static void main(String[] args) {
        CollectorRegistry origRegistry = new CollectorRegistry();
        PrometheusRegistry newRegistry = new PrometheusRegistry();
        SimpleclientCollector.builder()
                .collectorRegistry(origRegistry)
                .register(newRegistry);

        Histogram histogram = Histogram.build()
                .name("response_size_bytes")
                .help("response size in Bytes")
                .labelNames("status")
                .buckets(0.01, 0.1)
                .register(origRegistry);

        histogram.labels("200").observe(1);

        newRegistry.scrape(); // throws IllegalArgumentException: Counts in ClassicHistogramBuckets cannot be negative.
    }
}
```

PR #888, opened 2023-11-01
Fixes issue #887: Sort histogram bounds before validation

turchenkoalex:
Fixed issue #887 

Do sorting after the bounds array is filled. And added to the unit test the bucket value at which the error occurred.


fstab (approved):
Good catch, thanks a lot!

Issue #891, opened 2023-11-03
MultiCollector scrape(filter) is empty if prometheusNames() returns an empty list


fstab:
... todo

fstab:
Fixed with #893.

PR #893, opened 2023-11-04
Fix metric name filter in MultiCollector #891

fstab:
There was a bug that if a `MultiCollector` had empty `prometheusNames()` and `collect()` was called with a name filter, the collector was not scraped (wrong assumption that all names are excluded).

This PR fixes that bug.


fstab:
FYI @dhoard

PR #897, opened 2023-11-11
Unit class hardening

dhoard:
Added equals and hashcode to be able to compare a Unit.
Change code to catch an empty String on construction. 
Added JUnit test.


fstab (approved):
Thanks @dhoard

PR #906, opened 2023-12-11
avoid unnecessary ArrayList creations; use Collections.singletonList …

ispringer:
…rather than Arrays.asList where possible


fstab (commented):
Thanks a lot for the PR! To be honest, I find immutability in Java really hard. For example, take the following code:

```java
String[] excludedName = new String[1];

excludedName[0] = "some name";
ExporterFilterProperties filter1 = ExporterFilterProperties.builder().excludedNames(excludedName).build();

excludedName[0] = "some other name";
ExporterFilterProperties filter2 = ExporterFilterProperties.builder().excludedNames(excludedName).build();
```

To be fair, it's  a contrived example.

However, if we don't copy the array then `filter1` will be altered after it was created, so it won't be immutable.

I don't see a good way to avoid this other than creating copies of the arrays.

PR #911, opened 2024-01-03
Add Backwards-Compatibility

thinker0:
- metrics name

There is a problem where an exception occurs if '_total' is in the Legacy Gauge name.

@fstab


fstab (changes requested):
Thanks a lot, and sorry for the delay over Christmas.

This issue does not only affect the `_total` suffix, but also other suffixes like `_info` or `_created`. And it does not only affect Gauge, but all metrics.

So, what do you think of adding
```java
.name(sanitizeMetricName(samples.name))
```
with `import static io.prometheus.metrics.model.snapshots.PrometheusNaming.sanitizeMetricName;` for all metric types?

thinker0:
https://github.com/prometheus/client_java/blob/main/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java#L99-L108

In my case, an exception occurred here. MetricMetadata creation could all be a problem?

fstab:
I was thinking: It makes sense to add the `sanitizeMetricName()` to all types, like `convertCounter()`, `convertHistogram()`, `convertSummary()`, `convertStateSet()`, `convertUnknown()`.

I think the issue is the same for all metric types, not just for Gauge.

fstab (approved):
Awesome, thank you!

PR #914, opened 2024-01-19
Remove duplicate Maven dependencies

ParanoidUser:
Ahoy, @fstab / @dhoard / @tomwilkie! Found some duplicate Maven dependency declarations. Should be no problem to remove them.


fstab (approved):
Thanks a lot!

PR #918, opened 2024-01-29
Add a javax.servlet-api exporter 

kingster:
This PR adds module `prometheus-metrics-exporter-servlet-javax` which exposes a servlet compatible with the javax-servlet.api


fstab (approved):
Awesome, thanks a lot!

kingster:
@fstab When is the next planned release? If it is quite far away can we publish a snapshot version in the mean time?

PR #919, opened 2024-01-31
Add jvm native memory metrics

meier-christoph:
This pr adds 2 new metrics reporting on JVM native memory usage.

- `jvm_native_memory_committed_bytes` : Committed bytes of a given JVM. Committed memory represents the amount of memory the JVM is using right now.
- `jvm_native_memory_reserved_bytes` : Reserved bytes of a given JVM. Reserved memory represents the total amount of memory the JVM can potentially use.

A user needs to start the jvm with the `-XX:NativeMemoryTracking=summary` flag or else this will do nothing (but it will not fail either i.e. it is safe to use on any jvm by default).

The values reported are identical as those reported by `jcmd {pid} VM.native_memory`.

@fstab @dhoard @tomwilkie


fstab (approved):

fstab:
Hi Christoph, sorry for taking so long before reviewing this. The code looks great, thanks a lot for the contribution! I merged it. It will be shipped with the next release.

PR #930, opened 2024-03-20
Allow Exemplars on histogram count and sum

fstab:
Add Exemplars to histogram `_count`.

This will enable Examplars for

* Request rates, like `rate(request_duration_seconds_count[$__rate_interval])`
* Average response times, like `rate(request_duration_seconds_sum[$__rate_interval]) / rate(request_duration_seconds_count[$__rate_interval])`.

Note that Exemplars on all time series are not enabled by default. If you want to use this feature, you explicitly need to set `io.prometheus.exporter.exemplarsOnAllMetricTypes=true` in the configuration file (or via System property), or you need to enable it via the `OpenMetricsTextFormatWriter` constructor parameter if you initialize the `OpenMetricsTextFormatWriter` programmatically.

---

We don't add Exemplars to histogram `_sum`, and we also removed them for summary `_sum`. The reason is that there doesn't seem to be a good use case for exemplars on `_sum`.

### DIFF
--- a/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetricsTextFormatWriter.java
+++ b/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetricsTextFormatWriter.java
@@ -138,11 +138,9 @@ public class OpenMetricsTextFormatWriter implements ExpositionFormatWriter {
                 }
                 writeScrapeTimestampAndExemplar(writer, data, exemplar);
             }
+            // In OpenMetrics format, histogram _count and _sum are either both present or both absent.
             if (data.hasCount() && data.hasSum()) {
-                // In OpenMetrics format, histogram _count and _sum are either both present or both absent.
-                // While Prometheus allows Exemplars for histogram's _count and _sum now, we don't
-                // use Exemplars here to be backwards compatible with previous behavior.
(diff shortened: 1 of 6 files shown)


prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetricsTextFormatWriter.java
```diff
@@ -241,33 +243,23 @@ private void writeUnknown(OutputStreamWriter writer, UnknownSnapshot snapshot) t
...
         }
         if (data.hasSum()) {
             writeNameAndLabels(writer, metadata.getPrometheusName(), sumSuffix, data.getLabels());
             writeDouble(writer, data.getSum());
-            if (exemplars.size() > 0) {
-                writeScrapeTimestampAndExemplar(writer, data, exemplars.get(exemplarIndex));
-            } else {
-                writeScrapeTimestampAndExemplar(writer, data, null);
-            }
+            writeScrapeTimestampAndExemplar(writer, data, exemplars.getLatest());
```
jonatan-ivanov, line 262: What is the use case of having an exemplar on `_sum`? I'm thinking it might be unnecessary if there is no use-case for it.

fstab: You're right, I removed Exemplars on `_sum`. Both for histograms and summaries.

PR #939, opened 2024-04-01
Fix link to JavaDocs in Config documentation

shakuzen:
The links in the table were swapped. This swaps them so the right method is linked.


fstab (approved):
Thanks a lot!

PR #955, opened 2024-05-07
Start Prometheus HTTP server on the provided executor

mberndt123:
The metrics HTTP server should run on a daemon JVM
thread so as to not keep the JVM running even
though all relevant threads have already exited.
Unfortunately, the OpenJDK builtin HTTP server
likes to spawn its own HTTP Dispatcher thread on
startup.
https://github.com/openjdk/jdk/blob/02c95a6d7eb77ed17ae64d0f585197e87a67cc4a/src/jdk.httpserver/share/classes/sun/net/httpserver/ServerImpl.java#L190
This new thread inherits the daemon flag from the
thread that started it, therefore we need to call
the start method from the executor service,
because those threads are daemon threads.

### DIFF
--- a/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
+++ b/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
@@ -6,7 +6,6 @@ import com.sun.net.httpserver.HttpHandler;
 import com.sun.net.httpserver.HttpServer;
 import com.sun.net.httpserver.HttpsConfigurator;
 import com.sun.net.httpserver.HttpsServer;
-import io.prometheus.metrics.config.ExporterHttpServerProperties;
 import io.prometheus.metrics.config.PrometheusProperties;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
 
@@ -14,6 +13,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.SynchronousQueue;
@@ -55,7 +55,13 @@ public class HTTPServer implements Closeable {
         registerHandler("/", defaultHandler == null ? new DefaultHandler() : defaultHandler, authenticator);
         registerHandler("/metrics", new MetricsHandler(config, registry), authenticator);
         registerHandler("/-/healthy", new HealthyHandler(), authenticator);
-        this.server.start();
+        try {
+            executorService.submit(() -> this.server.start()).get();
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e.getCause());
+        }
     }
 
     private void registerHandler(String path, HttpHandler handler, Authenticator authenticator) {
--- a/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
+++ b/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
@@ -57,6 +57,7 @@ public class HTTPServer implements Closeable {
         registerHandler("/-/healthy", new HealthyHandler(), authenticator);
         try {
             executorService.submit(() -> this.server.start()).get();
+            // calling .get() on the Future here to avoid silently discarding errors
         } catch (InterruptedException e) {
             throw new RuntimeException(e);
         } catch (ExecutionException e) {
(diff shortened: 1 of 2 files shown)


hgranillo (approved):

mberndt123:
OK, here's the thing. This change is working exactly as intended and prevents Prometheus from spawning non-Daemon threads. In the tests however, this Prometheus thread is the _only_ non-Daemon thread other than the main thread, and the main thread exits right after starting the Prometheus HTTP server, leading to the termination of the program before the tests can complete.

There are probably programs out there that run everything in daemon threads, the only thing keeping their JVM alive being the Prometheus HTTP server, and so this change would break their program.
I would argue that this is a bug in their program, and that it's more important to prevent an application from entering “Zombie mode” where the only thing that's still running is the metrics HTTP server than it is to support such users. But that's a call that the library maintainers need to make.

dhoard:
> There are probably programs out there that run everything in daemon threads, the only thing keeping their JVM alive being the Prometheus HTTP server, and so this change would break their program. I would argue that this is a bug in their program, and that it's more important to prevent an application from entering “Zombie mode” where the only thing that's left running is the metrics HTTP server, than it is to support such users. But that's a call that the library maintainers need to make.

I feel this is a good change - the application should be responsible for preventing the application from exiting and shouldn't rely on `HTTPServer` threads.

If this PR causes test issues/failures in `client_java` due to the behavior change, then the `client_java` tests should be fixed as part of the PR.

---

We could add a `join` method to `HTTPServer` for convenience...

```
public void join() {
    try {
        Thread.currentThread().join();
        // DO NOTHING
    }
}
```

mberndt123:
Thanks you for your encouraging comment @dhoard. 

Regarding the `join`method you've proposed, I'm not quite sure I understand. When a thread waits for itself to terminate, that's just a deadlock, right? The tests already work thanks to the `sleep` statement that I added – 10 seconds should easily be enough for the test to complete, and it doesn't slow down the tests because afaics the test doesn't wait for the test server to terminate.

dhoard:
> Regarding the `join`method you've proposed, I'm not quite sure I understand. When a thread waits for itself to terminate, that's just a deadlock, right?

The thought was if someone wanted the `HTTPServer` to block the application from exiting, we could provide an  `HTTPServer` `join` method for convenience.

```
class SomeApplication {
    
        // code that creates and runs daemon threads
        
        // create the HTTPServer
        HTTPServer httpServer = ...
        
        // cause the main thread to sleep until interrupted,
        // blocking the application from exiting
        httpServer.join();
    }
}
```

mberndt123:
Oh, I see what you mean now. I think such functionality is beyond the scope of the library and it doesn't really save much code anyway.

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
```diff
@@ -55,7 +55,13 @@ private HTTPServer(PrometheusProperties config, ExecutorService executorService,
         registerHandler("/", defaultHandler == null ? new DefaultHandler() : defaultHandler, authenticator);
         registerHandler("/metrics", new MetricsHandler(config, registry), authenticator);
         registerHandler("/-/healthy", new HealthyHandler(), authenticator);
-        this.server.start();
+        try {
+            executorService.submit(() -> this.server.start()).get();
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e.getCause());
```
dhoard, line 63: When wrapping exceptions, pass the exception being wrapped.

```
throw new RuntimeException(e);
```

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
```diff
@@ -55,7 +55,13 @@ private HTTPServer(PrometheusProperties config, ExecutorService executorService,
         registerHandler("/", defaultHandler == null ? new DefaultHandler() : defaultHandler, authenticator);
         registerHandler("/metrics", new MetricsHandler(config, registry), authenticator);
         registerHandler("/-/healthy", new HealthyHandler(), authenticator);
-        this.server.start();
+        try {
+            executorService.submit(() -> this.server.start()).get();
```
dhoard, line 59: Remove the `get` call if the `Future` is not being used.

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
mberndt123: OK

dhoard (approved):

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
mberndt123: Hey @dhoard,
I'd like to push back on this. Starting the HTTP server might fail and throw an exception, e. g. when the port is already in use. Unless I call `.get()` on this `Future`, this error would be silently discarded, which is something that should be avoided.

dhoard: @mberndt123 good catch/makes sense. Can you add a comment to the code? ... anyone looking at the code wouldn't get this nuance - I didn't.

mberndt123: OK

dhoard (approved):

dhoard:
@mberndt123 Thanks for the PR!

PR #958, opened 2024-05-11
updated java version to 17

opensourcevk:


fstab:
Hi @opensourcevk, there are a lot of people out there using `client_java` in Java 8 applications. For example, the [jmx_exporter](https://github.com/prometheus/jmx_exporter) uses `client_java` and is supposed to be compatible with Java 8. We would like to remain compatible with Java 8 for the foreseeable future.

That being said, just because `client_java` is compatible with Java 8 doesn't mean you can't use it with newer Java versions. Nothing prevents you from using the `client_java` library with Java 22 and beyond.

PR #960, opened 2024-05-15
Prevent illegal metric names

fstab:
If a Prometheus scrape finds a metric that has `UNIT` metadata but a name that does not end with the unit, the entire scrape fails.

https://github.com/prometheus/prometheus/blob/2e30f1231b6686e69afe1abb199c57dfe98fba07/model/textparse/openmetricsparse.go#L321-L323

With this PR we prevent these illegal metric names by throwing an `IllegalArgumentException`. The goal is to make it impossible to create illegal metric names with `prometheus-metrics-model`.

### DIFF
--- a/prometheus-metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/ExpositionFormatsTest.java
+++ b/prometheus-metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/ExpositionFormatsTest.java
@@ -1,26 +1,11 @@
 package io.prometheus.metrics.expositionformats;
 
+import io.prometheus.metrics.model.snapshots.*;
 import io.prometheus.metrics.shaded.com_google_protobuf_3_25_3.TextFormat;
 import io.prometheus.metrics.expositionformats.generated.com_google_protobuf_3_25_3.Metrics;
-import io.prometheus.metrics.model.snapshots.CounterSnapshot;
 import io.prometheus.metrics.model.snapshots.CounterSnapshot.CounterDataPointSnapshot;
-import io.prometheus.metrics.model.snapshots.Exemplar;
-import io.prometheus.metrics.model.snapshots.Exemplars;
-import io.prometheus.metrics.model.snapshots.ClassicHistogramBuckets;
-import io.prometheus.metrics.model.snapshots.GaugeSnapshot;
 import io.prometheus.metrics.model.snapshots.GaugeSnapshot.GaugeDataPointSnapshot;
-import io.prometheus.metrics.model.snapshots.HistogramSnapshot;
-import io.prometheus.metrics.model.snapshots.InfoSnapshot;
-import io.prometheus.metrics.model.snapshots.Labels;
-import io.prometheus.metrics.model.snapshots.MetricSnapshot;
-import io.prometheus.metrics.model.snapshots.MetricSnapshots;
-import io.prometheus.metrics.model.snapshots.NativeHistogramBuckets;
-import io.prometheus.metrics.model.snapshots.Quantiles;
-import io.prometheus.metrics.model.snapshots.StateSetSnapshot;
-import io.prometheus.metrics.model.snapshots.SummarySnapshot;
 import io.prometheus.metrics.model.snapshots.SummarySnapshot.SummaryDataPointSnapshot;
-import io.prometheus.metrics.model.snapshots.Unit;
-import io.prometheus.metrics.model.snapshots.UnknownSnapshot;
 import io.prometheus.metrics.model.snapshots.UnknownSnapshot.UnknownDataPointSnapshot;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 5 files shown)


prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java
```diff
@@ -105,5 +105,11 @@ private void validate() {
             throw new IllegalArgumentException("'" + name + "': Illegal metric name. " + error
                     + " Call " + PrometheusNaming.class.getSimpleName() + ".sanitizeMetricName(name) to avoid this error.");
         }
+        if (hasUnit()) {
+            if (!name.endsWith("_" + unit) && !name.endsWith("." + unit)) {
+                throw new IllegalArgumentException("'" + name + "': Illegal metric name. The name must end with _" + unit + "."
```
jonatan-ivanov, line 110: What do you think about adding some extra information in the error message?
```suggestion
                throw new IllegalArgumentException("'" + name + "': Illegal metric name. If the unit is non-null, the name must end with it (_" + unit + ")."
```

prometheus-metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/ExpositionFormatsTest.java
```diff
@@ -1825,24 +1810,24 @@ public void testUnknownMinimal() throws IOException {
     @Test
     public void testUnknownWithDots() throws IOException {
         String openMetrics = "" +
-                "# TYPE some_unknown_metric unknown\n" +
-                "# UNIT some_unknown_metric bytes\n" +
-                "# HELP some_unknown_metric help message\n" +
-                "some_unknown_metric{test_env=\"7\"} 0.7\n" +
+                "# TYPE some_unknown_metric_bytes unknown\n" +
+                "# UNIT some_unknown_metric_bytes bytes\n" +
```
jonatan-ivanov, line 1814: With Micrometer we ran into the same: we had verifications in our tests for the invalid output so I added an integration test with a Prometheus server: https://github.com/micrometer-metrics/micrometer/pull/5039

What do you think about doing something similar in the Prometheus Client?
The test starts an http server (could be the one in the Prometheus Client), starts a Prometheus sever, Prometheus scrapes the http server, the test verifies if the scraped metrics are available in Prometheus or not.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java
jonatan-ivanov: Also, there is a twist: this is only a problem for the OpenMetrics format, for the Prometheus Text format it is completely ok doing this (since there is no `# UNIT` anyways). I think though for the sake of consistency this should be validated in both cases to avoid client-side logic that produces different names if the format is different.

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/MetricMetadataTest.java
```diff
@@ -25,9 +25,9 @@ public void testIllegalName() {
 
     @Test
     public void testSanitizationIllegalCharacters() {
-        MetricMetadata metadata = new MetricMetadata(sanitizeMetricName("my_namespace/http.server.duration"), "help string", Unit.SECONDS);
-        Assert.assertEquals("my_namespace_http.server.duration", metadata.getName());
-        Assert.assertEquals("my_namespace_http_server_duration", metadata.getPrometheusName());
+        MetricMetadata metadata = new MetricMetadata(sanitizeMetricName("my_namespace/http.server.duration", Unit.SECONDS), "help string", Unit.SECONDS);
```
jonatan-ivanov, line 28: Would it worth adding a test that verifies if the exception is thrown in case of the invalid scenario?

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java
```diff
@@ -121,6 +121,20 @@ public static String sanitizeMetricName(String metricName) {
         return sanitizedName;
     }
 
+    /**
+     * Like {@link #sanitizeMetricName(String)}, but also makes sure that the unit is appended
+     * as a suffix if the unit is not {@code null}.
+     */
+    public static String sanitizeMetricName(String metricName, Unit unit) {
```
jonatan-ivanov, line 128: Would it worth adding a test to this?

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java
fstab: Thanks, I adapted the Exception message.

I agree, while it's theoretically possible to expose the metric name without the unit suffix in text format this is not something we should support.

prometheus-metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/ExpositionFormatsTest.java
fstab: YES, I actually had the same idea and started doing this for our latest module: `prometheus-metrics-exporter-pushgateway`.

https://github.com/prometheus/client_java/blob/main/integration-tests/it-pushgateway/src/test/java/io/prometheus/metrics/it/pushgateway/PushGatewayIT.java

I would like to implement similar tests for other modules over time.

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/MetricMetadataTest.java
fstab: Added a test.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java
fstab: Added a test.

fstab:
Thanks for the review. Looks like I merged this prematurely, but I just pushed a new commit addressing the review comments.

PR #963, opened 2024-05-23
Simple MetricsSnapshot performance change

dhoard:
Simple MetricsSnapshot performance change to prevent calling `prometheusName(name)` in a loop.


fstab (approved):
Thanks a lot, looks good to me!

PR #968, opened 2024-06-18
feat: add bearerToken builder to pushgateway exporter

FUSAKLA:
Add possibility to set bearer token when pushingmetrics to pushgateway

The PGW itself does not support the bearer token auth but it is needed if behind some auth proxy for example


FUSAKLA:
@fstab Maybe would you PTAL? :pray: 

I initially added also integration tests, but PGW itself does not support the bearer token auth yet so I dropped them since it was testing nothing.

FUSAKLA:
Or maybe @dhoard ?

dhoard (approved):

FUSAKLA:
Thanks! :tada:

PR #969, opened 2024-06-24
Add abstract build method to MetricSnapshot.Builder

mimaison:
Having an abstract `build()` method allows writing generic code for the different builders.

For context, I'm currently updating a library from `simpleclient` to the new 1.X version. The library converts metrics from Apache Kafka where we can have multiple metric objects with the same metric name but different labels. 

With `simpleclient`, we could also have multiple `MetricFamilySamples` with the same name, but that's not the case anymore with `MetricSnapshots` where each `MetricSnapshot` must have a unique name.

So when converting metrics from Kafka, we need to keep un-built `MetricSnapshot.Builder` and add datapoints when we find metrics with the same names. In this example being able to store various `MetricSnapshot.Builder` objects in a collection is useful and allows building them all easily.

### DIFF
--- a/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/CounterSnapshot.java
+++ b/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/CounterSnapshot.java
@@ -139,6 +139,7 @@ public class CounterSnapshot extends MetricSnapshot {
             return this;
         }
 
+        @Override
         public CounterSnapshot build() {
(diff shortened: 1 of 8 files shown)


mimaison:
@fstab @dhoard @tomwilkie Can you take a look? Thanks

mimaison:
Ping @fstab @dhoard @tomwilkie

dhoard (approved):

dhoard:
@mimaison Thanks for the PR!

PR #972, opened 2024-07-19
Nullify noLabels in StatefulMetric during clear() to prevent no-label inc to stop working

### LINKED ISSUE #971: StatefulMetric clear() invocation causes noLabels increase to stop reporting on collect()
It looks like the current implementation of the `io.prometheus.metrics.core.metrics.StatefulMetric#clear` method (see #935) is clearing the data, but not nullifying the `io.prometheus.metrics.core.metrics.StatefulMetric#noLabels` field, allowing that to be returned by the `io.prometheus.metrics.core.metrics.StatefulMetric#getNoLabels` method, but preventing it from being correctly collected.

This is easily reproducible in the `io.prometheus.metrics.core.metrics.StatefulMetricTest#testClearNoLabels` test by adding an `inc()` call and verification after the first clear:

```
...
        counter.clear();
        // No labels is always present, but as no value has been observed after clear() the value should be 0.0
        Assert.assertEquals(1, counter.collect().getDataPoints().size());
        Assert.assertEquals(0.0, counter.collect().getDataPoints().get(0).getValue(), 0.0);

        // Now we're testing label-less inc() still works correctly
        counter.inc();
        Assert.assertEquals(1, counter.collect().getDataPoints().size());
        Assert.assertEquals(1.0, counter.collect().getDataPoints().get(0).getValue(), 0.0);
```

If my understanding of the issue is correct, the `io.prometheus.metrics.core.metrics.StatefulMetric#clear` should also set `noLabels = null`.

iaco86:
Fixes #971

- nullified volatile `noLabels` variable on `clear()`
- enhanced `testClearNoLabels`

@fstab @dhoard @tomwilkie

### DIFF
--- a/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/StatefulMetric.java
+++ b/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/StatefulMetric.java
@@ -114,6 +114,7 @@ abstract class StatefulMetric<D extends DataPoint, T extends D> extends MetricWi
      */
     public void clear() {
         data.clear();
+        noLabels = null;
     }
 
     protected abstract T newDataPoint();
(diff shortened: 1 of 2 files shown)


dhoard (approved):

dhoard:
@iaco86 Thanks for the PR!

PR #979, opened 2024-08-09
add support for guava instrumentation with 1.X.X

pheyken:
Hi all, this is my attempt on migrating one of the simpleclient collectors over to the new Prometheus 1.X Client. This is supposed to behave exactly the same as simpleclient_guava, thus keeping the exact same metrics and the same naming. Let me know if you see any issue with this and I'm happy to adopt the Collector


pheyken:
Hey @fstab & @dhoard, if you have a few minutes to spare, a review would be highly appreciated

dhoard:
@pheyken merged. Thanks for the PR!

PR #983, opened 2024-08-30
add support for caffeine instrumentation with 1.X.X

pheyken:
Hi @dhoard & @fstab, thanks so much for merging #979. This is basically the same change, just for caffeine caches this time


dhoard:
@pheyken merged. Thanks for the PR!

PR #985, opened 2024-09-13
additional improve MetricSnapshots.Builder performance

burov4j:
The function `io.prometheus.metrics.model.registry.PrometheusRegistry.scrape` has O(n^2) complexity: higher size of the list `multiCollectors` generates higher amount of `snapshots` in the `io.prometheus.metrics.model.snapshots.MetricSnapshots.Builder` class on each loop iteration.

Because of this we have x100 times slower metrics scrape in our project on production.

This PR gave us good performance improvement: https://github.com/prometheus/client_java/pull/963
But it's not enough, because the complexity is still O(n^2). The current PR resolves the performance issue.


dhoard (approved):

dhoard:
@burov4j thanks for the PR!

PR #988, opened 2024-09-27
move to github actions / require google java format for changed files

zeitlinger:


zeitlinger:
@fstab what is the min java version?

PR #991, opened 2024-09-30
fix buffer sync logic using modern concurrency primitives

zeitlinger:
Alternative suggestion for https://github.com/prometheus/client_java/pull/970

### DIFF
--- a/benchmarks/src/main/java/io/prometheus/metrics/benchmarks/HistogramBenchmark.java
+++ b/benchmarks/src/main/java/io/prometheus/metrics/benchmarks/HistogramBenchmark.java
@@ -21,11 +21,11 @@ import org.openjdk.jmh.annotations.Threads;
  *
  * <pre>
  * Benchmark                                     Mode  Cnt      Score     Error  Units
- * HistogramBenchmark.openTelemetryClassic      thrpt   25    258.660 ±   6.736  ops/s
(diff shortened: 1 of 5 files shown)


benchmarks/src/main/java/io/prometheus/metrics/benchmarks/HistogramBenchmark.java
```diff
@@ -21,11 +21,11 @@
  *
  * <pre>
  * Benchmark                                     Mode  Cnt      Score     Error  Units
- * HistogramBenchmark.openTelemetryClassic      thrpt   25    258.660 ±   6.736  ops/s
- * HistogramBenchmark.openTelemetryExponential  thrpt   25    210.963 ±  11.288  ops/s
- * HistogramBenchmark.prometheusClassic         thrpt   25   1528.871 ±  43.598  ops/s
- * HistogramBenchmark.prometheusNative          thrpt   25   1282.643 ± 110.210  ops/s
- * HistogramBenchmark.simpleclient              thrpt   25   3376.016 ± 173.545  ops/s
+ * HistogramBenchmark.openTelemetryClassic      thrpt   25    333.576 ±   17.158  ops/s
```
zeitlinger, line 24: @fstab some are better some are worse (same computer) - is this also your experience?

PR #992, opened 2024-09-30
Various cleanups in prometheus-metrics-model

mimaison:
- Removed unused imports
- Added a few missing tests
- Fixed indentation issues
- Addressed a few Java style issues (interface types, enhanced for loops, adjusted visibility, etc)


prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/HistogramSnapshotTest.java
```diff
@@ -54,20 +54,21 @@ public void testGoodCaseComplete() {
                                 .createdTimestampMillis(createdTimestamp)
                                 .scrapeTimestampMillis(scrapeTimestamp)
                                 .build())
-                .dataPoint(HistogramDataPointSnapshot.builder()
-                        .count(3) // TODO how is that not a compile error? This is a protected method!
```
mimaison, line 58: Removed the TODO, we're in the same package so we can access a protected method

mimaison:
Thanks! Similar cleanups can be done in the other modules, so I'll follow up with a few more PRs if that's ok.

zeitlinger:
> Thanks! Similar cleanups can be done in the other modules, so I'll follow up with a few more PRs if that's ok.

yes - that's very welcome!

PR #1012, opened 2024-10-01
Various cleanups in instrumentation-jvm

mimaison:
The first commit contains cleanups and refactorings. The second commit was generated using `mvn spotless:apply` so spotless does not fail the build. I can separate them in 2 PRs if you'd prefer that, the spotless changes are pretty large.

PR #1014, opened 2024-10-01
Various cleanups in instrumentation-jvm

mimaison:
Same as #1012, let's see if the build runs here


mimaison:
Here the build fails because of spotless issues in te prometheus-metrics-model module which this PR does not touch: https://github.com/mimaison/client_java/actions/runs/11127458358/job/30919802107

zeitlinger:
> Here the build fails because of spotless issues in the prometheus-metrics-model module which this PR does not touch: https://github.com/mimaison/client_java/actions/runs/11127458358/job/30919802107

problem seems to be that the action runs in your repo - where `origin/main` is different - that's why the build check here is not working

zeitlinger:
> > I wonder if we should apply spotless across the whole project at once, and then restart working on code changes.
> 
> wondering that too - makes it easier to review changes

done - https://github.com/prometheus/client_java/pull/1015

mimaison:
Awesome, let me rebase!

PR #1062, opened 2024-10-02
don't use extra otel artifact

zeitlinger:


zeitlinger:
@fstab can you take a look?

PR #1063, opened 2024-10-02
include shaded protobuf directly

zeitlinger:


zeitlinger:
@fstab can you take a look?

PR #1085, opened 2024-10-02
Various cleanups in prometheus-metrics-core

mimaison:
Also added unit tests for CounterWithCallback, GaugeWithCallback, Summary and SummaryWithCallback.

### DIFF
--- a/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java
+++ b/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java
@@ -481,8 +481,8 @@ public class Histogram extends StatefulMetric<DistributionDataPoint, Histogram.D
       nativeZeroCount.reset();
       count.reset();
       sum.reset();
-      for (int i = 0; i < classicBuckets.length; i++) {
-        classicBuckets[i].reset();
+      for (LongAdder classicBucket : classicBuckets) {
+        classicBucket.reset();
       }
       nativeZeroThreshold = nativeMinZeroThreshold;
       nativeSchema = Histogram.this.nativeInitialSchema;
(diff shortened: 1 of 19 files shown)


prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java
```diff
@@ -481,8 +481,8 @@ private boolean maybeReset() {
       nativeZeroCount.reset();
       count.reset();
       sum.reset();
-      for (int i = 0; i < classicBuckets.length; i++) {
-        classicBuckets[i].reset();
+      for (LongAdder classicBucket : classicBuckets) {
```
zeitlinger, line 484: the performance is not the same - but I guess it's ok here

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/CounterWithCallbackTest.java
```diff
@@ -1,6 +1,42 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.Test;
+
 public class CounterWithCallbackTest {
 
-  // TODO :). Anyway, callbacks are implicitly covered by the JVM metrics tests as well.
+  @Test
+  public void testCounter() {
```
zeitlinger, line 14: wdyt of using assertj for new tests?

prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java
mimaison: My understanding is that on a modern JVM the performance difference should be negligible.

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/CounterWithCallbackTest.java
mimaison: As I'm not a maintainer on this project, I wasn't sure about adding new dependencies. I see you're doing the switch to assertj in https://github.com/prometheus/client_java/pull/1092

Issue #1088, opened 2024-10-03
[Feature] add ability to execute handlers wrapped a Subject doAs from a custom authenticator


gtully:
When the mbean server is locked down with an authorisation filter, it needs an authenticated subject to verify the identity of the caller. A custom authenticator can provide the subject in an HttpExchange named attribute. But we then need to invoke our handlers using that subject for a doAs call.
A generic filter is called to early and system filters are not visible. Using a handler wrapper that can respect an attribute name configuration option and make the necessary doAs call can work.

Then adding a customAuthenticator to the jmx_exporter can work nicely.

PR #1089, opened 2024-10-03
http server, add subject.doAs handler wrapper for exchange attribute …

gtully:
…configured via authenticatedSubjectAttributeName #1088


prometheus-metrics-exporter-httpserver/pom.xml
```diff
@@ -42,6 +42,12 @@
             <artifactId>prometheus-metrics-exporter-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
zeitlinger, line 45: junit5 is now included in all projects

prometheus-metrics-exporter-httpserver/src/test/java/io/prometheus/metrics/exporter/httpserver/HTTPServerTest.java
```diff
@@ -0,0 +1,79 @@
+package io.prometheus.metrics.exporter.httpserver;
+
+import static org.junit.Assert.assertTrue;
```
zeitlinger, line 3: please use assertj

Issue #984, opened 2024-09-11
opentelemetry-exporter does not consider resource attributes set via env var


clawoflight:
OpenTelemetry [allows](https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_resource_attributes) setting resource attributes via the environment variable `OTEL_RESOURCE_ATTRIBUTES`.

It would be great if the opentelemetry-exporter would take these into account and automatically augment the metrics with the corresponding attributes. As a workaround, I need to do that in the builder.

PR #1147, opened 2024-10-14
use otel autoconfigure 

zeitlinger:
Fixes https://github.com/prometheus/client_java/issues/984


zeitlinger:
@fstab here's the diff of resource attributes of agent and prom client - now that the client can use resource detectors (with this pr)

I think all we really want to copy is the service instance ID - wdyt?

(instance will be added to client)

image

PR #1150, opened 2024-10-15
fix deadlock

zeitlinger:
actual deadlock https://gist.github.com/zeitlinger/2c547a181d2f5f0b67c0a53b385354eb


zeitlinger:
close in favor of https://github.com/prometheus/client_java/pull/1152

Issue #1174, opened 2024-10-24
[prometheus-metrics-instrumentation-caffeine] Cannot register multiple Caffeine cache


kaustubhDsarathi:
Context:
I'm using prometheus-metrics-instrumentation-caffeine to instrument my caffeine cache.
```java
public class UserCache {

    public static final String KEY_REMOVED_FROM_CACHE = "key={} removed from cache";
    @Getter
    private LoadingCache<String, Optional<UserDetailDto>> cache1;
    private LoadingCache<String, Optional<UserDetailDto>> cache2;
    @Getter
    private LoadingCache<String, Optional<UserDetailDto>> cache3;
    private final UserRepository userRepository;
    private final CacheConfig cacheConfig;
    private final PrometheusMeterRegistry prometheusMeterRegistry;

    @PostConstruct
    void init() {
        if(cacheConfig.isEnabled()) {
            ExecutorService threadPool = Executors.newFixedThreadPool(4);
            cache1 = Caffeine.newBuilder()
                    .maximumSize(100_000) // 100_000 test-user data will be cached.
                    .removalListener(TestCache::onRemoval)
                    .refreshAfterWrite(2, TimeUnit.HOURS)
                    .executor(threadPool)
                    .recordStats()
                   .build(userRepository::loadCache1);

            cache2 = Caffeine.newBuilder()
                    .maximumSize(5_000) // 5000 test-user data will be cached.
                    .removalListener(TestCache::onRemoval)
                    .refreshAfterWrite(5, TimeUnit.HOURS) // refresh cache after 5 hours write
                    .executor(threadPool)
                    .build(userRepository::loadCache2);


            cache3 = Caffeine.newBuilder()
                    .maximumSize(100_000) // 100_000 test-user data will be cached.
                    .removalListener(TestCache::onRemoval)
                    .refreshAfterWrite(1, TimeUnit.HOURS)
                    .executor(threadPool)
                    .build(userRepository::loadCache3);

            CacheMetricsCollector collector = new CacheMetricsCollector();
            collector.addCache("cache1", cache1);
            collector.addCache("cache2", cache2);
            collector.addCache("cache3", cache3);

            prometheusMeterRegistry.getPrometheusRegistry().register(collector);
        }

    private static void onRemoval(@Nullable Object key, @Nullable Object value, RemovalCause cause) {
        log.atDebug().log(KEY_REMOVED_FROM_CACHE, key);
    }
}
```
[...]

PR #1203, opened 2024-11-13
fix register part of documentation for caffeine & guava instrumentati…

pheyken:
…on + fix typo in method name

Hey @zeitlinger,

this is fixing the documentation error as pointed out by @kaustubhDsarathi in #1174

Also, included is a fix for a typo which was introduced by mistake in #1138, hope its alright to include this breaking change


zeitlinger:
> Also, included is a fix for a typo which was introduced by mistake in #1138, hope its alright to include this breaking change

yes, just saw that I changed that by mistake in the last release

PR #1206, opened 2024-11-15
[prometheus-metrics-instrumentation-caffeine] implement getPrometheus…

pheyken:
…Names

Hey all, implementing this should have following benefits:
- registering the CacheCollector twice will fail immediately and not when collecting metrics
- avoid collecting metrics from the CacheCollector which would be discarded when a filter is passed to the PrometheusRegistry when scraping


zeitlinger:
Thanks!

PR #1211, opened 2024-11-22
[prometheus-metrics-instrumentation-guava] implement getPrometheusNames

pheyken:
Hey @zeitlinger, thanks so much for taking care of #1206. This is the same change for the guava instrumentation.


zeitlinger:
Thanks for the  contribution!

PR #1217, opened 2024-11-29
Fix next SNAPSHOT version

fstab:


zeitlinger:
10.0.0-SNAPSHOT was intentional - so that it never has to change locally.
Is it causing any issue?

dhoard:
I like using `<PREVIOUS_VERSION>-post` for the development/unreleased main branch code.

- it identifies the latest release as the code base
- doesn't infer what the next version will be

If the latest release is `1.3.4`, the next release could be...

- `1.3.5` (bug fixes)
- `1.4.0` (new features)
- `2.0.0` (major change)

PR #1223, opened 2024-12-06
Remove the otel-agent-resources Maven module

fstab:
The `ResourceAttributesFromOtelAgent` class lived in its own Maven module named `prometheus-metrics-exporter-opentelemetry-otel-agent-resources`. This is unnecessary. As this is only used in `prometheus-metrics-exporter-opentelemetry`, we can merge the two modules and only keep one.


zeitlinger:
I think the problem is that this line is shaded - and thus the wrong class loaded:

https://github.com/prometheus/client_java/blame/e95a2ee993e9c8bd23da4e86a998773f603c647b/prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/ResourceAttributesFromOtelAgent.java#L48

I forgot to add that as a comment.

PR #1241, opened 2024-12-22
Use buffered writer

michaldo:
Without buffered reader, every single output character through given stack goes to "StreamEncoder public void write(int c)" and single element char array is created. It makes pressure to GC and CPU

In my case 800K char[1] arrays are created and garbage collected

BufferedReader reduce the pressure

```
write:130, StreamEncoder (sun.nio.cs)
write:201, OutputStreamWriter (java.io)
writeMetadata:338, PrometheusTextFormatWriter (io.prometheus.metrics.expositionformats) 
writeGauge:129, PrometheusTextFormatWriter (io.prometheus.metrics.expositionformats) 
write:68, PrometheusTextFormatWriter (io.prometheus.metrics.expositionformats) 
write:48, PrometheusOutputFormat$1 (org.springframework.boot.actuate.metrics.export.prometheus) 
scrape:87, PrometheusScrapeEndpoint (org.springframework.boot.actuate.metrics.export.prometheus)
```
```
StreamEncoder public void write(int c) throws IOException {
  char[] cbuf = new char[1];
  cbuf[0] = (char) c;
```


zeitlinger:
Thanks for the contribution!

Issue #1246, opened 2025-01-03
Add weighted size counter to caffeine instrumentation


jhominal:
Hello,

I am using caffeine in a context where I am using a `Weigher` to bound the cache instance memory consumption.

In that context, I find that the weighted size (which can be found on the `EvictionPolicy` object in caffeine) is an interesting statistic to track.

I would be happy to author a PR to add that stat to `caffeine` instrumentation. (guava cache does not seem to expose an API that would make that statistic available)

zeitlinger:
Great - PR would be welcome :smile:

jhominal:
@zeitlinger While looking at the details of the metrics collected in `caffeine-instrumentation`, I have come to believe that `caffeine_cache_eviction_weight` should be a `Counter` instead of a `Gauge` (`Weigher.weigh` is annotated as returning a non-negative value.)

Would you approve a change of metric type from `gauge` to `counter`?
What provisions, if any, should be made for backward compatibility?

jhominal:
@zeitlinger Semantically the eviction weight is also described in the [`CacheStats` class doc string](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/stats/CacheStats.java#L46-L47)

/**
 * Statistics about the performance of a {@link Cache}.
 * <p>
 * Cache statistics are incremented according to the following rules:
 * <ul>
 *   <li>When an entry is evicted from the cache, {@code evictionCount} is incremented and the
 *       weight added to {@code evictionWeight}.
 * </ul>
```

That is, each time there is an eviction:

 * `evictionCount` is incremented by 1;
 * `evictionWeight` is incremented by the evicted entry's weight. By contract (on the [`Weigher`](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Weigher.java#L43) interface), the weight of an entry must be non-negative.

Because of that, I do believe that `evictionWeight` is a monotically increasing counter.

zeitlinger:
You're right :smile: 

@fstab do you think we can make a breaking change in a minor release?

jhominal:
In case we are not comfortable with making a breaking change in a minor release (which is quite understandable!), I am willing to make any changes necessary to ease such a migration:

 * Adding an optional constructor parameter to opt-in into the change from `gauge` to `counter`;
[...]

PR #1251, opened 2025-01-07
Caffeine instrumentation: add weighted size metric

jhominal:
There are two modifications in this PR:

 * `caffeine_cache_eviction_weight` has been changed from a `gauge` to a `counter` - that seems semantically more pertinent to me, as that statistic can only increase (though there is a debate on that point - see #1246);
 * `caffeine_cache_weighted_size` has been added;

There are at least two backwards compatibility concerns that should be given a thought:

 1. If the change of type of `caffeine_cache_eviction_weight` is accepted, users would have to migrate to using the `_total` suffix;
 2. If someone already uses `caffeine-instrumentation`, and has defined their own `caffeine_cache_weighted_size` metric, upgrading to a version that includes this PR would break their application;

Point 1 obviously depends on whether the type change is accepted, both from a theoretical point of view (how it should be modelled as a metric), and from a practical point of view (is the change worth a migration for current users).
For point 2, I guess my main point would be to understand the project's policy on backward compatibility when adding metric names in instrumentation libraries. I would be happy to add documentation pertaining to:

 * Guava and Caffeine instrumentation, and backwards compatibility policy as it pertains to new metric added by these instrumentation libraries in the future;
 * Backwards compatibility policy around adding new metrics to JVM instrumentation;

### DIFF
--- a/prometheus-metrics-instrumentation-caffeine/src/main/java/io/prometheus/metrics/instrumentation/caffeine/CacheMetricsCollector.java
+++ b/prometheus-metrics-instrumentation-caffeine/src/main/java/io/prometheus/metrics/instrumentation/caffeine/CacheMetricsCollector.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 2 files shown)


zeitlinger (approved):
LGTM

PR #1255, opened 2025-01-09
UTF-8 support in metric and label names

fedetorres93:
Adds UTF-8 support for metric and label names.

These changes are based on the work done on the Prometheus common libraries [here](https://github.com/prometheus/common/pull/537) and [here](https://github.com/prometheus/common/pull/570)

- The `prometheus-metrics-exposition-formats` module will use the new quoting syntax `{"foo"}` iff the metric does not conform to the legacy name format (`foo{}`)
- The `prometheus-metrics-model` module has a new flag (`NameValidationScheme`) that determines if validation is done using the legacy or the UTF-8 scheme. This flag can be set via a property in the properties file.
- Scrapers can announce via content negotiation that they support UTF-8 names by adding `escaping=allow-utf-8` in the Accept header. In cases where UTF-8 is not available, metric providers can be configured to escape names in a few different ways: values (`U__` UTF value escaping for perfect round-tripping), underscores (all invalid chars become `_`), dots (dots become `_dot_`, `_` becomes `__`, all other values become `___`). Escaping can either be a global default (`PrometheusNaming.nameEscapingScheme`) or can also be specified in Accept header with the `escaping=` term, which can be `allow-utf-8` (for UTF-8-compatible), `underscores`, `dots`, or `values`.
This should still be a noop for existing configurations because scrapers will not be passing the escaping key in the Accept header. Existing functionality is maintained.
- The `prometheus-metrics-exporter-pushgateway` module will [escape](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#text-escaping) UTF-8 grouping keys in the URL path used when pushing metrics (see https://github.com/prometheus/pushgateway/pull/689)

Work towards https://github.com/prometheus/common/issues/527


fedetorres93:
Given the ongoing discussion about unit suffixes for OM 2.0 (https://github.com/prometheus/OpenMetrics/issues/286), I think we can take this UTF-8 work as a basis and then add the necessary changes to comply with the final consensus on suffixes.

zeitlinger:
> Given the ongoing discussion about unit suffixes for OM 2.0 ([prometheus/OpenMetrics#286](https://github.com/prometheus/OpenMetrics/issues/286)), I think we can take this UTF-8 work as a basis and then add the necessary changes to comply with the final consensus on suffixes.

@fstab are you ok with that?

fstab:
Update: The `client_java` maintainers just decided that we wait until 1 April 2025. If we have OpenMetrics 2.0 by then we will implement that. If OpenMetrics 2.0 is still under discussion in 2.0 we will merde this PR.

fedetorres93:
Hello @fstab, I just wanted to follow-up on your last comment. Seems like OM 2.0 is still under discussion, so do you think now is a good time to reconsider merging this PR?

fstab:
We have our `client_java` community call tomorrow, and can discuss this there. If you have time, feel free to join https://calendar.google.com/calendar/u/0/embed?src=prometheus.io_bdf9qgm081nrd0fe32g3olsld0%40group.calendar.google.com

fedetorres93:
@zeitlinger Thanks for the feedback. I made the statics in `PrometheusNaming` final as you suggested. 

About adding a setting in `ExporterOpenTelemetryProperties`, IIUC that module is translating from Prometheus to OTel format, so it should continue working as it is now. Prometheus' UTF-8 configs affect translations from OTel to Prometheus.

prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/NamingProperties.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private NamingProperties(String validation) {
+    this.validationScheme = validation;
+  }
+
+  public String getValidationScheme() {
+    return validationScheme;
+  }
+
+  static NamingProperties load(String prefix, Map<Object, Object> properties) throws PrometheusPropertiesException {
```
zeitlinger: the prefix is always the same

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java
```diff
@@ -41,11 +77,30 @@ public class PrometheusNaming {
     ".total", ".created", ".bucket", ".info"
   };
 
+  static ValidationScheme initValidationScheme() {
+    if (PrometheusProperties.get() != null && PrometheusProperties.get().getNamingProperties() != null) {
```
zeitlinger: can't be null (both)

prometheus-metrics-exporter-pushgateway/src/test/java/io/prometheus/metrics/exporter/pushgateway/PushGatewayTest.java
```diff
@@ -141,6 +148,26 @@ public void testPushWithGroupingKey() throws IOException {
     pg.push();
   }
 
+  @Test
+  public void testPushWithEscapedGroupingKey() throws IOException {
+    try (MockedStatic<PrometheusNaming> mock = mockStatic(PrometheusNaming.class, CALLS_REAL_METHODS)) {
```
zeitlinger: I wonder if the tests would be easier to understand by just setting the sys property directly - it's documenting how to use it 

We could add a `resetForTest` method to `PrometheusProperties`

zeitlinger:
> PrometheusNaming.nameEscapingScheme

I can't find that

prometheus-metrics-exporter-pushgateway/src/main/java/io/prometheus/metrics/exporter/pushgateway/PushGateway.java
```diff
@@ -206,7 +209,7 @@ private void doRequest(PrometheusRegistry registry, String method) throws IOExce
       try {
         if (!method.equals("DELETE")) {
           OutputStream outputStream = connection.getOutputStream();
-          writer.write(outputStream, registry.scrape());
+          writer.write(outputStream, registry.scrape(), EscapingScheme.NO_ESCAPING);
```
zeitlinger: should the scheme be configured in the properties?

prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/NameType.java
```diff
@@ -0,0 +1,6 @@
+package io.prometheus.metrics.expositionformats;
+
+public enum NameType {
```
zeitlinger:
```suggestion
enum NameType {
```

prometheus-metrics-exposition-textformats/src/test/java/io/prometheus/metrics/expositionformats/ExpositionFormatsTest.java
```diff
@@ -2737,11 +2766,86 @@ public void testLabelValueEscape() throws IOException {
     assertPrometheusText(prometheus, counter);
   }
 
+  @Test
+  public void testFindWriter() {
+    ExpositionFormats expositionFormats = ExpositionFormats.init();
+
+    // delimited format - should use default escaping scheme
```
zeitlinger: good candidate for parameterized test - e.g. like here: https://github.com/prometheus/client_java/blob/3133f66c35dbe9dfa4bdfbc661a4b6580851521b/integration-tests/it-exporter/it-exporter-test/src/test/java/io/prometheus/metrics/it/exporter/test/ExporterIT.java#L92

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java
```diff
@@ -280,4 +371,368 @@ private static String replaceIllegalCharsInUnitName(String name) {
...
+        }
+        if (l.getName() == null || isValidLegacyMetricName(l.getName())) {
+          outLabelsBuilder.label(l.getName(), l.getValue());
+          continue;
+        }
+        outLabelsBuilder.label(escapeName(l.getName(), scheme), l.getValue());
+      }
+
+      Labels outLabels = outLabelsBuilder.build();
+      DataPointSnapshot outDataPointSnapshot = null;
```
zeitlinger: better extract a method that returns the snapshot rather than reassigning the variable

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java
```diff
@@ -280,4 +371,368 @@ private static String replaceIllegalCharsInUnitName(String name) {
...
+          .value(((UnknownSnapshot.UnknownDataPointSnapshot) d).getValue())
+          .exemplar(((UnknownSnapshot.UnknownDataPointSnapshot) d).getExemplar())
+          .scrapeTimestampMillis(d.getScrapeTimestampMillis())
+          .build();
+      }
+
+      outDataPoints.add(outDataPointSnapshot);
+    }
+
+    MetricSnapshot out;
```
zeitlinger: same here: better a method that returns the snapshot

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/PrometheusNamingTest.java
```diff
@@ -96,4 +104,435 @@ public void testEmptyUnitName() {
...
+    assertThat(isValidLabelName("a lid_23name")).isFalse();
+    assertThat(isValidLabelName(":leading_colon")).isFalse();
+    assertThat(isValidLabelName("colon:in:the:middle")).isFalse();
+    assertThat(isValidLabelName("a\ud800z")).isFalse();
+  }
+
+  @Test
+  public void testEscapeName() {
+    // empty string
+    String got = escapeName("", EscapingScheme.UNDERSCORE_ESCAPING);
```
zeitlinger: avoid reassigning variables - I think the expression for `got` can be inlined directly

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/PrometheusNamingTest.java
```diff
@@ -96,4 +104,435 @@ public void testEmptyUnitName() {
...
+    assertThat(got).isEqualTo("");
```
zeitlinger:
```suggestion
    assertThat(got).isEmpty();
```

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/PrometheusNamingTest.java
```diff
@@ -96,4 +104,435 @@ public void testEmptyUnitName() {
...
+      assertThat(data.getValue()).isEqualTo(34.2);
+      assertThat((Iterable<? extends Label>) data.getLabels()).isEqualTo(Labels.builder()
+        .label("__name__", "my_metric")
+        .label("some_label", "labelvalue")
+        .build());
+    }
+  }
+
+  @Test
+  public void testEscapeMetricSnapshotLabelNameEscapingNeeded() {
```
zeitlinger: almost a copy of the previous method - could extract common logic

zeitlinger (commented):
great - I think I get it now 😄

prometheus-metrics-exporter-pushgateway/src/test/java/io/prometheus/metrics/exporter/pushgateway/PushGatewayTest.java
fedetorres93: I considered using that approach but then realized that the fact the fields involved were static made that option unfeasible, and using reflection didn’t seem appropriate either, since IIUC it’s restricted in recent versions of the JVM.

If there’s another way to approach this that I’m not considering, please let me know, since I don’t have much experience doing this kind of testing in Java.

zeitlinger: I've made an attempt and pushed it to this PR - please double check that the tests still make sense.
I also added more parameterized tests, because I couldn't follow them.

zeitlinger:
From review with @fstab 

Conceptually

1. we should consider getting rid of the validation scheme: the java client is mostly used indirectly, e.g. from spring, JMX exporter, OTel SDK - where it's strange to have to pass a sys property to unlock UTF-8 chars in metric names and labels
2. [sanitizeMetricName](https://github.com/prometheus/client_java/blob/bc859f3e6c75cbcad0ec7715b74aadf5517053eb/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java#L228-L232) should not replace UTF-8 chars anymore
3. `allow-utf-8` escaping in text formats must still escape some characters like whitespace, newlines, `{` - this is currently not done

Minor issues

4. `legacy` is an ambiguous name - better to have `nonUtf8` or similar
5. move escaping related methods (`escapeMetricSnapshot`) to separate utils class

ywwg:
> * `legacy` is an ambiguous name - better to have `nonUtf8` or similar

"legacy" is the term used in upstream prometheus libraries, for better or worse.  It's short for "legacy valid prometheus character set".  "nonutf8" doesn't quite capture the idea either, because "." is not really a "UTF-8" character.

fedetorres93:
> allow-utf-8 escaping in text formats must still escape some characters like whitespace, newlines, { - this is currently not done

@zeitlinger AFAIU only backslashes (`\`) newlines (`\n`) and double quotes (`"`) should be escaped, other UTF-8 characters are valid with the new quoting syntax.

zeitlinger:
> > allow-utf-8 escaping in text formats must still escape some characters like whitespace, newlines, { - this is currently not done
> 
> @zeitlinger AFAIU only backslashes (`\`) newlines (`\n`) and double quotes (`"`) should be escaped, other UTF-8 characters are valid with the new quoting syntax.

yes, this is also the outcome of the meeting yesterday :smile:

prometheus-metrics-exporter-pushgateway/src/main/java/io/prometheus/metrics/exporter/pushgateway/PushGateway.java
```diff
@@ -432,6 +437,25 @@ private String getJob(ExporterPushgatewayProperties properties) {
...
+            return EscapingScheme.VALUE_ENCODING_ESCAPING;
+          case "underscores":
+            return EscapingScheme.UNDERSCORE_ESCAPING;
+          case "dots":
+            return EscapingScheme.DOTS_ESCAPING;
+          default:
+            return EscapingScheme.NO_ESCAPING;
+        }
+      }
+      return EscapingScheme.NO_ESCAPING;
```
zeitlinger: is it OK that the  push gateway defaults to no-escape?

fedetorres93: Thanks for pointing it out, you're right, it should default to underscores to match the original behavior.

zeitlinger: OK - and I guess you should write `allow-utf-8` in the properties file instead of `no-escaping` to be consistent with the accept header.

fedetorres93: Yes, that's right

prometheus-metrics-model/src/test/resources/prometheus.properties
```diff
@@ -0,0 +1 @@
+io.prometheus.naming.validationScheme=legacy
```
fstab: This file is no longer needed and can be removed.

PR #1262, opened 2025-01-19
Add documentation pages for Caffeine and Guava cache instrumentation libraries

jhominal:
After working on #1251, I think that the proposed changes would warrant some kind of notice in the project documentation, and in order to do that I thought it would be cleaner to write a documentation page for caffeine instrumentation to serve as a baseline.

As the guava cache instrumentation is very similar to caffeine, I decided to write a similar documentation page for guava.

### DIFF
--- a/docs/content/instrumentation/caffeine.md
+++ b/docs/content/instrumentation/caffeine.md
@@ -1,0 +1,101 @@
+---
+title: Caffeine Cache
+weight: 1
+---
+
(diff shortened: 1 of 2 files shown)


zeitlinger:
@jhominal do you want to update this PR to reflect https://github.com/prometheus/client_java/pull/1251?

jhominal:
@zeitlinger yes, I will do that later today :)

zeitlinger:
thanks for the contribution :smile:

PR #1266, opened 2025-01-22
Dropwizard apply label mapper to snapshot metrics

KacperFilutowski-TomTom:
Also allowed to override getHelpMessage


zeitlinger:
allowing subclassing is a slippery slope - I'd rather accept a help message customizer function in a builder

KacperFilutowski-TomTom:
> allowing subclassing is a slippery slope - I'd rather accept a help message customizer function in a builder

I reverted the getHelpMessage change as it is only a nice to have.
I might follow up on this in a separate PR

Issue #1277, opened 2025-02-18
TextFormatUtil.writeTimestamp writes a value with a decimal point but the spec specifies an int64


garthy:
The method here https://github.com/prometheus/client_java/blob/7016bf7c5749e806ef91a918fdb294d394b83a16/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/TextFormatUtil.java#L24C15-L35

```  
static void writeTimestamp(Writer writer, long timestampMs) throws IOException {
    writer.write(Long.toString(timestampMs / 1000L));
    writer.write(".");
```

Writes a decimal point.

The spec https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details

specifies the timestamp as an int64 

```
The timestamp is an int64 (milliseconds since epoch, i.e. 1970-01-01 00:00:00 UTC, excluding leap seconds), represented as required by Go's [ParseInt()](https://golang.org/pkg/strconv/#ParseInt) function.
```

This causes Prometheus/promtool to error on metrics with a timestamp

```
error while linting: text format parsing error in line 17: expected integer as timestamp, got "1739889638.000"
```

zeitlinger:
@fstab do you have an idea why we might have added the decimal point?

garthy:
Hi any idea when this is getting in a Release? Thanks

zeitlinger:
It's here: https://github.com/prometheus/client_java/releases/tag/v1.3.7

PR #1292, opened 2025-02-27
fix timestamp representation

zeitlinger:
Fixes https://github.com/prometheus/client_java/issues/1277


zeitlinger:
@dhoard @fstab 

this changes quite a lot - am I on the wrong track?

E.g. this would change in ExpositionFormatsTest:

image

dhoard:
It feels like the correct approach to me.

zeitlinger:
Note: this change should only affect prom format - not openmetrics - see https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#timestamps

PR #1293, opened 2025-02-28
add option to skip the shaded protobuf libraries

zeitlinger:


soc:
Is there any information to understand why this spins up a separate artifact by hand and introduces a different module name, instead of being a standard classifier?

zeitlinger:
> Hey, is there any information to read up on to understand why this spins up a separate artifact by hand and introduces a different module name, instead of being a standard classifier?
> 
> Thanks!

I had not considered that option - sounds much better actually :smile: 

I think this is so new that we can still make that change - a PR would be highly welcome!

soc:
@zeitlinger I'd love to, though I'm also not an expert, but I'm away from development for a while starting *today*. :-(

zeitlinger:
I tried the following - but it doesn't work because both maven projects still share the same pom

```sh
ll ~/.m2/repository/io/prometheus/prometheus-metrics-exposition-formats/1.4.0-SNAPSHOT 
total 2,0M
-rw-rw-r-- 1 gregor gregor 1,2K Mai 16 17:28 maven-metadata-local.xml
-rw-rw-r-- 1 gregor gregor 174K Mai 16 17:28 prometheus-metrics-exposition-formats-1.4.0-SNAPSHOT.jar
-rw-rw-r-- 1 gregor gregor 172K Mai 16 17:28 prometheus-metrics-exposition-formats-1.4.0-SNAPSHOT-no-protobuf.jar
-rw-rw-r-- 1 gregor gregor 4,9K Mai 16 17:26 prometheus-metrics-exposition-formats-1.4.0-SNAPSHOT.pom
-rw-rw-r-- 1 gregor gregor 1,6M Mai 16 17:21 prometheus-metrics-exposition-formats-1.4.0-SNAPSHOT-sources.jar
```

```patch
Index: prometheus-metrics-exposition-formats/pom.xml
IDEA additional info:
<+>UTF-8
diff --git a/prometheus-metrics-exposition-formats/pom.xml b/prometheus-metrics-exposition-formats/pom.xml
--- a/prometheus-metrics-exposition-formats/pom.xml	(revision 523ed272db0d98df55986bf3cafd74c5e5633647)
+++ b/prometheus-metrics-exposition-formats/pom.xml	(date 1747409199333)
@@ -10,7 +10,7 @@
     <version>1.4.0-SNAPSHOT</version>
   </parent>
 
-  <artifactId>prometheus-metrics-exposition-formats-no-protobuf</artifactId>
+  <artifactId>prometheus-metrics-exposition-formats</artifactId>
   <packaging>bundle</packaging>
 
   <name>Prometheus Metrics Exposition Formats no protobuf</name>
@@ -63,6 +63,22 @@
 
   <build>
     <plugins>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <configuration>
+              <classifier>no-protobuf</classifier>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>build-helper-maven-plugin</artifactId>
```

zeitlinger:
> @zeitlinger
> 
> Hello, I am very sorry, probably I missed something, but... Is it still possible to use "shaded" protobuf? Without need to always use `prometheus-metrics-exposition-formats-no-protobuf` (with regular protobuf dependency)?

yes - this is still the default

Can you create a new issue and add the dependency tree?

PR #1335, opened 2025-04-25
Eagerly reject null label values

benjaminp:
Scraping generally doesn't support null label values and can throw NPEs at various points. It's easiest to debug such problems at the point null is introduced.

Signed-off-by: Benjamin Peterson <benjamin@engflow.com>


zeitlinger:
looks good - can you add a test case?

benjaminp:
Thank you for the review. Test added

zeitlinger:
thanks for the contribution

PR #1339, opened 2025-04-30
Introduce removeIf method to StatefulMetric

dao-jun:
Introduce removeIf method to StatefulMetric that it could help users to clear DataPoints.


zeitlinger:
what is your (real) use case where this would help?

dao-jun:
> what is your (real) use case where this would help?

@zeitlinger 
Our case is in message system(Kafka on Pulsar, support Kafka protocol on Apache Pulsar and to provide a message system that separates storage and computation).

We need to expose consumer group level metrics. Consumer groups hosted by a topic-partition via hash algorithm, the metrics just like `kafka_consumergroup_current_offset(group = x, topic = y, partition = z)`

The topic-partition can be unloaded by loadbalancer and transfer to another broker, so when we observed the topic-partition being unloaded, we have to remove all the metric children which their consumergroup hosted by the topic-partition. To avoid the broker exports the metrics which the groups are  not hosted by this broker and reduce the TSDB pressure.

In one word, we need a method and it can iter the metric children, so we can remove the metric children with specified more efficiently.

zeitlinger:
> > what is your (real) use case where this would help?
> 
> @zeitlinger Our case is in message system(Kafka on Pulsar, support Kafka protocol on Apache Pulsar and to provide a message system that separates storage and computation).
> 
> We need to expose consumer group level metrics. Consumer groups hosted by a topic-partition via hash algorithm, the metrics just like `kafka_consumergroup_current_offset(group = x, topic = y, partition = z)`
> 
> The topic-partition can be unloaded by loadbalancer and transfer to another broker, so when we observed the topic-partition being unloaded, we have to remove all the metric children which their consumergroup hosted by the topic-partition. To avoid the broker exports the metrics which the groups are not hosted by this broker and reduce the TSDB pressure.
> 
> In one word, we need a method and it can iter the metric children, so we can remove the metric children with specified labels more efficiently.

Thanks for the explanation - makes sense now!

The tests are still failing.

dao-jun:
@zeitlinger I've fixed the DCO check, PTAL

Issue #1353, opened 2025-05-14
Provide means for DropwizardExports to deal with data point validation errors


mnssn:
Hello!

In its current form the `DropwizardExports.collect()` throws an exception if a registered metric contains an invalid data point (e.g. a Counter with negative value).

```java
// file: DropwizardExportsTest.java

@Test
void collectInvalidMetricFails() {
    metricRegistry.counter("my.application.namedCounter1").inc(-10);
    assertThatThrownBy( () -> convertToOpenMetricsFormat() ).isInstanceOf( IllegalArgumentException.class );
}
```

In large applications, it may be preferable to gracefully degrade the available metrics and report the successful subset of metrics, while reporting the failure via e.g. logging. This would allow operators to maintain some visibility into the application state, while at the same time becoming aware of the "invalid metric".

There seems to be a certain precedence for this, since `DropwizardExports.fromGauge` currently ignores invalid gauges rather than failing the entire collection call.

PR #1354, opened 2025-05-14
Let clients of DropwizardExports handle validation errors

mnssn:
A possible solution for #1353.

In the current form, if e.g. a Counter contains invalid data the entire `collect` calls fails with InvalidArgumentException.

In large applications, it may be preferable to gracefully degrade the available metrics and report the failure via e.g. logging. This change allows just that - by changing the access modifiers to the `fromXYZ` methods subclasses can now handle the errors.

Finally, to allow for skipping of invalid metrics, we follow the example in `fromGauge` to interpret the meaning of a returned null to be that the metric should be skipped.

### DIFF
--- a/prometheus-metrics-instrumentation-dropwizard/src/main/java/io/prometheus/metrics/instrumentation/dropwizard/DropwizardExports.java
+++ b/prometheus-metrics-instrumentation-dropwizard/src/main/java/io/prometheus/metrics/instrumentation/dropwizard/DropwizardExports.java
@@ -9,6 +9,7 @@ import com.codahale.metrics.MetricFilter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.Timer;
+import io.prometheus.metrics.instrumentation.dropwizard5.InvalidMetricHandler;
 import io.prometheus.metrics.instrumentation.dropwizard5.labels.CustomLabelMapper;
 import io.prometheus.metrics.model.registry.MultiCollector;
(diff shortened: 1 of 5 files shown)


zeitlinger:
I like the idea of making this customizable.

Instead of subclassing, I'd prefer to use a callback and builder - would that work?

mnssn:
Yes - that is a much cleaner solution. I have updated the PR to allow the callback to decide if an exception is to be suppressed.

prometheus-metrics-instrumentation-dropwizard/src/main/java/io/prometheus/metrics/instrumentation/dropwizard/DropwizardExports.java
```diff
@@ -67,6 +73,23 @@ public DropwizardExports(
...
     this.labelMapper = Optional.ofNullable(labelMapper);
+    this.invalidMetricHandler = InvalidMetricHandler.ALWAYS_THROW;
+  }
+
+  /**
+   * @param registry a metric registry to export in prometheus.
+   * @param metricFilter a custom metric filter.
+   * @param labelMapper a labelMapper to use to map labels.
+   */
+  public DropwizardExports(
```
zeitlinger, line 84: this ctor should be private - i.e. you have to use the builder for that

mnssn: Fixed.

PR #1450, opened 2025-07-05
Add `prometheus-metrics-exposition-textformats` to BOM

technicallyjames:
Hi all, I noticed a reference to `prometheus-metrics-exposition-textformats` on https://prometheus.github.io/client_java/exporters/formats/, but I didn't see the dependency in the BOM, so I'm adding it now.

@fstab @zeitlinger @dhoard @tomwilkie Please take a look. Thanks!

### DIFF
--- a/prometheus-metrics-bom/pom.xml
+++ b/prometheus-metrics-bom/pom.xml
@@ -81,6 +81,11 @@
       </dependency>
       <dependency>
         <groupId>io.prometheus</groupId>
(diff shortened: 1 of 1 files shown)


zeitlinger:
Thanks - I think we need a unit test so that we don't forget that in the future - I'll add that, too.

zeitlinger:
closed in favor of https://github.com/prometheus/client_java/pull/1451

zeitlinger:
is released in https://github.com/prometheus/client_java/releases/tag/v1.3.10

PR #1471, opened 2025-07-21
Discussion, improve CounterDataPoint.inc() critical path

cruftex:
@fstab, @dhoard, @tomwilkie, @zeitlinger

Thanks for the cool library. Its a pleasure to work with!

I was using it for a new project and started with placing own `LongAdder` in the code and expose via the `collect()` method. This way, I would I could manipulate it directly in the class implementing the service resulting in maximum performance. The code started to be come repetitive and add some point I realised that I am starting to implement something like `CounterDataPoint` again. To avoid this, I started digging into the code to see what the overhead would be.

With a code like `counter.labelValues().inc(1)` there is a bit of indirection until we hit `LongAdder`. However, there is already a faster path as the documentation in `CounterDataPoint` describes:

````
CounterDataPoint pendingTasks = counter.labelValues("pending");
pendingTasks.inc();
````

Looking deeper into the code there is a tiny bit more indirection, which could be avoided.

This patch is returning a specialised class in case the exemplars support is not enabled. The intended usage is:

````
    Counter counter =
      Counter.builder()
        .name("count_total")
        .withoutExemplars()
        .build();
    Counter.DataPointIgnoringExemplars fasterCounter = (Counter.DataPointIgnoringExemplars)
      counter.labelValues();
    fasterCounter.inc(1);
````

The unit test run successfully, however I realise that its breaking with the concept that the configuration can override what is configured in the code. I am not sure whether this would be tolerable for that special case. If an SRE enables exemplars for further diagnosis, the application would crash completely. Not good. However, probably 

The patch is still worth discussion and worth applying without propagating to use `Counter.DataPointIgnoringExemplars` class directly. This class can be private to avoid wrong usage. Which brings me to the class visibility of `Counter.DataPoint`.

My IDE marks code with:

Screenshot_20250721_135228

The problem is that it's used as type parameter in the main `Counter` class and therefor becomes visible. My recommendation would be to make this public, deprecate the interface `CounterDataPoint` and mandate to use `Counter.DataPoint` instead. That is painful and probably will not happen fast, however, it is a mismatch, since `Counter` is a class that is not open for extension and the only one that will ever provide a `CounterDataPoint`.

The patch also removes two unnecessary boolean fields. If that makes sense, I can also extract that as a housekeeping PR.

Thoughts?

### DIFF
--- a/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Counter.java
+++ b/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Counter.java
@@ -38,7 +38,11 @@ public class Counter extends StatefulMetric<CounterDataPoint, Counter.DataPoint>
     super(builder);
     MetricsProperties[] properties = getMetricProperties(builder, prometheusProperties);
     boolean exemplarsEnabled = getConfigProperty(properties, MetricsProperties::getExemplarsEnabled);
-    if (exemplarsEnabled) {
+    // exemplars might be enabled specifically for a metric, however, if the code
+    // says withoutExemplars they should stay disabled.
+    boolean notTurnedOffWithinCode =
+      builder == null || builder.exemplarsEnabled == null || builder.exemplarsEnabled;
+    if (exemplarsEnabled && notTurnedOffWithinCode) {
       exemplarSamplerConfig =
           new ExemplarSamplerConfig(prometheusProperties.getExemplarProperties(), 1);
     } else {
--- a/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Counter.java
+++ b/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Counter.java
@@ -101,7 +101,7 @@ public class Counter extends StatefulMetric<CounterDataPoint, Counter.DataPoint>
     if (isExemplarsEnabled()) {
       return new DataPoint(new ExemplarSampler(exemplarSamplerConfig));
     } else {
-      return new DataPoint(null);
+      return new DataPointIgnoringExemplars();
     }
   }
 
@@ -112,7 +112,7 @@ public class Counter extends StatefulMetric<CounterDataPoint, Counter.DataPoint>
     return name;
   }
 
-  class DataPoint implements CounterDataPoint {
+  public static class DataPoint implements CounterDataPoint {
 
     private final DoubleAdder doubleValue = new DoubleAdder();
     // LongAdder is 20% faster than DoubleAdder. So let's use the LongAdder for long observations,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by cruftex** [comment]
@fstab
Hmm.... According to the code comment actually `withoutExemplars()` should turn it off. Then I would not expect that it can turned on by the configuration again. I did another updated, so that for counter exemplars flag the code always takes precedence. Maybe that should be the case for all metrics?

````
    /** Allow Exemplars for this metric. */
    public B withExemplars() {
      this.exemplarsEnabled = true;
      return self();
    }

    /** Turn off Exemplars for this metric. */
    public B withoutExemplars() {
      this.exemplarsEnabled = false;
      return self();
    }
````

File: prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Counter.java [hunk 1]
```diff
@@ -32,14 +32,17 @@
...
     super(builder);
     MetricsProperties[] properties = getMetricProperties(builder, prometheusProperties);
-    exemplarsEnabled = getConfigProperty(properties, MetricsProperties::getExemplarsEnabled);
-    if (exemplarsEnabled) {
+    boolean exemplarsEnabled = getConfigProperty(properties, MetricsProperties::getExemplarsEnabled);
+    // exemplars might be enabled specifically for a metric, however, if the code
+    // says withoutExemplars they should stay disabled.
+    boolean notTurnedOffWithinCode =
+      builder == null || builder.exemplarsEnabled == null || builder.exemplarsEnabled;
+    if (exemplarsEnabled && notTurnedOffWithinCode) {
```
zeitlinger, line 45:
```suggestion
    if (exemplarsEnabled && !Objects.equals(builder.exemplarsEnabled, false)) {
```

zeitlinger: builder can't be null (adding NullAway is on my list...)

cruftex:
I think its better to address the `withoutExemplars` first and separately:

https://github.com/prometheus/client_java/pull/1477

Regarding the counter optimization, I think this code looks rather ugly and hurts a lot:

````
    Counter counter =
      Counter.builder()
        .name("count_total")
        .withoutExemplars()
        .build();
    Counter.DataPointIgnoringExemplars fasterCounter = (Counter.DataPointIgnoringExemplars)
      counter.labelValues();
    fasterCounter.inc(1);
````

### Better idea:

````
   HotCounter counter =
      HotCounter.builder()
        .name("count_total")
        .build();
   HotCounter.DataPoint fasterCounter = counter.labelValues();
    fasterCounter.inc(1);
````

While       


````
Counter.builder()
        .name("count_total")
        .withoutExemplars()
        .build
````

Would also return a `HotCounter`.

zeitlinger:
> ... Would also return a `HotCounter`.

sounds good :smile:

PR #1477, opened 2025-07-23
Builder.withoutExemplars() setting cannot be overridden via properties

cruftex:
The PR will fix the behaviour that exemplar support can be switched on via the properties, although it was disabled explicitly in the code definition by `withoutExemplars`. Currently this would enable exemplars:

````
    PrometheusProperties properties = PrometheusProperties.builder()
      .putMetricProperty("count", MetricsProperties.builder()
        .exemplarsEnabled(true)
        .build())
      .build();
    Counter counter =
      Counter.builder(properties)
        .name("count_total")
        .withoutExemplars()
        .build();
````

In TDD fashion I will add failing tests first.

### DIFF
--- a/prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/PrometheusProperties.java
+++ b/prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/PrometheusProperties.java
@@ -36,6 +36,10 @@ public class PrometheusProperties {
     return instance;
   }
 
+  public static Builder builder() {
+    return new Builder();
+  }
+
   public PrometheusProperties(
       MetricsProperties defaultMetricsProperties,
       Map<String, MetricsProperties> metricProperties,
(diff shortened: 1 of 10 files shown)


cruftex:
@zeitlinger 
This PR is now ready for review and merging.

Somehow the build is not starting. Is there something else I need to do?

zeitlinger:
> Somehow the build is not starting. Is there something else I need to do?

no - that needs to be approved

zeitlinger:
now there's a checkstyle issue

[WARN] /home/runner/work/client_java/client_java/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Summary.java:52:5: Distance between variable 'exemplarsEnabled' declaration and its first usage is 4, but allowed 3.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value). [VariableDeclarationUsageDistance]
[WARN] /home/runner/work/client_java/client_java/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java:118:5: Distance between variable 'exemplarsEnabled' declaration and its first usage is 7, but allowed 3.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value). [VariableDeclarationUsageDistance]
Audit done.

cruftex:
@zeitlinger ping

PR #1493, opened 2025-08-06
Update PrometheusScrapeHandler compression handling

### LINKED ISSUE #1488: PrometheusScrapeHandler should compress the response on the fly if it is required
current process seems like this:

- create plain response in memory
- check if response should be compressed
  - compress the memory response

instead, it would be desirable to do

- check if response should be compressed
- create zip stream based on the memory stream
- save the output into the zipped stream

This would save the memory, especially when there are huge output

Thanks,
jerry

jaydeluca:
Resolves #1488 

Don't create response buffer for compressed responses


zeitlinger:
great :smile:

Issue #1498, opened 2025-08-08
Feature Request: Support `createdTimestampMillis` in `PrometheusProtobufWriterImpl`


cmaune:
**Background:**

We're using **Spring Boot 3.5.x** with **Micrometer 1.15.2**, and would like to leverage the [`createdTimestamp`](https://prometheus.io/docs/prometheus/latest/feature_flags/) feature introduced behind a feature flag on the Prometheus server.

On Prometheus server side this feature is only available if the Prometheus Protobuf Protocol is used.

In the Prometheus Java client, we noticed that the [`PrometheusProtobufWriterImpl`](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/internal/PrometheusProtobufWriterImpl.java) currently only maps `scrapeTimestampMillis` from the [`DataPointSnapshot`](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/DataPointSnapshot.java), but **not** the available `createdTimestampMillis`.

**What we did:**

We tested adding support for `createdTimestampMillis` in a fork of the latest release and confirmed that our counter metrics are now exposed with an initialized value of `0` as expected. This also builds on recent support added to Micrometer for this field ([micrometer-metrics/micrometer#2625](https://github.com/micrometer-metrics/micrometer/issues/2625)).

**Proposal:**

We propose updating `PrometheusProtobufWriterImpl` to also map and include `createdTimestampMillis`, similar to how `scrapeTimestampMillis` is currently handled. This would improve compatibility with Prometheus’ feature flag and help downstream libraries like Micrometer to fully support this newer functionality.

**Would you be open to a PR for this?** We're happy to contribute the necessary change if desired.

Thanks for your great work on this library!

jaydeluca:
fixed by #1532

PR #1532, opened 2025-08-22
feat: adds createdTimestamp to counters exposed via prometheus protouf protocol

cmaune:
#1498


zeitlinger (approved):

zeitlinger:
> mvn clean install

works for me locally - but I've also had that in the past sometimes

PR #1561, opened 2025-09-10
avoid the breaking change in PrometheusNaming

zeitlinger:
PrometheusNaming changes were incompatible - so deprecating instead and create e new PrometheusNames

### DIFF
--- a/prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/CallbackMetricTest.java
+++ b/prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/CallbackMetricTest.java
@@ -2,6 +2,9 @@ package io.prometheus.metrics.core.metrics;
 
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
+import io.prometheus.metrics.expositionformats.ExpositionFormats;
+import io.prometheus.metrics.model.registry.PrometheusRegistry;
+import java.io.IOException;
 import org.junit.jupiter.api.Test;
 
 class CallbackMetricTest {
@@ -45,5 +48,16 @@ class CallbackMetricTest {
         .withMessage(
             "Cannot pass label values to a CounterWithCallback that was created without label"
                 + " names.");
+  }
+
+  @Test
+  void manual() throws IOException {
+    PrometheusRegistry prometheusRegistry = new PrometheusRegistry();
+    Counter counter =
+        Counter.builder().name("te:st").labelNames("test").register(prometheusRegistry);
+    counter.labelValues("te:st").inc();
+    ExpositionFormats.init()
+        .getPrometheusTextFormatWriter()
+        .write(System.out, prometheusRegistry.scrape());
   }
 }
--- a/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNames.java
+++ b/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNames.java
@@ -1,0 +1,356 @@
+package io.prometheus.metrics.model.snapshots;
+
+import static java.lang.Character.MAX_CODE_POINT;
+import static java.lang.Character.MAX_LOW_SURROGATE;
+import static java.lang.Character.MIN_HIGH_SURROGATE;
+
+import io.prometheus.metrics.config.EscapingScheme;
+import java.nio.charset.StandardCharsets;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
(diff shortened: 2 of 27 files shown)


prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/CallbackMetricTest.java
```diff
@@ -46,4 +49,15 @@ void makeLabels() {
...
+
+  @Test
+  void manual() throws IOException {
+    PrometheusRegistry prometheusRegistry = new PrometheusRegistry();
+    Counter counter =
+        Counter.builder().name("te:st").labelNames("test").register(prometheusRegistry);
+    counter.labelValues("te:st").inc();
+    ExpositionFormats.init()
+        .getPrometheusTextFormatWriter()
+        .write(System.out, prometheusRegistry.scrape());
```
jaydeluca: should there be an assert here or are we just testing that no exception is thrown?

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNames.java
```diff
@@ -0,0 +1,356 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Utility for Prometheus Metric and Label naming.
+ *
+ * <p>Note that this library allows dots in metric and label names. Dots will automatically be
+ * replaced with underscores in Prometheus exposition formats. However, if metrics are exposed in
+ * OpenTelemetry format the dots are retained.
+ */
+public class PrometheusNames {
```
jaydeluca, line 19: Is there a name we could use that has a more drastic contrast to `PrometheusNaming`? Or maybe even `PrometheusNamingV2`? while reviewing I got confused as to which is which a few times

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNames.java
```diff
@@ -0,0 +1,356 @@
...
+          }
+          i += Character.charCount(c);
+        }
+        return escaped.toString();
+      default:
+        throw new IllegalArgumentException("Invalid escaping scheme " + scheme);
+    }
+  }
+
+  public static boolean needsEscaping(String name, EscapingScheme scheme) {
```
jaydeluca: just want to double check that all the public methods in this class are intentionally public?

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/PrometheusNamesTest.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class PrometheusNamesTest {
+
+  @Test
+  public void testSanitizeMetricName() {
```
jaydeluca: nit: none of these methods need to be public in the test classes

```suggestion
 void testSanitizeMetricName() {
```

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/CallbackMetricTest.java
zeitlinger: that was committed by mistake :upside_down_face:

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNames.java
zeitlinger: yeah - that seems to be the hardest part

- PrometheusUnicodeNaming
- same name but different package
- sanitizer interface: `PrometheusNaming.UNICODE.sanitize(..)` 
  - would also work if we want to have a variant that allows `_total` suffix in combination with https://github.com/prometheus/client_java/pull/1557

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNames.java
zeitlinger: good catch - and I also moved it to `NameEscaper`

I'm a bit hesitant to add internal classes now - since it would be potentially a breaking change
Alternatively, only new classes as internal classes - but that might make it look like the other classes are all public - which they are not.

PR #1772, opened 2026-01-08
Prevent exception in HttpServer exporter when using custom root metrics path

Donnerbart:
The behavior of `HttpServer.createContext()` was changed between Java 11 and 21. In the current JDKs the server doesn't allow to register multiple contexts with the same path. So if you configure
```java
HTTPServer.builder()
            .port(0)
            .registry(registry)
            .metricsHandlerPath("/")
            .buildAndStart()
```
you get the following exception:
```
java.lang.IllegalArgumentException: cannot add context to list
	at jdk.httpserver/sun.net.httpserver.ContextList.add(ContextList.java:37)
	at jdk.httpserver/sun.net.httpserver.ServerImpl.createContext(ServerImpl.java:306)
	at jdk.httpserver/sun.net.httpserver.HttpServerImpl.createContext(HttpServerImpl.java:69)
	at jdk.httpserver/sun.net.httpserver.HttpServerImpl.createContext(HttpServerImpl.java:34)
	at io.prometheus.metrics.exporter.httpserver.HTTPServer.registerHandler(HTTPServer.java:111)
```

This PR fixes the issue by skipping the default handler registration in case the metrics are configured for the root path.

I also improved the unit test so we can properly assert `expectedStatusCode` and `expectedBody`. With this improvement we can be sure that an endpoint actually returns the expected handler content, not just a matching status code.

### DIFF
--- a/prometheus-metrics-exporter-httpserver/src/test/java/io/prometheus/metrics/exporter/httpserver/HTTPServerTest.java
+++ b/prometheus-metrics-exporter-httpserver/src/test/java/io/prometheus/metrics/exporter/httpserver/HTTPServerTest.java
@@ -10,26 +10,42 @@ import com.sun.net.httpserver.HttpPrincipal;
 import com.sun.net.httpserver.HttpsConfigurator;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
 import io.prometheus.metrics.model.registry.PrometheusScrapeRequest;
+import io.prometheus.metrics.model.snapshots.CounterSnapshot;
+import io.prometheus.metrics.model.snapshots.CounterSnapshot.CounterDataPointSnapshot;
+import io.prometheus.metrics.model.snapshots.Labels;
+import io.prometheus.metrics.model.snapshots.MetricMetadata;
 import io.prometheus.metrics.model.snapshots.MetricSnapshots;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.Socket;
-import java.nio.charset.StandardCharsets;
-import java.security.NoSuchAlgorithmException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
 import java.security.Principal;
+import java.util.List;
 import java.util.concurrent.Executors;
 import javax.net.ssl.SSLContext;
 import javax.security.auth.Subject;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class HTTPServerTest {
 
-  @Test
-  @SuppressWarnings({"removal"})
-  public void testSubjectDoAs() throws Exception {
+  private PrometheusRegistry registry;
 
+  @BeforeEach
+  void setUp() {
+    final MetricMetadata metadata = new MetricMetadata("my-counter");
+    final CounterDataPointSnapshot dataPointSnapshot =
+        new CounterDataPointSnapshot(1.0, Labels.EMPTY, null, System.currentTimeMillis());
+
+    registry = new PrometheusRegistry();
+    registry.register(() -> new CounterSnapshot(metadata, List.of(dataPointSnapshot)));
+  }
+
+  @Test
+  public void testSubjectDoAs() throws Exception {
     final String user = "joe";
     final Subject subject = new Subject();
     subject.getPrincipals().add(() -> user);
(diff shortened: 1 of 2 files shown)


Donnerbart:
What exactly do you think breaks the backwards compatibility? With Java 11 it should already behave exactly like this (replacing the previously registered default handler). An with a newer Java runtime the current implementation doesn't work at all.

prometheus-metrics-exporter-httpserver/src/test/java/io/prometheus/metrics/exporter/httpserver/HTTPServerTest.java
```diff
@@ -10,25 +10,41 @@
...
+import java.util.List;
 import java.util.concurrent.Executors;
 import javax.net.ssl.SSLContext;
 import javax.security.auth.Subject;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class HTTPServerTest {
 
+  private PrometheusRegistry registry;
```
jaydeluca, line 35: so it looks like [this](https://github.com/openjdk/jdk/commit/f485171ce8c7e9c9d7d2c24e1807efaa6ff137e8) is the related change you are fixing? 

When I comment out your change in `HttpServer.java` and run these tests, they do not fail. Could you add a test that validates your fix?

Donnerbart: The new test `metricsCustomRootPath()` fails without the fix:
```java
java.lang.IllegalArgumentException: cannot add context to list

	at jdk.httpserver/sun.net.httpserver.ContextList.add(ContextList.java:37)
	at jdk.httpserver/sun.net.httpserver.ServerImpl.createContext(ServerImpl.java:306)
	at jdk.httpserver/sun.net.httpserver.HttpServerImpl.createContext(HttpServerImpl.java:69)
	at jdk.httpserver/sun.net.httpserver.HttpServerImpl.createContext(HttpServerImpl.java:34)
	at io.prometheus.metrics.exporter.httpserver.HTTPServer.registerHandler(HTTPServer.java:111)
	at io.prometheus.metrics.exporter.httpserver.HTTPServer.<init>(HTTPServer.java:74)
	at io.prometheus.metrics.exporter.httpserver.HTTPServer.<init>(HTTPServer.java:38)
	at io.prometheus.metrics.exporter.httpserver.HTTPServer$Builder.buildAndStart(HTTPServer.java:300)
```
Did you run the test with Java 21? The old code still works with Java 11, it's a behavior change in the JDK that breaks the test case.

jaydeluca: thats what I did, i commented out your code and then i ran the tests with java 25 and it didn't fail for me, but maybe there was caching in place, i'll have to try again

Donnerbart: We already fixed this in our [HiveMQ Prometheus Extension](https://github.com/hivemq/hivemq-prometheus-extension/pull/152), because we ran into the same issue. Due to a breaking change in the metrics naming in the current Prometheus Stack that project is still on the old 0.16.0 simpleclient stack.

We have a custom variant of the current `HTTPServer` class, so it had the same potential duplicate handler registration. But the project was still on Java 11, so nothing failed.

When we deployed the extension in our MQTT broker running with Java 21 and `/` as metric path, it started to fail (can't show the build jobs, it's on a closed source repo). So we have seen this failure on CI runs and could also reproduce it locally.

zeitlinger:
> What exactly do you think breaks the backwards compatibility? 

That registration would not work the first time

What about this change?

```diff
Index: prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java b/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java
--- a/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java	(revision 7f93e015dc64160204b51d13aebf75c83044a364)
+++ b/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HTTPServer.java	(date 1768917816428)
@@ -9,6 +9,9 @@
 import com.sun.net.httpserver.HttpsServer;
 import io.prometheus.metrics.config.PrometheusProperties;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
+
+import javax.annotation.Nullable;
+import javax.security.auth.Subject;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
@@ -21,8 +24,6 @@
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import javax.annotation.Nullable;
-import javax.security.auth.Subject;
 
 /**
  * Expose Prometheus metrics using a plain Java HttpServer.
@@ -66,11 +67,21 @@
     this.server = httpServer;
     this.executorService = executorService;
     String metricsPath = getMetricsPath(metricsHandlerPath);
+    try {
+      server.removeContext("/");
+    } catch (IllegalArgumentException e) {
+      // context "/" not registered yet, ignore
+    }
     registerHandler(
         "/",
         defaultHandler == null ? new DefaultHandler(metricsPath) : defaultHandler,
         authenticator,
         authenticatedSubjectAttributeName);
+    try {
+      server.removeContext(metricsPath);
+    } catch (IllegalArgumentException e) {
+      // context metricsPath not registered yet, ignore
+    }
     registerHandler(
         metricsPath,
         new MetricsHandler(config, registry),

```

Donnerbart:
To my knowledge contexts/handlers are registered per HttpServer instance, not globally. So how could `/` be registered already? Same with the `metricsPath`, unless the metrics path is `/` (then we would have already registered the default handler on this path).

At least I'm not able to come up with a test that already has a context registered on `/`, so the first `removeContext()` call actually does something. And the double registration of the metrics handler on `/` is fixed with less complexity by my `if equals` check.

I think both approaches will effectively do the same, so I pushed your suggestion.

zeitlinger:
thanks @Donnerbart - I'll merge once https://github.com/prometheus/client_java/pull/1781 is merged

PR #1781, opened 2026-01-15
Add test matrix for different java versions

jaydeluca:


.github/workflows/multi-version-test.yml
```diff
@@ -0,0 +1,45 @@
...
+permissions: {}
+
+jobs:
+  compatibility-test:
+    name: Test on Java ${{ matrix.java }}
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [17, 21, 25]
```
zeitlinger, line 15: why not 8, 11?

.mise/tasks/build-release.sh
```diff
@@ -9,4 +9,4 @@ set -euo pipefail
 VERSION=${usage_tag#v}
 
 mise run set-version "$VERSION"
-mvn -B package -P 'release,!default' -Dmaven.test.skip=true
+mvn -B package -P 'release,!default,!examples-and-integration-tests' -Dmaven.test.skip=true -Dgpg.skip=true
```
zeitlinger, line 12: why is this needed?

jaydeluca: adding the profiles has added some issues to the way some of these other tasks resolve everything.

from:
https://github.com/prometheus/client_java/actions/runs/21207964631/job/61008982563?pr=1781

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.14.1:testCompile (default-testCompile) on project prometheus-metrics-exposition-formats: Compilation failure: Compilation failure: 
Error:  /home/runner/work/client_java/client_java/prometheus-metrics-exposition-formats-shaded/target/metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/ProtobufExpositionFormatsTest.java:[11,45] cannot find symbol
Error:    symbol: class ExpositionFormatsTest
Error:  /home/runner/work/client_java/client_java/prometheus-metrics-exposition-formats-shaded/target/metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/ProtobufExpositionFormatsTest.java:[13,3] method does not override or implement a method from a supertype
```

the shaded module's tests need a test-jar dependency that's explicitly excluded from the release profile, so thats why i added the `!examples-and-integration-tests` to disable them (i assumed that would be ok to skip for testing the release build)

I was also encountering gpg issues both in CI and locally, not sure if this is a "me" thing, but didn't think it was necessary for this test step

.github/workflows/multi-version-test.yml
jaydeluca: i couldnt get them to work, can't remember why, let me try again to refresh my memory

jaydeluca: theres a bunch of test code that uses switch statements and """...""" text blocks. I can refactor them all if we think it's worth it?

jaydeluca: also, JUnit Jupiter 6.0.2 requires Java 17+ to run

jaydeluca: ive put a lot of effort into getting this to work for 8 and 11, and it's becoming a massive messy change. and adds a ton of complexity to the pom files. I propose that we keep the scope of this PR to what it is now, running the tests on 17+, and i can open an issue to track a followup where instead of running our entire test suite on every java version, we create some smoke tests that are more multi-version friendly in terms of dependencies etc, and then we can run that across 8 and 11 as well

.github/workflows/multi-version-test.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Check out
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+        with:
+          persist-credentials: false
+
+      - name: Set up Java ${{ matrix.java }}
+        id: setup-java
+        uses: actions/setup-java@f2beeb24e141e01a676f977032f5a29d81c9e27e # v5
```
jaydeluca:
```suggestion
        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
```

PR #1728, opened 2025-12-08
Allow duplicate metric names

jaydeluca:
Resolves #696

The goal of this change is to be able to have multiple metrics with the same name but different labels, which had previously been prevented by a check during registration.

While implementing this, I operated under a few assumptions:
- We will allow metrics with different labels to have the same name, but scraping metrics with the same name _and_ same labels results in an exception
- Metrics with the same name but different metric types will result in an exception
- Metrics with the same name but different labels should all be merged under a single HELP/TYPE heading

I originally attempted to implement the validation during metric registration, but that required doing some things that felt wrong, so I moved the validation to scrape time. This does result in a bit of a performance hit (~10% reduction in op/s in the benchmarks I ran) on the scrape operation.

The benchmarks I ran were on an apple M4. I did 4 runs total - 2 of the main branch for a baseline and 2 for this feature branch and then averaged them.

<details>

<summary>Benchmark results</summary>

## TextFormatUtilBenchmark

| Benchmark | Main #1 | Main #2 | Feature #1 | Feature #2 | Main Avg | Feature Avg |
|-----------|-------------|-------------|------------|------------|----------|---------|
| openMetricsWriteToByteArray | 1,179,213 ± 5,959 | 1,170,787 ± 5,579 | 1,080,037 ± 5,836 | 1,068,906 ± 5,288 | 1,175,000 | 1,074,471 |
| openMetricsWriteToNull | 1,197,768 ± 16,513 | 1,191,845 ± 13,897 | 1,083,832 ± 3,873 | 1,076,979 ± 4,816 | 1,194,806 | 1,080,406 |
| prometheusWriteToByteArray | 1,281,014 ± 21,437 | 1,273,552 ± 26,889 | 1,144,375 ± 3,303 | 1,138,252 ± 5,744 | 1,277,283 | 1,141,313 |
| prometheusWriteToNull | 1,254,177 ± 7,513 | 1,308,925 ± 14,326 | 1,167,007 ± 6,183 | 1,152,662 ± 2,565 | 1,281,551 | 1,159,835 |

---

## CounterBenchmark

| Benchmark | Main #1 | Main #2 | Feature #1 | Feature #2 | Main Avg | Feature Avg |
|-----------|-------------|-------------|------------|------------|----------|---------|
| codahaleIncNoLabels | 71,040 ± 1,676 | 81,355 ± 9,287 | 87,105 ± 8,193 | 78,829 ± 12,370 | 76,197 | 82,967 |
[...]

PR #1800, opened 2026-01-24
Allow metrics with the same name different labels

### LINKED ISSUE #696: Same metric name, different set of label names
We have a lot of Micrometer users requesting that we support registering metrics with the same name but a different set of labels (https://github.com/micrometer-metrics/micrometer/issues/877) for Prometheus. This repo, the official prometheus java client, does not allow this using the built-in types, as can be demonstrated by the following unit test:

```java
@Test
void promClientSameMetricNameDifferentLabels() {
    CollectorRegistry registry = new CollectorRegistry();
    Counter counter = Counter.build().name("api_response").labelNames("outcome").help(" ")
            .register(registry);
    Counter counter2 = Counter.build().name("api_response").labelNames("outcome", "error").help(" ")
            .register(registry); // fails with "Collector already registered that provides name: api_response_total"
}
```

It is interesting to note, however, that the Prometheus server will scrape the following without any seeming issue.

```
# HELP responses
# TYPE app_responses_total counter
app_responses_total{uri="/", outcome="SUCCESS"} 10
app_responses_total{uri="/", outcome="FAILURE", error="INVALID_REQUEST"} 2
app_responses_total{uri="/", outcome="FAILURE", error="DATA_NOT_FOUND"} 3
```

We have a custom Collector implementation for converting from Micrometer metrics to Prometheus metrics. We have received a pull request in Micrometer with changes that would allow Micrometer to produce a scape endpoint like above, circumventing the exception the Prometheus java client would otherwise throw when registering a metric with a different set of labels. See https://github.com/micrometer-metrics/micrometer/pull/2653.

I am opening an issue here because I would rather we did not do something in Micrometer that is intentionally guarded against in the official java client for Prometheus. I am not sure of all the implications supporting something like this might have on consumers. Is preventing different labels still an intentional choice here and are there plans for this library to support that? Would you recommend we avoid allowing it as well in Micrometer? Why does the Prometheus server scrape such output without error if it is considered an error in the client?
I appreciate any advice we can get on this from the Prometheus maintainers.

jaydeluca:
Closes #696

Alternate approach to #1728 

Validation occurs at registration time:

- Metrics with the same name, type, unit, and help - but different labels, are allowed to be registered
- Custom Collectors that do not implement the required fields (prometheus name, metric type) will skip validation and could risk creating duplicate series that will cause issues at scrape time.
  - I looked into adding a fallback, but it requires calling `collect()` at registration in order to introspect existing metadata and that seemed not great due to potential side effects


jaydeluca:
> What does that mean for clients that only work on snapshots like the OTel SDK?

Should be no change in behavior, I've added [some tests](https://github.com/prometheus/client_java/pull/1800/commits/1e897dc74c338804c765bcc65645c5faa27b2afa) that emulate the same type of interactions

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
```diff
@@ -1,96 +1,344 @@
...
+      MetricType metricType = collector.getMetricType();
+      Set<String> normalizedLabels = immutableLabelNames(collector.getLabelNames());
+      MetricMetadata metadata = collector.getMetadata();
+      String help = metadata != null ? metadata.getHelp() : null;
+      Unit unit = metadata != null ? metadata.getUnit() : null;
+
+      // Only perform validation if collector provides sufficient metadata.
+      // Collectors that don't implement getPrometheusName()/getMetricType() will skip validation.
+      if (prometheusName != null && metricType != null) {
+        final String name = prometheusName;
```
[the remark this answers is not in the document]
jaydeluca: adding additional complexity to solve for this seems unnecessary

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
```diff
@@ -1,96 +1,344 @@
...
+    List<MultiCollectorRegistration> registrations = new ArrayList<>();
+
+    try {
+      for (String prometheusName : prometheusNamesList) {
+        MetricType metricType = collector.getMetricType(prometheusName);
+        Set<String> normalizedLabels = immutableLabelNames(collector.getLabelNames(prometheusName));
+        MetricMetadata metadata = collector.getMetadata(prometheusName);
+        String help = metadata != null ? metadata.getHelp() : null;
+        Unit unit = metadata != null ? metadata.getUnit() : null;
+
```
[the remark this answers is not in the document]
jaydeluca: adding additional complexity to solve for this seems unnecessary

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
```diff
@@ -1,96 +1,326 @@
...
+      Unit unit = metadata != null ? metadata.getUnit() : null;
+
+      // Only perform validation if collector provides sufficient metadata.
+      // Collectors that don't implement getPrometheusName()/getMetricType() will skip validation.
+      if (prometheusName != null && metricType != null) {
+        validateRegistration(prometheusName, metricType, normalizedLabels, help, unit);
+        collectorMetadata.put(
+            collector, new CollectorRegistration(prometheusName, normalizedLabels));
       }
+    } catch (Exception e) {
```
zeitlinger: can we catch a specific ex here?

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
```diff
@@ -1,96 +1,326 @@
...
+        Unit unit = metadata != null ? metadata.getUnit() : null;
+
+        if (metricType != null) {
+          validateRegistration(prometheusName, metricType, normalizedLabels, help, unit);
+          registrations.add(new MultiCollectorRegistration(prometheusName, normalizedLabels));
+        }
       }
+
+      multiCollectorMetadata.put(collector, registrations);
+    } catch (Exception e) {
```
zeitlinger: can this be a specific ex?

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/Collector.java
```diff
@@ -78,4 +80,59 @@ default MetricSnapshot collect(
...
+   * same name. Returning {@code null} means type validation is skipped for this collector.
+   *
+   * <p>Validation is performed only at registration time. If this method returns {@code null}, no
+   * type validation is performed for this collector, and duplicate or conflicting metrics may
+   * result in invalid exposition output.
+   *
+   * @return the metric type, or {@code null} to skip validation
+   */
+  @Nullable
+  default MetricType getMetricType() {
```
zeitlinger, line 97: I like default methods over instanceof checks!

prometheus-metrics-exposition-formats-shaded/pom.xml
```diff
@@ -23,6 +23,13 @@
   </properties>
 
   <dependencies>
+    <!-- Ensures reactor builds exposition-formats first so generated protobuf sources exist before copy -->
```
zeitlinger: are these changes related to this PR?

prometheus-metrics-exposition-formats/generate-protobuf.sh
```diff
@@ -6,6 +6,24 @@ set -euo pipefail
 # I could not figure out how to use a protoc Maven plugin to use the shaded module,
 # so I ran this command to generate the sources manually.
 
+# Use gsed on macOS (requires: brew install gnu-sed) for in-place edits
```
zeitlinger: are these changes related to this PR?

zeitlinger (commented):
great - we're almost there 😄

prometheus-metrics-exposition-formats-shaded/pom.xml
jaydeluca: no, i updated them while fighting with build/generation stuff, but I think maybe I was running things out of order or something, i'll revert

prometheus-metrics-exposition-formats/generate-protobuf.sh
jaydeluca: it was just preventing me from being able to run `mise run generate` on my mac, but I can pull this out into a separate PR

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
zeitlinger: so now you have `RuntimeException` - can you list the specific exception that is thrown - or is this unknown?

jaydeluca: we invoke of a handful of methods in these blocks that could potentially have buggy implementations from custom collector implementations, which could throw NPEs or other runtime exceptions. And we need to catch them so we don't end up with partial registrations.

zeitlinger: OK, can you add that as a comment?

PR #1803, opened 2026-01-25
Run nightly benchmarks

jaydeluca:
Tested on my fork, results can be seen here: https://github.com/jaydeluca/client_java/tree/benchmarks

### DIFF
--- a/.github/workflows/nightly-benchmarks.yml
+++ b/.github/workflows/nightly-benchmarks.yml
@@ -47,28 +47,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
-      - name: Build benchmarks module
-        run: ./mvnw -pl benchmarks -am -DskipTests clean package
-
       - name: Run JMH benchmarks
-        run: |
-          # 3 forks, 3 warmup, 5 measurement iterations (~60 min total)
-          DEFAULT_ARGS="-f 3 -wi 3 -i 5"
-          JMH_ARGS="${INPUT_JMH_ARGS:-$DEFAULT_ARGS}"
-
-          echo "Running benchmarks with args: $JMH_ARGS"
-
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-benchmarks.yml
```diff
@@ -0,0 +1,116 @@
...
+          java -jar ./benchmarks/target/benchmarks.jar \
+            -rf json \
+            -rff benchmark-results.json \
+            $JMH_ARGS
+        env:
+          INPUT_JMH_ARGS: ${{ github.event.inputs.jmh_args }}
+
+      - name: Generate benchmark summary
+        run: |
+          python3 .mise/tasks/generate_benchmark_summary.py \
```
zeitlinger: I think this should also work as `mise run generate_benchmark_summary`

.github/workflows/nightly-benchmarks.yml
```diff
@@ -0,0 +1,116 @@
...
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Build benchmarks module
+        run: ./mvnw -pl benchmarks -am -DskipTests clean package
+
+      - name: Run JMH benchmarks
+        run: |
+          # 3 forks, 3 warmup, 5 measurement iterations (~60 min total)
```
zeitlinger: I see you added mise commands - maybe also make this one

.github/workflows/nightly-benchmarks.yml
jaydeluca: updated to use a mise task

.github/workflows/nightly-benchmarks.yml
jaydeluca: done

PR #1817, opened 2026-01-26
Make super-linter ARM friendly

jaydeluca:


.mise/tasks/lint/super-linter.sh
```diff
@@ -4,14 +4,20 @@
 
 set -euo pipefail
 
-# renovate: datasource=docker depName=ghcr.io/super-linter/super-linter
-SUPER_LINTER_VERSION="v8.3.2@sha256:e9d1895a1bdc1f9d9df41f688b27aa891743f23f9fae0f22a3e25eeda8f102db"
+# Super-linter doesn't publish ARM64 images, so force amd64 on Apple Silicon
+PLATFORM_FLAG=""
+if [[ "$(uname -m)" == "arm64" ]]; then
+	PLATFORM_FLAG="--platform linux/amd64"
```
zeitlinger: does it hurt to always add the flag?

jaydeluca: probably not, updated

PR #1738, opened 2025-12-14
Add GC duration histogram

gniadeck:
introduces a Prometheus histogram metric `jvm_gc_duration_seconds` to record JVM garbage collection pause durations. existing metrics provide limited visibility into long GC pauses, making it difficult to detect latency spikes. this change leverages the already registered GC notifications (as used in `JvmMemoryPoolAllocationMetrics`) to capture pause durations without additional instrumentation.

the histogram uses `0.01, 0.1, 1, 10` buckets and includes labels for `gc name`, `action`, and `cause`, enabling detailed monitoring of both short and long GC pauses. this addresses the lack of visibility highlighted in community discussions such as [this one](https://github.com/open-telemetry/semantic-conventions/issues/274). buckets are also defined according to the [opentelemetry semantic conventions spec](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#metric-jvmgcduration)

Example result:
```
# HELP jvm_gc_duration_seconds JVM GC pause duration histogram.
# TYPE jvm_gc_duration_seconds histogram
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="0.01"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="0.1"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="1.0"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="10.0"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="+Inf"} 806
jvm_gc_duration_seconds_count{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation"} 806
jvm_gc_duration_seconds_sum{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation"} 0.7360000000000005
```

PR #1828, opened 2026-01-28
support environment variables in properties

zeitlinger:
## Summary                                                                                                                                                                                                             
                                                                                                                                                                                      
 No  Breaking Changes. This is fully backward compatible:                                                                                                                                                                                              
   - ✅ Existing camelCase properties in files still work                                                                                                                                                                                
   - ✅ Existing system properties still work                                                                                                                                                                                            
   - ✅ Property precedence maintains existing behavior (external > system > file)                                                                                                                                                       
   - ✅ New environment variable layer added between external and system properties                                                                                                                                                      
                                                                                                                                                                                                                                         
   ## Migration Guide                                                                                                                                                                                                                    
                                                                                                                                                                                                                                         
   No migration required. However, to adopt the new features:                                                                                                                                                                            
                                                                                                                                                                                                                                         
   1. **To use environment variables:** Convert property names to uppercase with underscores                                                                                                                                             
      - `io.prometheus.exporter.httpServer.port` → `IO_PROMETHEUS_EXPORTER_HTTP_SERVER_PORT`                                                                                                                                             
                                                                                                                                                                                                                                         
   2. **To use snake_case (recommended):** Update properties files                                                                                                                                                                       
      - `io.prometheus.metrics.exemplarsEnabled` → `io.prometheus.metrics.exemplars_enabled`                                                                                                                                             
      - Old camelCase names continue to work                                 

## Notes 

- [ ] adjust new property added in  https://github.com/prometheus/client_java/pull/1738

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 max_line_length = 100
 indent_size = 2
 
-[{version-rules.xml,maven-wrapper.properties,checkstyle.xml,docker-compose.yaml,docker-compose.yml,Dockerfile,example_target_info.json,mise.toml,mvnm,mvnw.cmd,generate-protobuf.sh,.gitleaksignore}]
+[{version-rules.xml,maven-wrapper.properties,checkstyle.xml,docker-compose.yaml,docker-compose.yml,Dockerfile,example_target_info.json,mise.toml,mvnm,mvnw.cmd,generate-protobuf.sh,.gitleaksignore,prometheus.properties}]
 max_line_length = 200
 
 [{grafana-dashboard-*.json,.editorconfig,super-linter.env,renovate.json5}]


prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/PrometheusPropertiesLoader.java
```diff
@@ -84,26 +103,43 @@ private static Map<String, MetricsProperties> loadMetricsConfigs(Map<Object, Obj
   // If there are properties left starting with io.prometheus it's likely a typo,
   // because we didn't use that property.
   // Throw a config error to let the user know that this property doesn't exist.
-  private static void validateAllPropertiesProcessed(Map<Object, Object> properties) {
-    for (Object key : properties.keySet()) {
-      if (key.toString().startsWith("io.prometheus")) {
+  private static void validateAllPropertiesProcessed(PropertySource propertySource) {
+    for (String key : propertySource.getRemainingKeys()) {
+      if (key.startsWith("io.prometheus") || key.startsWith("io_prometheus")) {
         throw new PrometheusPropertiesException(key + ": Unknown property");
```
jaydeluca, line 109: i know this isn't new behavior, but I found this to be surprising that we would throw for this condition as opposed to just logging a warning. For example, if we deprecate a property and someone forgets to remove it from their config, their service will crash?

zeitlinger: Fail early is an idiom in this library (@fstab probably has more background info). That means that we have to be very careful with removing - probably only in a major release.

PR #1851, opened 2026-02-05
add branch coverage, better javadoc lints

zeitlinger:


prometheus-metrics-config/src/test/java/io/prometheus/metrics/config/ExporterPushgatewayPropertiesTest.java
```diff
@@ -30,6 +30,50 @@ void load() {
...
+            "io.prometheus.exporter.pushgateway.escapingScheme: Illegal value. Expecting"
+                + " 'allow-utf-8', 'values', 'underscores', or 'dots'. Found: invalid");
+  }
+
+  @Test
+  void loadWithTimeouts() {
+    ExporterPushgatewayProperties properties =
+        load(
+            Map.of(
+                "io.prometheus.exporter.pushgateway.connectTimeoutSeconds", "5",
```
jaydeluca: i know these are backwards compatible, but just noting we probably want to update these after #1828 is merged

zeitlinger: good catch! We can wait with this PR - it's not urgent.

zeitlinger: fixed now

PR #1854, opened 2026-02-06
Add GC duration histogram

zeitlinger:
Supercedes https://github.com/prometheus/client_java/pull/1738

introduces a Prometheus histogram metric `jvm_gc_duration_seconds` to record JVM garbage collection pause durations. existing metrics provide limited visibility into long GC pauses, making it difficult to detect latency spikes. this change leverages the already registered GC notifications (as used in `JvmMemoryPoolAllocationMetrics`) to capture pause durations without additional instrumentation.

the histogram uses `0.01, 0.1, 1, 10` buckets and includes labels for `gc name`, `action`, and `cause`, enabling detailed monitoring of both short and long GC pauses. this addresses the lack of visibility highlighted in community discussions such as [this one](https://github.com/open-telemetry/semantic-conventions/issues/274). buckets are also defined according to the [opentelemetry semantic conventions spec](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#metric-jvmgcduration)

Example result:
```
# HELP jvm_gc_duration_seconds JVM GC pause duration histogram.
# TYPE jvm_gc_duration_seconds histogram
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="0.01"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="0.1"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="1.0"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="10.0"} 806
jvm_gc_duration_seconds_bucket{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation",le="+Inf"} 806
jvm_gc_duration_seconds_count{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation"} 806
jvm_gc_duration_seconds_sum{action="end of minor GC",cause="G1 Evacuation Pause",gc="G1 Young Generation"} 0.7360000000000005
```

### DIFF
--- a/docs/content/instrumentation/jvm.md
+++ b/docs/content/instrumentation/jvm.md
@@ -130,6 +130,8 @@ OTel metrics). This replaces the standard metric with a histogram
 implemented according to the
 [OpenTelemetry Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#metric-jvmgcduration). <!-- editorconfig-checker-disable-line -->
 
+<!-- editorconfig-checker-disable -->
+
 ```text
 # HELP jvm_gc_duration_seconds Duration of JVM garbage collection actions.
 # TYPE jvm_gc_duration_seconds histogram
@@ -141,6 +143,8 @@ jvm_gc_duration_seconds_bucket{jvm_gc_action="end of minor GC",jvm_gc_name="G1 Y
 jvm_gc_duration_seconds_count{jvm_gc_action="end of minor GC",jvm_gc_name="G1 Young Generation"} 4
 jvm_gc_duration_seconds_sum{jvm_gc_action="end of minor GC",jvm_gc_name="G1 Young Generation"} 0.029
 ```
+
+<!-- editorconfig-checker-enable -->
 
 ## JVM Memory Metrics
 


prometheus-metrics-instrumentation-jvm/src/main/java/io/prometheus/metrics/instrumentation/jvm/JvmGarbageCollectorMetrics.java
```diff
@@ -40,6 +44,7 @@
 public class JvmGarbageCollectorMetrics {
 
   private static final String JVM_GC_COLLECTION_SECONDS = "jvm_gc_collection_seconds";
+  private static final String JVM_GC_DURATION = "jvm.gc.duration";
```
jaydeluca, line 47: should we be consistent with the way the other metric is defined?

```suggestion
  private static final String JVM_GC_DURATION = "jvm_gc_duration";
```

prometheus-metrics-instrumentation-jvm/src/main/java/io/prometheus/metrics/instrumentation/jvm/JvmGarbageCollectorMetrics.java
```diff
@@ -75,6 +87,54 @@ private void register(PrometheusRegistry registry) {
...
+                if (!GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION.equals(
+                    notification.getType())) {
+                  return;
+                }
+
+                GarbageCollectionNotificationInfo info =
+                    GarbageCollectionNotificationInfo.from(
+                        (CompositeData) notification.getUserData());
+
+                observe(gcDurationHistogram, info);
```
jaydeluca, line 124: try/catch here to avoid crashing the JVM is the notification listener throws something

```suggestion
                try {
                  if (!GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION.equals(
                    notification.getType())) {
                    return;
                  }

                  GarbageCollectionNotificationInfo info =
                    GarbageCollectionNotificationInfo.from(
                      (CompositeData) notification.getUserData());

                  observe(gcDurationHistogram, info);
                } catch (Exception e) {
                  logger.warning(
                      "Exception while processing garbage collection notification: " + e.getMessage());
                }
```

prometheus-metrics-instrumentation-jvm/src/main/java/io/prometheus/metrics/instrumentation/jvm/JvmGarbageCollectorMetrics.java
```diff
@@ -75,6 +87,54 @@ private void register(PrometheusRegistry registry) {
...
 
+  private void registerOtel(PrometheusRegistry registry) {
+    double[] buckets = {0.01, 0.1, 1, 10};
+
+    Histogram gcDurationHistogram =
+        Histogram.builder(config)
+            .name(JVM_GC_DURATION)
+            .unit(Unit.SECONDS)
+            .help("Duration of JVM garbage collection actions.")
+            .labelNames("jvm.gc.action", "jvm.gc.name", "jvm.gc.cause")
```
jaydeluca, line 98: `jvm.gc.cause` is still experimental, should we hold off on that one until it stabilizes?

prometheus-metrics-instrumentation-jvm/src/main/java/io/prometheus/metrics/instrumentation/jvm/JvmGarbageCollectorMetrics.java
```diff
@@ -75,6 +87,54 @@ private void register(PrometheusRegistry registry) {
...
+            .classicUpperBounds(buckets)
+            .register(registry);
+
+    registerNotificationListener(gcDurationHistogram);
+  }
+
+  private void registerNotificationListener(Histogram gcDurationHistogram) {
```
jaydeluca, line 105: taking a look at the OTel implementation, they also [track the listeners](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/runtime-telemetry/runtime-telemetry-java8/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/java8/internal/GarbageCollector.java#L101) and do a cleanup, not sure if it makes sense for us to implement something like that, by making `JvmMetrics` an AutoClosable. Looks like it ends up [requiring a lot of changes](https://github.com/prometheus/client_java/commit/ef939eb9b9fe52773804fdf9113fb2bfe84a761c), not sure if its worth it


Another thing they do is check to see if the [class exists first ](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/runtime-telemetry/runtime-telemetry-java8/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/java8/internal/GarbageCollector.java#L153-L163), which I think would be good to do here too

prometheus-metrics-instrumentation-jvm/src/test/java/io/prometheus/metrics/instrumentation/jvm/JvmGarbageCollectorMetricsTest.java
```diff
@@ -70,4 +84,170 @@ void testIgnoredMetricNotScraped() {
     verify(mockGcBean1, times(0)).getCollectionCount();
     assertThat(snapshots.size()).isZero();
   }
+
+  @Test
+  public void testNonOtelMetricsAbsentWhenUseOtelEnabled() {
```
jaydeluca, line 89:
```suggestion
  void testNonOtelMetricsAbsentWhenUseOtelEnabled() {
```

prometheus-metrics-instrumentation-jvm/src/test/java/io/prometheus/metrics/instrumentation/jvm/JvmGarbageCollectorMetricsTest.java
```diff
@@ -70,4 +84,170 @@ void testIgnoredMetricNotScraped() {
...
+        .register(registry);
+    registry.scrape();
+
+    verify(mockGcBean1, times(0)).getCollectionTime();
+    verify(mockGcBean1, times(0)).getCollectionCount();
+  }
+
+  @Test
+  @SuppressWarnings("rawtypes")
+  public void testGCDurationHistogramLabels() throws Exception {
```
jaydeluca, line 107:
```suggestion
  void testGCDurationHistogramLabels() throws Exception {
```

zeitlinger:
Thinking about this more I created https://github.com/prometheus/client_java/pull/1861 as an alternative that doesn't duplicate the otel instrumentation work.

Would that work for you @gniadeck ?

gniadeck:
@zeitlinger for my use case this would work fine, although i have some doubts about this implementation as a goal solution - in my PR implementing this metric was just a few lines of code, and now i’d need to import (and in the future update) `opentelemetry-sdk, opentelemetry-exporter-prometheus, opentelemetry-runtime-telemetry-java-8` (according to the demo), just to get better GC visibility. i haven’t tested the performance, so i’m not sure what the footprint would be, but it could be a no-go for some low-latency projects that choose prometheus for its speed.

i like this approach from an architectural standpoint tho, nice to see that client_java is so easily extensible with otel work :)

PR #1855, opened 2026-02-06
Make generate protobuf script compatible with mac

jaydeluca:


mise.toml
```diff
@@ -30,7 +30,6 @@ run = "./mvnw install -DskipTests -Dspotless.check.skip=true -Dcoverage.skip=tru
 [tasks.generate]
 description = "bare compile, ignoring formatting and linters"
 run = [
-  "mise use --pin protoc@latest",
```
jaydeluca, line 33: this caused me a lot of confusion because my local mise wasn't up to date, so it kept downgrading the version and then generating code that broke the build. I think the versions are updated by renovate so i dont think this is needed?

zeitlinger: I think I had the issue once that the releases were spread apart and renovate didn't group them. Happy to remove it if helps avoid confusion.

PR #1865, opened 2026-02-10
Add registry-level metric name filter to PrometheusRegistry

zeitlinger:
This is especially helpful when adding instrumentation libraries where you only need a few metrics.

See https://prometheus.github.io/client_java/getting-started/registry/#filtering-metrics (once merged)


zeitlinger:
This functionality already exists in https://prometheus.github.io/client_java/config/config/#exporter-filter-properties

PR #1868, opened 2026-02-12
Add OTel support module and make examples standalone

zeitlinger:
Inspired by https://github.com/prometheus/client_java/pull/1854#issuecomment-3885982843

## Summary

- Add `prometheus-metrics-otel-support` POM module that bundles the OTel SDK and Prometheus exporter as a single dependency, with the OTel instrumentation BOM imported for automatic version management
- Add `prometheus-metrics-otel-support` to the BOM so users can manage it via `dependencyManagement`
- Make all example projects standalone (no parent POM) — each example has its own `groupId`, `version`, and imports `prometheus-metrics-bom` directly, so users can copy-paste an example and build it without the parent chain
- Add docs page for the OTel Support module (`docs/content/otel/support.md`)
- Replace hardcoded `1.5.0` versions in OTel docs with `$version` placeholder and update `set-release-version-github-pages.sh` to process all markdown files under `docs/content`

## Test plan

- [ ] `mise run build` passes
- [ ] `mise run lint:super-linter` passes
- [ ] `mise run test` passes
- [ ] Verify `set-release-version-github-pages.sh` replaces `$version` in both `quickstart.md` and OTel docs

### DIFF
--- a/.mise/tasks/set-release-version-github-pages.sh
+++ b/.mise/tasks/set-release-version-github-pages.sh
@@ -5,5 +5,8 @@
 set -euox pipefail
 
 version=$(git tag -l | grep 'v' | sort | tail -1 | sed 's/v//')
-marker="\$version"
-find ./docs/content -name '*.md' -exec sed -i "s/$marker/$version/g" {} +
+otelVersion=$(grep -oP '<otel.instrumentation.version>\K[^<]+' pom.xml | sed 's/-alpha$//')
+
+find ./docs/content -name '*.md' \
+	-exec sed -i "s/\$version/$version/g" {} + \
+	-exec sed -i "s/\$otelVersion/$otelVersion/g" {} +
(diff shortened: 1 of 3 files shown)


examples/example-otel-jvm-runtime-metrics/pom.xml
```diff
@@ -19,10 +20,17 @@
...
+        <artifactId>prometheus-metrics-bom</artifactId>
+        <version>1.5.0-SNAPSHOT</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
       <dependency>
         <groupId>io.opentelemetry.instrumentation</groupId>
         <artifactId>opentelemetry-instrumentation-bom-alpha</artifactId>
-        <version>${otel.instrumentation.version}</version>
+        <version>2.24.0-alpha</version>
```
jaydeluca, line 33: are we intentionally hard coding this? is it because we no longer have access to the variable because we removed the parent? Will this get updated automatically?

zeitlinger: - yes, relying on parent makes it impossible to copy-paste
- yes, will get updated (this will be obvious when https://github.com/grafana/docker-otel-lgtm/blob/main/.github/renovate-tracked-deps.json is used in this repo)

PR #1904, opened 2026-02-20
ci: add mise version+sha256 pinning to CI

zeitlinger:
## Summary
- Pin mise `version` and `sha256` in all workflow files for reproducibility and supply-chain security
- Renovate will keep these up to date via the custom manager in [grafana/flint#43](https://github.com/grafana/flint/pull/43)

## Test plan
- [ ] CI passes with pinned mise version

### DIFF
--- a/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/internal/ProtobufUtil.java
+++ b/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/internal/ProtobufUtil.java
@@ -14,9 +14,6 @@ public class ProtobufUtil {
   }
 
   public static String shortDebugString(MessageOrBuilder protobufData) {
-    return TextFormat.printer()
-        .emittingSingleLine(true)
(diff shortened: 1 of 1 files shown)


.github/renovate-tracked-deps.json
```diff
@@ -2,6 +2,33 @@
...
+  },
+  ".github/workflows/nightly-benchmarks.yml": {
+    "regex": ["mise"]
+  },
+  ".github/workflows/release.yml": {
+    "regex": ["mise"]
+  },
+  ".github/workflows/test-release-build.yml": {
+    "regex": ["mise"]
+  },
```
jaydeluca, line 31: could we just do something like this?

```
  ".github/workflows/*": {
    "regex": ["mise"]
  },
```

zeitlinger: Thanks for the suggestion! Unfortunately, this linter doesn't support wildcard patterns like that. If you'd like to see this feature added, please create an issue in the flint repository to discuss the enhancement.

PR #1920, opened 2026-02-27
ci: automate protobuf source generation for Renovate updates

zeitlinger:
## Please don't merge

I use this PR to test a new git helper tool

## Summary
- Group `protobuf-java` and `protoc` into a single Renovate PR (`separateMajorMinor: false`)
- Add workflow that triggers on `renovate/protobuf` branch to regenerate protobuf sources
- Remove `mise use --pin protoc@latest` from generate task (no longer needed with grouping)

## Known limitation
The workflow pushes with `GITHUB_TOKEN`, which [doesn't trigger CI re-runs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). After the generated commit is pushed, **close and reopen the PR** to trigger CI.

Follow-up: switch to `PROMBOT_GITHUB_TOKEN` so CI re-runs automatically.

## Test plan
- [ ] Close PRs #1918 and #1919, let Renovate re-create as a grouped PR
- [ ] Verify the workflow triggers and regenerates protobuf sources
- [ ] Verify the generated commit appears on the PR
- [ ] Close and reopen the PR, verify CI runs on the generated commit

### DIFF
--- a/.github/workflows/generate-protobuf.yml
+++ b/.github/workflows/generate-protobuf.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Commit and push generated sources
         run: |
           git diff --quiet && exit 0
-          UNEXPECTED=$(git diff --name-only | grep -v '\.java$' | grep -v '^mise\.toml$' || true)
+          UNEXPECTED=$(git diff --name-only | grep -v '\.java$' || true)
           if [[ -n "$UNEXPECTED" ]]; then
             echo "::error::Unexpected files changed:"
             echo "$UNEXPECTED"
@@ -59,6 +59,6 @@ jobs:
           # TODO: switch to PROMBOT_GITHUB_TOKEN once it's added to this repo.
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git add '*.java' mise.toml
+          git add '*.java'
           git commit -m "chore: regenerate protobuf sources"
           git push


.github/workflows/generate-protobuf.yml
```diff
@@ -0,0 +1,64 @@
...
+          if ! echo "$DIFF_MISE" | grep -q 'protoc'; then
+            echo "::error::protoc not updated in mise.toml"
+            exit 1
+          fi
+      - name: Generate protobuf sources
+        run: mise run generate
+      - name: Commit and push generated sources
+        run: |
+          git diff --quiet && exit 0
+          UNEXPECTED=$(git diff --name-only | grep -v '\.java$' | grep -v '^mise\.toml$' || true)
```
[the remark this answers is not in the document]
zeitlinger: Accepted — removed the `mise.toml` exception from the unexpected files check.

.github/workflows/generate-protobuf.yml
```diff
@@ -0,0 +1,64 @@
...
+            echo "::error::Unexpected files changed:"
+            echo "$UNEXPECTED"
+            exit 1
+          fi
+          # Note: GITHUB_TOKEN pushes don't trigger CI re-runs.
+          # Close and reopen the PR to trigger CI after this commit.
+          # TODO: switch to PROMBOT_GITHUB_TOKEN once it's added to this repo.
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add '*.java' mise.toml
```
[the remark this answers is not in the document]
zeitlinger: Accepted — now only adding `*.java` to the commit.

.github/workflows/generate-protobuf.yml
```diff
@@ -0,0 +1,64 @@
+---
+name: Generate Protobuf
+
+on:
+  push:
+    branches:
+      - "renovate/protobuf"
```
martincostello, line 7: Note that a push for this from renovate could still theoretically contain untrusted code if the dependency being updated was compromised and a malicious version published.

If the update contained malicious code that tried to steal `GITHUB_TOKEN` (or whatever secret is used), it would able to perform arbitrary write operations on the repository, or whatever permissions that token had.

It's risky performing automated writes to the repository with elevated permissions in this context.

It's not something I've tried to do myself, but a safer approach could be to have this job only generate a Git patch for the changes as a build artifact, and then have another job/workflow take that artifact and apply it after applying some sanity checks (e.g. the diff only updates the expected files). Then any compromised code isn't able to actually execute with access to the secrets but the protobuf files can still be regenerated.

.github/workflows/generate-protobuf.yml
```diff
@@ -0,0 +1,64 @@
...
+          git diff --quiet && exit 0
+          UNEXPECTED=$(git diff --name-only | grep -v '\.java$' || true)
+          if [[ -n "$UNEXPECTED" ]]; then
+            echo "::error::Unexpected files changed:"
+            echo "$UNEXPECTED"
+            exit 1
+          fi
+          # Note: GITHUB_TOKEN pushes don't trigger CI re-runs.
+          # Close and reopen the PR to trigger CI after this commit.
+          # TODO: switch to PROMBOT_GITHUB_TOKEN once it's added to this repo.
```
martincostello, line 59: If `PROMBOT_GITHUB_TOKEN` has permissions greater than `contents: write` (e.g. if it had admin access) then that might be even riskier than using `GITHUB_TOKEN` regardless of whether it can trigger CI or not.

.github/workflows/generate-protobuf.yml
zeitlinger: The attack vector here is a compromised `protoc` binary or `protobuf-java` library. The mitigations are:
1. Only `.java` files can be committed — anything else fails the workflow
2. `protoc` is downloaded from Google's official GitHub releases via mise
3. The workflow only has `contents: write` — no admin, no secrets beyond `GITHUB_TOKEN`
4. The generated code still goes through normal CI and review before merge

This is the same trust model as running `protoc` locally — we're trusting Google's releases. If protoc itself were compromised, we'd have bigger problems regardless of automation.

.github/workflows/generate-protobuf.yml
zeitlinger: Good point — `PROMBOT_GITHUB_TOKEN` should be scoped to `contents: write` only, same as `GITHUB_TOKEN`. I'll note that in the TODO. For now the workflow uses `GITHUB_TOKEN` which can't have elevated permissions.

PR #1927, opened 2026-03-02
chore: Add automation for stale issues and PRs

jaydeluca:
Based on the automation used in the otel java projects, but with longer thresholds


.github/workflows/issue-management-stale-action.yml
```diff
@@ -0,0 +1,59 @@
+---
+name: Issue management - run stale action
+
+on:
+  schedule:
+    # hourly at minute 23
+    - cron: "23 * * * *"
```
jaydeluca:
```suggestion
    - cron: "23 6 * * *"
```

PR #1931, opened 2026-03-03
fix: exclude standalone examples from `mise run format`

zeitlinger:
## Summary
- `mise run format` (i.e. `spotless:apply`) fails on JDK 25+ because
  standalone example POMs don't inherit the spotless plugin from the
  project parent, but are pulled into the reactor via the
  `examples-and-integration-tests` profile
- Fix: use the fully-qualified plugin goal with profile deactivation
- Add `lint:example-poms` check to catch this class of issue

Relates https://github.com/prometheus/client_java/pull/1927#issuecomment-3984540872

## Test plan
- [x] `mise run format` succeeds
- [x] `lint:example-poms` passes with the fix
- [x] `lint:example-poms` fails when reverting the fix

### DIFF
--- a/.mise/tasks/lint/example-poms.py
+++ b/.mise/tasks/lint/example-poms.py
@@ -46,9 +46,7 @@ def format_task_excludes_examples() -> bool:
     mise_toml = ROOT / "mise.toml"
     text = mise_toml.read_text(encoding="utf-8")
     # Look for the format task run command
-    match = re.search(
-        r'\[tasks\.format\].*?run\s*=\s*"([^"]*)"', text, re.DOTALL
-    )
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -143,6 +142,12 @@ mise run lint
 # or to autofix: mise run fix
 ```
 
+### Before Pushing
+
+**ALWAYS** run `mise run lint` before pushing to verify
+all lints pass. CI runs the same checks and will fail
+if any lint is violated.
+
```
jaydeluca, line 150: 👍

jaydeluca (approved):
thanks for jumping on this!

Issue #1942, opened 2026-03-10
Move suffix handling from creation time to scrape time


zeitlinger:
Part of #1912. Prerequisite for OM2 suffix-related PRs.

## TL;DR

Today, metric names are modified at both creation time
(reserved suffix rejection) and scrape time (suffix
appending/stripping). This design moves all suffix handling
to scrape time. Going forward, the SDK does not modify
metric names — only the OM1 legacy format writer appends
suffixes.

Reference: [Metric Names: A Guide for Client Library
Maintainers](https://docs.google.com/document/d/1VbqxXmtZmU9MWvK936uf80ulaR7X9xU_PjrR05avKO8/edit?tab=t.0)

Dev summit consensus (2024-09-13):
> Suffixes like `_total` or `_info` are only ever enforced
> in OM v1 and some existing SDK implementations. They MUST
> NOT be enforced in any other situation.
>
> We intend to get rid of automatically enforced suffixes
> with OM 2.0.

## Principles

1. **Don't break existing apps.** Existing names exposed
   via OM1 (pull) and OTel (push) must not change without
   explicit opt-in.
2. **The SDK does not modify metric names.** The OM2 spec's
   SHOULDs about unit suffixes and `_total` are guidance
   for users naming their metrics, not for the SDK to
   enforce. OM2 and OTel preserve output the name exactly
   as provided.
3. **OM1 is legacy.** It continues to append `_total`, unit
   suffixes, etc. at scrape time. No changes to OM1 output.
4. **OTel legacy behavior behind a flag.** The existing OTel
   name stripping (`_total`, unit suffix) is kept as default
   for backward compat, flipped in v2 (#1943).
5. **Fail early on collisions.** Registry detects name
   collisions across all formats at registration time.

## Key table

| User provides | OM1 | OM2 | OTel | OTel + otel_preserve_names |
|---|---|---|---|---|
| `Counter("events")` | `events_total` | `events` | `events` | `events` |
| `Counter("events_total")` | `events_total` | `events_total` | `events` | `events_total` |
| `Counter("req_bytes").unit(BYTES)` | `req_bytes_total` | `req_bytes` | name `req`, unit `By` | name `req_bytes`, unit `By` |
| `Counter("req").unit(BYTES)` | `req_bytes_total` | `req` | name `req`, unit `By` | name `req`, unit `By` |
| `Gauge("connections")` | `connections` | `connections` | `connections` | `connections` |
[...]

Issue #1912, opened 2026-02-24
Experimental OpenMetrics 2.0 support


zeitlinger:
The [OpenMetrics 2.0 spec](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md) is nearing completion. We should start implementing experimental support to validate the design and provide early feedback to the spec authors.

All features are behind opt-in flags (off by default) so there's no risk to existing users. The spec is still in draft, so the API is explicitly experimental and subject to change.

## Status

Implemented:

- PR 1: Flag infrastructure (#1939) ✅
- PR 2: OM2 writer skeleton (#1951) ✅
- PR 3: enabled flag (#1953) ✅
- PR 4a: Core + OM1 writers (#1955) ✅
- PR 4b: OTel preserve_names (#1956) ✅
- PR 4c: OM2 writer no-suffix (#1957) ✅
- PR 5: content negotiation (#1986) ✅
- PR 6: compositeValues + exemplarCompliance (#1991, closes #1989) ✅
- PR 7: nativeHistograms (#2042, closes #1990) ✅

Remaining:

- PR 8: Documentation (#1954)

## Flag design

Top-level gate + individual feature flags. The top-level gate (`enableOpenMetrics2()`)
sets `enabled=true` implicitly and enables the sub-flags. Each sub-flag
independently controls one OM2 feature.

**Java builder API:**

```java
PrometheusProperties config = PrometheusProperties.builder()
    .enableOpenMetrics2(om2 -> om2
        .contentNegotiation(true)         // version=2.0.0 content type
        .compositeValues(true)            // single-line histogram/summary + st@
        .exemplarCompliance(true)         // mandatory timestamps, no size limit
        .nativeHistograms(true)           // exponential buckets
    )
    .build();

// Or enable everything at once
PrometheusProperties config = PrometheusProperties.builder()
    .enableOpenMetrics2(om2 -> om2.enableAll())
    .build();
```

**Properties equivalent:**

```properties
io.prometheus.openmetrics2.enabled=true
io.prometheus.openmetrics2.content_negotiation=true
io.prometheus.openmetrics2.composite_values=true
io.prometheus.openmetrics2.exemplar_compliance=true
io.prometheus.openmetrics2.native_histograms=true
```

For the minimal OM2 experience (names as provided, no suffix appending):

```properties
io.prometheus.openmetrics2.enabled=true
```

| Flag | Description | Spec reference |
|------|-------------|----------------|
[...]

PR #1956, opened 2026-03-17
feat: add OTel preserve_names for scrape-time suffix handling

zeitlinger:
## Summary

Adds `preserve_names` configuration to the OpenTelemetry exporter.
When enabled, metric names are passed through exactly as the user
wrote them instead of stripping `_total` and unit suffixes.
Part of #1942.

### Key changes

- Add `preserve_names` to `ExporterOpenTelemetryProperties`
- `MetricDataFactory` uses `originalName` + `preserve_names` to
  decide naming
- `OtelAutoConfig` wires the new property

### Key table

| User provides | OTel | OTel preserve_names |
|---|---|---|
| `Counter("events")` | `events` | `events` |
| `Counter("events_total")` | `events` | `events_total` |
| `Counter("req").unit(BYTES)` | name `req`, unit `By` | name `req`, unit `By` |
| `Counter("req_bytes").unit(BYTES)` | name `req`, unit `By` | name `req_bytes`, unit `By` |
| `Gauge("events_total")` | `events_total` | `events_total` |

### PR stack

1. Core model + OM1/protobuf writers (#1955)
2. **This PR** — OTel `preserve_names`
3. OM2 writer no-suffix (independent)

## Test plan

- [x] `mise run compile` passes
- [x] Tests for `preserve_names=true` with units, unit already in
  name, and without unit
- [x] `OtelAutoConfigTest` covers new property wiring

Part of #1912.


prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/InfoSnapshotTest.java
```diff
@@ -61,13 +61,13 @@ void testDataImmutable() {
 
   @Test
   void testNameMustNotIncludeSuffix() {
```
jaydeluca: i think we need to update these test names since this validation is no longer occurring and these are allowed now

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java
```diff
@@ -161,6 +207,23 @@ private void validateRegistration(
     final Set<String> names = normalizedLabels;
     final String helpForValidation = help;
     final Unit unitForValidation = unit;
+
+    // Check exposition name collisions before modifying any state.
+    Set<String> expositionNames = computeExpositionNames(prometheusName, type);
+    for (String expositionName : expositionNames) {
+      String owner = expositionNameOwners.get(expositionName);
```
jaydeluca: we are checking the expositionNameOwners map here, but then we aren't updating it until line 258, which could open us up to some race conditions

Instead we could handle it all at once:

```java
Set<String> expositionNames = computeExpositionNames(prometheusName, type);
for (String expositionName : expositionNames) {
  String existingOwner = expositionNameOwners.putIfAbsent(expositionName, prometheusName);
  if (existingOwner != null && !existingOwner.equals(prometheusName)) {
    // Rollback any names we already claimed
    for (String claimed : expositionNames) {
      if (claimed.equals(expositionName)) break;
      expositionNameOwners.remove(claimed, prometheusName);
    }
    throw new IllegalArgumentException(
        "'" + prometheusName + "' and '" + existingOwner +
        "' have conflicting exposition name: '" + expositionName + "'");
  }
}

registered.compute(
        prometheusName,
...
```

prometheus-metrics-exporter-opentelemetry/src/test/java/io/prometheus/metrics/exporter/opentelemetry/ExportTest.java
```diff
@@ -324,6 +325,69 @@ void metricsWithoutDataPointsAreNotExported() {
     assertThat(metrics).isEmpty();
   }
 
+  @Test
+  void preserveNamesWithUnit() throws NoSuchFieldException, IllegalAccessException {
+    PrometheusRegistry preserveRegistry = new PrometheusRegistry();
+    Field field = testing.getClass().getDeclaredField("metricReader");
+    field.setAccessible(true);
+    MetricReader reader = (MetricReader) field.get(testing);
```
jaydeluca: just an idea, but I think we could just use an InMemoryReader instead of needing to use reflection in all these tests in order to use the reader from the extension, since I don't think we're getting anything extra from that.

```suggestion
    InMemoryMetricReader reader = InMemoryMetricReader.create();
```

It would just change the way we are accessing it, instead of

```
 List<MetricData> metrics = testing.getMetrics();
```
it would be

```
List<MetricData> metrics = new ArrayList<>(reader.collectAllMetrics());
```

Issue #645, opened 2021-03-24
The MapperConfig


muuki88:
Hi,

Thanks for the custom mapper config on the dropwizard export. It makes creating tags really easy :heart: 

I found a restriction in the regex that makes it impossible for me to create a certain configuration. `:` are not allowed and we have metrics that contain `ip:port` in the metric name.

https://github.com/prometheus/client_java/blob/29dfd0a2bf9b93282764d31447b843a1d36d4f10/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/samplebuilder/MapperConfig.java#L25

This is what the mapper config would look like

```java
// this port is dynamically set from the outside
final int httpPort = 8080
final MapperConfig vertxServerHttp = new MapperConfig(
     // the IP address is always 0.0.0.0
    "vertx.http.servers.0.0.0.0:" + httpPort + ".*", "vertx.http.servers.${4}",
     ImmutableMap.of("endpoint", "http")
);
```

So the metric currentely ends up like this in prometheus

```
vertx_http_servers_0_0_0_0:8080_metric
```

which gets messy as ports change.

## Suggestion

Add `:` to the global regex.

I don't know what implications this would have

fstab:
Sorry for the delay. I put this on my TODO list, but I would like to get Exemplars done first before I look into Dropwizard. I will get back to this issue.

muuki88:
No worries :heart: I would open a pull request with this feature if it helps you and there's a chance this can get merged :smile:

Issue #518, opened 2019-12-18
Metric pattern validation in simpleclient_dropwizard is overly strict


LiamClarkeNZ:
I am exposing metrics in an existing app that uses Dropwizard to Prometheus, and they have a wide variety of names. From `app_metric_some_count` to `appMetric.someCount`. For legacy reasons I am unable to rename these metrics. 

When attempting to use a CustomMappingSampleBuilder to map these existing names to metric names that align with Prometheus best practices, I am unable to do so because both `MappingConfig` and `GraphiteNamePattern` (used within the `CustomMappingSampleBuilder`) expect via regex a very particular form of metric naming - `app_example` will not survive validation, but `app.example` will. 

Given that Dropwizard itself does not enforce these metric naming conventions, I'm unsure as to why `simpleclient_dropwizard` does.

brian-brazil:
Would you like to send a PR?

LiamClarkeNZ:
Hi @brian-brazil,

Yep, I'm intending to, I have a few days free coming up so will put something together then. 

Can I ask your opinion before I do, though? I was intending, in order to preserve backwards compatibility for anyone using the existing CustomMappingSampleBuilder, to create class(es) providing similar functionality alongside  it, instead of modifying the classes mentioned above directly. How does that sit with you?

Cheers, 

Liam Clarke

brian-brazil:
I'd modify directly, and I'm not even sure how this could break anyone. I'd suggest seeing what the statsd/graphite exporter do here to, as we should be in line with them.

LiamClarkeNZ:
Thanks Brian, look forward to your feedback on the PR.

LiamClarkeNZ:
PR: https://github.com/prometheus/client_java/pull/519

peetzen:
I was running into the exact same problem when writing a dropwizard wrapper bundle.
When trying to map _io.dropwizard.jetty.MutableServletContextHandler.*-requests_ to _servlet.requests_ with a label defined as _method: ${0}_ the validation does not let it pass through due to _*_ not followed by a period.

The whole _match_ pattern validation seem unnecessary, as what matters at the end is the actual metric name.

Issue #461, opened 2019-03-05
Match a Graphite metric with only one level and add a prefix


nkafr:
When using DropwizardExports, if we want to add a base prefix name to a metric we can do this:

```
MapperConfig depth3 = new MapperConfig();
depth3.setMatch("*.*.*");  //mathches  depth1.depth2.depth3
depth3.setName("basename" + "_${0}" + "_${1}" + "_${2}");
```
However, a metric with a single depth(no dot in its name) cannot be matched (because the matcher expects at least one dot), and thus in this case we cannot add a custom base prefix. Is there a way to achieve this functionality?

Thanks in advance,
Nick

PR #1985, opened 2026-03-25
fix: Relax metric name validation in Dropwizard5

LiamClarkeNZ:
Decided to revisit a very old PR of mine that addressed an issue that seems to reoccur periodically for others. 

## Changes
- Relaxes overly strict `METRIC_GLOB_REGEX` in `MapperConfig`/`GraphiteNamePattern` to accept a broader range of metric name patterns
- The `CustomMappingSampleBuilder` can now remap metrics with underscores, colons, hyphens, and single-level names
- Still rejects clearly invalid inputs (empty strings, double-star globs, trailing/double dots)

## Fixes
- #461: Cannot match a Graphite metric with only one level
- #518: Custom metric name remapping fails for non-Graphite-style names
- #645: MapperConfig does not allow `:` in regex

The validation was originally designed for Graphite bridge metrics but is too restrictive for `CustomMappingSampleBuilder` which needs to handle arbitrary Dropwizard metric names.

@fstab @dhoard @zeitlinger @jaydeluca when you have a chance would appreciate your review :)

### DIFF
--- a/prometheus-metrics-instrumentation-dropwizard5/src/main/java/io/prometheus/metrics/instrumentation/dropwizard5/labels/MapperConfig.java
+++ b/prometheus-metrics-instrumentation-dropwizard5/src/main/java/io/prometheus/metrics/instrumentation/dropwizard5/labels/MapperConfig.java
@@ -16,11 +16,14 @@ import javax.annotation.Nullable;
  * and new labels based on this config.
  */
 public final class MapperConfig {
-  // each part of the metric name between dots
-  private static final String METRIC_PART_REGEX = "[a-zA-Z_0-9](-?[a-zA-Z0-9_])+";
-  // Simplified GLOB: we can have "*." at the beginning and "*" only at the end
+  // Each part of the metric name between dots. Accepts letters, digits, underscores, hyphens,
+  // colons, and glob wildcards (*) to support a broad range of metric naming conventions.
+  private static final String METRIC_PART_REGEX = "[a-zA-Z_0-9*][a-zA-Z0-9_:\\-*]*";
+  // Simplified GLOB: accepts single-level names, dot-separated names, and glob patterns with '*'.
+  // The pattern requires at least one non-empty segment and does not allow empty segments (double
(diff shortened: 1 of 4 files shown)


prometheus-metrics-instrumentation-dropwizard5/src/main/java/io/prometheus/metrics/instrumentation/dropwizard5/labels/MapperConfig.java
```diff
@@ -16,11 +16,14 @@
  * and new labels based on this config.
  */
 public final class MapperConfig {
-  // each part of the metric name between dots
-  private static final String METRIC_PART_REGEX = "[a-zA-Z_0-9](-?[a-zA-Z0-9_])+";
-  // Simplified GLOB: we can have "*." at the beginning and "*" only at the end
+  // Each part of the metric name between dots. Accepts letters, digits, underscores, hyphens,
+  // colons, and glob wildcards (*) to support a broad range of metric naming conventions.
+  private static final String METRIC_PART_REGEX = "[a-zA-Z_0-9*][a-zA-Z0-9_:\\-*]*";
```
jaydeluca: moving the hyphen to the end removes the need to escape it
```suggestion
  private static final String METRIC_PART_REGEX = "[a-zA-Z_0-9*][a-zA-Z0-9_:*-]*";
```

jaydeluca (approved):
knocking out 3 issues in one PR, going for the hat-trick! 😄 

just one small nitpick comment but otherwise LGTM

prometheus-metrics-instrumentation-dropwizard5/src/main/java/io/prometheus/metrics/instrumentation/dropwizard5/labels/MapperConfig.java
LiamClarkeNZ: @jaydeluca Done, thank you :)

PR #1987, opened 2026-03-30
fix: handle multi-type Accept headers in parseOpenMetricsVersion

zeitlinger:
## Summary
- Fix bug where `parseOpenMetricsVersion` could pick up `version` parameters from unrelated media types in multi-type Accept headers (e.g. `application/openmetrics-text, text/plain; version=0.0.4`)
- Split Accept header by `,` first to isolate individual media types before parsing parameters
- Add two test cases covering multi-type Accept headers with and without an explicit openmetrics version

## Test plan
- [x] All 55 existing `ExpositionFormatsTest` tests pass
- [x] New test: multi-type header where only `text/plain` has version → OM1 writer selected
- [x] New test: multi-type header where openmetrics has `version=2.0.0` → OM2 writer selected


jaydeluca (approved):
you could've left a code review comment and I could have done this, but thanks!

PR #1991, opened 2026-04-09
feat: compositeValues and exemplarCompliance flags for OM2 writer

### LINKED ISSUE #1989: OM2 - CompositeValue + start timestamp + exemplar compliance
- `compositeValues`:
  - Histogram as single line: `foo {count:17,sum:324789.3,bucket:[0.1:8,0.25:10,0.5:11,1.0:14,+Inf:17]} 1.0 st@0.5`
  - Summary as single line: `bar {count:17,sum:324789.3,quantile:[0.95:123.7,0.99:150.0]}`
  - GaugeHistogram as single line with `count`/`sum` (not `gcount`/`gsum`)
  - `_created` lines replaced by inline [`st@` start timestamp](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md#start-timestamp)
  - `_count`, `_sum`, `_bucket` suffixed lines no longer emitted for these types
  - Reserved label prefix changed from `_` to `__`
  - Histogram/GaugeHistogram Count and Sum always present (MUST, was SHOULD for Sum)
  - Histogram bucket values may be float (SHOULD be integer, was MUST)
  - Spec: [CompositeValue](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md#compositevalue), [Histogram](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md#histogram-1), [Summary](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md#summary-1)
- `exemplarCompliance`:
  - Exemplar timestamps always emitted ([OM2: MUST](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md#exemplar), was MAY in OM1)
  - 128-char LabelSet hard limit removed

zeitlinger:
Fixes https://github.com/prometheus/client_java/issues/1989

## Summary

- **`compositeValues=true`**: Histogram, GaugeHistogram, and Summary are
  written as a single composite-value line per the OM2 spec:
  ```
  foo {count:17,sum:324789.3,bucket:[0.1:8,0.25:10,0.5:11,1.0:14,+Inf:17]} 1520879607.789 st@1520430000.123
  bar {count:17,sum:324789.3,quantile:[0.95:123.7,0.99:150.0]}
  ```
  - Bucket counts are cumulative (same semantics as OM1 `_bucket` lines)
  - GaugeHistogram uses `gcount`/`gsum` per spec (note: the tracking issue
    says "count/sum" but the spec is explicit about `gcount`/`gsum`)
  - Created timestamp moves inline as `st@`; no separate `_created` line
  - Latest exemplar (if present) is appended inline
  - No separate `_bucket`, `_count`, `_sum` lines emitted
  - When `compositeValues=false` (default), output is unchanged from before

- **`exemplarCompliance=true`**: Exemplars without a timestamp are silently
  dropped. The OM2 spec mandates that all exemplars MUST have a timestamp.
  The 128-char LabelSet limit from OM1 was never enforced in the writer, so
  no change needed there.

## Test plan

- [x] `testCompositeHistogram` — basic single-line histogram output
- [x] `testCompositeHistogramWithLabelsTimestampAndCreated` — labels, scrape
  timestamp, and `st@` start timestamp
- [x] `testCompositeHistogramWithExemplar` — exemplar appended inline
- [x] `testCompositeGaugeHistogram` — `gcount`/`gsum` fields
- [x] `testCompositeSummary` — basic single-line summary output
- [x] `testCompositeSummaryWithCreatedAndExemplar` — `st@` + exemplar
- [x] `testExemplarComplianceSkipsExemplarWithoutTimestamp` — compliance mode
  drops exemplars without timestamps; default mode still emits them
- [x] All 22 existing tests pass unchanged (non-composite path unaffected)

### DIFF
--- a/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
+++ b/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
@@ -200,7 +200,8 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
 
   private void writeHistogram(Writer writer, HistogramSnapshot snapshot, EscapingScheme scheme)
       throws IOException {
-    if (!openMetrics2Properties.getCompositeValues()) {
+    if (!openMetrics2Properties.getCompositeValues()
+        && !openMetrics2Properties.getExemplarCompliance()) {
       om1Writer.writeHistogram(writer, snapshot, scheme);
       return;
     }
@@ -273,7 +274,8 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
 
   private void writeSummary(Writer writer, SummarySnapshot snapshot, EscapingScheme scheme)
       throws IOException {
-    if (!openMetrics2Properties.getCompositeValues()) {
+    if (!openMetrics2Properties.getCompositeValues()
+        && !openMetrics2Properties.getExemplarCompliance()) {
       om1Writer.writeSummary(writer, snapshot, scheme);
       return;
     }
--- a/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
+++ b/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
@@ -14,9 +14,7 @@ import io.prometheus.metrics.config.OpenMetrics2Properties;
 import io.prometheus.metrics.model.snapshots.ClassicHistogramBuckets;
 import io.prometheus.metrics.model.snapshots.CounterSnapshot;
 import io.prometheus.metrics.model.snapshots.DataPointSnapshot;
-import io.prometheus.metrics.model.snapshots.DistributionDataPointSnapshot;
 import io.prometheus.metrics.model.snapshots.Exemplar;
-import io.prometheus.metrics.model.snapshots.Exemplars;
 import io.prometheus.metrics.model.snapshots.GaugeSnapshot;
 import io.prometheus.metrics.model.snapshots.HistogramSnapshot;
 import io.prometheus.metrics.model.snapshots.InfoSnapshot;
@@ -36,7 +34,6 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
-import java.util.List;
 import javax.annotation.Nullable;
 
 /**
@@ -92,6 +89,7 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
   private final OpenMetrics2Properties openMetrics2Properties;
   private final boolean createdTimestampsEnabled;
   private final boolean exemplarsOnAllMetricTypesEnabled;
+  private final OpenMetricsTextFormatWriter om1Writer;
 
   /**
    * @param openMetrics2Properties OpenMetrics 2.0 feature flags
@@ -106,6 +104,8 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
     this.openMetrics2Properties = openMetrics2Properties;
     this.createdTimestampsEnabled = createdTimestampsEnabled;
     this.exemplarsOnAllMetricTypesEnabled = exemplarsOnAllMetricTypesEnabled;
+    this.om1Writer =
+        new OpenMetricsTextFormatWriter(createdTimestampsEnabled, exemplarsOnAllMetricTypesEnabled);
   }
 
   public static Builder builder() {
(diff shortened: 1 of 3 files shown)


prometheus-metrics-exposition-textformats/src/test/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriterTest.java
```diff
@@ -325,6 +328,236 @@ void testEmptySnapshot() throws IOException {
...
+
+    // Compliance mode: exemplar WITH timestamp is emitted
+    assertThat(write(withTs, complianceWriter)).contains("# {trace_id=\"aaa\"} 1.0 1672850685.829");
+
+    // Compliance mode: exemplar WITHOUT timestamp is skipped
+    assertThat(write(withoutTs, complianceWriter)).doesNotContain("# {");
+
+    // Default mode: exemplar without timestamp is still emitted (just no timestamp)
+    assertThat(write(withoutTs, defaultWriter)).contains("# {trace_id=\"bbb\"} 2.0\n");
+  }
```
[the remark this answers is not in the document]
zeitlinger: Added coverage in `1ac9cc83` for histogram and summary exemplars without timestamps under `exemplarCompliance=true`, alongside the delegation fix in `OpenMetrics2TextFormatWriter`.

prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
```diff
@@ -442,23 +469,33 @@ private void writeNameAndLabels(
...
+      throws IOException {
+    if (exemplar == null) {
+      return;
+    }
+    if (!openMetrics2Properties.getExemplarCompliance()) {
+      om1Writer.writeExemplar(writer, exemplar, scheme);
+      return;
+    }
+    // exemplarCompliance=true: exemplars MUST have a timestamp per the OM2 spec.
+    if (exemplar.hasTimestamp()) {
```
jaydeluca, line 494: if its a MUST, should we be throwing an exception here if it's not set instead of just skipping? or at the very least we should be logging something?

prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
```diff
@@ -200,50 +200,66 @@ private void writeGauge(Writer writer, GaugeSnapshot snapshot, EscapingScheme sc
...
-        writeScrapeTimestampAndExemplar(writer, data, exemplar, scheme);
+    writeNameAndLabels(writer, name, null, data.getLabels(), scheme);
+    writer.write('{');
+    writer.write(countKey);
+    writer.write(':');
+    writeLong(writer, data.getCount());
+    writer.write(',');
+    writer.write(sumKey);
+    writer.write(':');
+    writeDouble(writer, data.getSum());
```
jaydeluca, line 239: for the [summary metrics ](https://github.com/prometheus/client_java/pull/1991/files#diff-3ec3af9a1fe8e649310c4733780012c41c874331ea5c7c8ef5f2885d20ee9d85R311)we do a check first

```suggestion
	if (data.hasSum()) {
	  writer.write(',');
	  writer.write(sumKey);
	  writer.write(':');
	  writeDouble(writer, data.getSum());
	}
```

prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
```diff
@@ -200,50 +200,66 @@ private void writeGauge(Writer writer, GaugeSnapshot snapshot, EscapingScheme sc
 
   private void writeHistogram(Writer writer, HistogramSnapshot snapshot, EscapingScheme scheme)
       throws IOException {
+    if (!openMetrics2Properties.getCompositeValues()
+        && !openMetrics2Properties.getExemplarCompliance()) {
+      om1Writer.writeHistogram(writer, snapshot, scheme);
+      return;
```
jaydeluca, line 206: If we set only `exemplarCompliance(true)` as a side effect changes the entire output format from multi-line OM1 histograms to single-line composite values.

the existing tests `testExemplarComplianceSkipsHistogramExemplarWithoutTimestamp` and `testExemplarComplianceSkipsSummaryExemplarWithoutTimestamp` only assert doesNotContain("# {") and they never check the surrounding structure, so the silent switch to composite format is not tested

prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
zeitlinger: This is already guarded on the current head. `writeCompositeSummaryDataPoint(...)` now emits `sum` only when `data.hasSum()` is true.

prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
zeitlinger: I’d avoid throwing or logging here. `exemplarCompliance=true` is meant to make exposition compliant with the OM2 requirement that exemplars carry a timestamp, and silently dropping only the invalid exemplar is the least disruptive behavior. Failing the whole scrape or adding hot-path log noise felt like too much for an output-side compliance flag.

jaydeluca (approved):

PR #2042, opened 2026-04-23
feat: Add OM2 native histogram text output

zeitlinger:
## Summary
Adds OpenMetrics 2.0 text output for native histograms behind the `nativeHistograms` flag.

- serializes native histogram fields (`schema`, `zero_threshold`, `zero_count`, native spans/buckets)
- supports native gauge histograms with `gcount`/`gsum`
- supports mixed native + classic histograms with classic `bucket` output last
- emits all histogram exemplars on composite/native histogram samples
- adds unit/integration coverage for native histogram content negotiation and edge cases

## Validation
- `./mvnw test -pl prometheus-metrics-exposition-textformats -Dtest=OpenMetrics2TextFormatWriterTest,ExpositionFormatsTest -Dcoverage.skip=true -Dcheckstyle.skip=true`
- `mise run build`

### DIFF
--- a/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
+++ b/prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
@@ -212,8 +212,7 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
       throws IOException {
     boolean compositeHistogram =
         openMetrics2Properties.getCompositeValues() || openMetrics2Properties.getNativeHistograms();
-    if (!compositeHistogram
-        && !openMetrics2Properties.getExemplarCompliance()) {
+    if (!compositeHistogram && !openMetrics2Properties.getExemplarCompliance()) {
       om1Writer.writeHistogram(writer, snapshot, scheme);
       return;
     }
@@ -341,8 +340,8 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
     writer.write(']');
   }
 
-  private void writeNativeBucketFields(
-      Writer writer, String prefix, NativeHistogramBuckets buckets) throws IOException {
+  private void writeNativeBucketFields(Writer writer, String prefix, NativeHistogramBuckets buckets)
+      throws IOException {
     if (buckets.size() == 0) {
       return;
     }
@@ -382,8 +381,8 @@ public class OpenMetrics2TextFormatWriter implements ExpositionFormatWriter {
     writeNativeBucketSpan(writer, spanOffset, spanLength, firstSpan);
   }
 
-  private boolean writeNativeBucketSpan(
-      Writer writer, int offset, int length, boolean firstSpan) throws IOException {
+  private boolean writeNativeBucketSpan(Writer writer, int offset, int length, boolean firstSpan)
+      throws IOException {
     if (!firstSpan) {
       writer.write(',');
     }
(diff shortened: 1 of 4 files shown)


prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/OpenMetrics2TextFormatWriter.java
```diff
@@ -177,8 +179,16 @@ private void writeCounter(Writer writer, CounterSnapshot snapshot, EscapingSchem
     for (CounterSnapshot.CounterDataPointSnapshot data : snapshot.getDataPoints()) {
       writeNameAndLabels(writer, counterName, null, data.getLabels(), scheme);
       writeDouble(writer, data.getValue());
-      writeScrapeTimestampAndExemplar(writer, data, data.getExemplar(), scheme);
-      writeCreated(writer, counterName, data, scheme);
+      if (data.hasScrapeTimestamp()) {
+        writer.write(' ');
+        writeOpenMetricsTimestamp(writer, data.getScrapeTimestampMillis());
+      }
+      if (createdTimestampsEnabled && data.hasCreatedTimestamp()) {
```
jaydeluca:
```suggestion
      if (data.hasCreatedTimestamp()) {
```

should we even have `createdTimestampsEnabled`? It isn't a config in OM2 as far as I can tell and only exists in OM1, and the other metric types don't account for it, meaning the counter currently has a differnet default behavior than the others. 

tests that demonstrate this (they both fail right now):

```java
  @Test
  void testCounterStartTimestampWithDefaultWriter() throws IOException {
    // Demonstrates the gap: the default OM2 writer (createdTimestampsEnabled=false) silently
    // drops st@ for counters, even though histograms and summaries emit st@ unconditionally.
    // This test currently FAILS. It will pass once the createdTimestampsEnabled guard is
    // removed from writeCounter so that counter behaviour matches histogram/summary behaviour.
    MetricSnapshots snapshots =
        MetricSnapshots.of(
            CounterSnapshot.builder()
                .name("my_counter")
                .dataPoint(
                    CounterSnapshot.CounterDataPointSnapshot.builder()
                        .value(42.0)
                        .createdTimestampMillis(1672850385800L)
                        .build())
                .build());

    String output = writeWithOM2(snapshots);

    assertThat(output)
        .isEqualTo(
            "# TYPE my_counter counter\n"
                + "my_counter 42.0 st@1672850385.800\n"
                + "# EOF\n");
  }

  @Test
  void testCounterAndHistogramEmitStartTimestampConsistently() throws IOException {
    // Demonstrates the asymmetry: a histogram with a created timestamp emits st@ with the
    // default writer; a counter with an identical created timestamp does not.
    // The histogram assertion already passes; the counter assertion currently FAILS.
    // Both will pass once the createdTimestampsEnabled guard is removed from writeCounter.
    long createdMs = 1672850385800L;

    MetricSnapshots counterSnapshots =
        MetricSnapshots.of(
            CounterSnapshot.builder()
                .name("my_counter")
                .dataPoint(
                    CounterSnapshot.CounterDataPointSnapshot.builder()
                        .value(1.0)
                        .createdTimestampMillis(createdMs)
                        .build())
                .build());

    MetricSnapshots histogramSnapshots =
        MetricSnapshots.of(
            HistogramSnapshot.builder()
                .name("my_histogram")
                .dataPoint(
                    HistogramSnapshot.HistogramDataPointSnapshot.builder()
                        .sum(1.0)
                        .createdTimestampMillis(createdMs)
                        .classicHistogramBuckets(
                            ClassicHistogramBuckets.builder()
                                .bucket(Double.POSITIVE_INFINITY, 1)
                                .build())
                        .build())
                .build());

    OpenMetrics2TextFormatWriter writer =
        OpenMetrics2TextFormatWriter.builder()
            .setOpenMetrics2Properties(
                OpenMetrics2Properties.builder().compositeValues(true).build())
            .build();

    String counterOutput = write(counterSnapshots, writer);
    String histogramOutput = write(histogramSnapshots, writer);

    assertThat(counterOutput).contains("st@1672850385.800");
    assertThat(histogramOutput).contains("st@1672850385.800");
  }

```

zeitlinger: Thanks, good catch. I removed the counter-only `createdTimestampsEnabled` guard so OM2 counters now emit `st@` whenever the datapoint has a created timestamp, matching histograms/summaries and the OM2 sample format.

I also added coverage for the default OM2 writer and counter/histogram start timestamp consistency. `createdTimestampsEnabled` remains only for delegated OM1 `_created` output.

jaydeluca (approved):

PR #2045, opened 2026-04-23
chore: update flint to 0.21.0

zeitlinger:
Updates the flint binary pin to v0.20.4 and reruns `flint init -y` so the generated lint workflow and linter config match the current release.

Validation:
- `flint init -y`
- `flint init -y` again for idempotence
- `git diff --check`


zeitlinger:
@jaydeluca please take a look

.github/config/.rumdl.toml
```diff
@@ -0,0 +1,5 @@
+[MD013]
```
jaydeluca, line 1: im not sure i understand why we have so many config files? can you explain a little bit how this is all is suposed to work? If i encounter a problem for example, is it going to be easy to know whether i need to update rumdl, taplo, yamllint, flint, etc?

zeitlinger: Great feedback - I'll create a PR to flint to explain it better - https://github.com/grafana/flint/#built-in-linter-registry already shows what linter is used per file type.

In essence, each linter has it's own config file and flint is just an glue layer on top.

jaydeluca: its getting hard to know whether flint is adding or removing complexity here

jaydeluca (approved):

PR #2097, opened 2026-05-05
docs: clarify suffix-handling behavior after #1955

### LINKED ISSUE #2095: bug: restore OM1/protobuf suffix compatibility after #1955
## Summary

`feat: move suffix handling to scrape time (#1955)` introduced a user-visible
compatibility regression for non-OpenMetrics exposition.

For gauge names that merely end in suffix-like dotted strings such as
`.created` or `.total`, OM1 and protobuf output changed from the legacy stripped
name to the literal suffixed name.

This surfaced while wiring downstream Micrometer compatibility in
https://github.com/prometheus/client_java/pull/2093.

## Regression

Examples:

| Input | Previous OM1/protobuf output | Regressed output | Format |
|---|---|---|---|
| `Gauge("test3.created")` | `test3` | `test3_created` | OM1, protobuf |
| `Gauge("test6.total")` | `test6` | `test6_total` | OM1, protobuf |

OpenMetrics can reasonably keep the newer literal behavior, but the OM1 and
protobuf change breaks downstream expectations and is not just an internal-only
representation change.

## Why this is a bug

This was initially easy to interpret as a docs/migration issue, but the
Micrometer downstream test failure turned out to be exercising a real behavior
change in what users see from OM1/protobuf scrapes.

That is distinct from the separate Micrometer adapter bug also found during the
same work:

- Micrometer needed to expose snapshot-derived `MultiCollector` metadata for
  registration-time collision detection.

But for the suffix case, the prom-side issue is specifically that non-OM2
output changed incompatibly for dotted gauge names.

## Expected behavior

For OM1 and protobuf compatibility:

- `Gauge("test3.created")` should expose as `test3`
- `Gauge("test6.total")` should expose as `test6`

OpenMetrics may continue to expose the literal encoded names.

## Status

Addressed in:

- PR: https://github.com/prometheus/client_java/pull/2093

That PR restores the legacy OM1/protobuf behavior while keeping OpenMetrics on
literal-name handling. The added Micrometer job in that PR is regression
coverage for the bug, not the bug fix itself.

ADITYA-CODE-SOURCE:
Fixes #2095.

Clarifies that the suffix-handling change from #1955 is not only an internal or OM2-only change.
It also affects user-visible exposition in Prometheus-facing formats for some arbitrary metric names
with suffix-like endings, while real counter/info semantics mostly preserve previous visible output.

Changes:
- clarify the 1.6.0 release notes in `CHANGELOG.md`
- add concrete before/after examples in `docs/content/getting-started/registry.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,24 @@
 > As part of the OM2 work, metric-name suffix handling moved from metric creation time to scrape
 > time. A positive side effect is that metric names are now more flexible across the board, for
 > example names ending in suffixes like `_total` are accepted where they were previously rejected.
+> This is not just an internal model change and not just an OM2 change: it also affects
+> user-visible exposition in OM1, OpenMetrics, and protobuf for some arbitrary metric names ending
+> in suffix-like strings such as `.created`, `.total`, `.info`, or `.bucket`.
+>
+> For real counter and info semantics, the exposed names are mostly preserved:
+>
+> | Example | Before 1.6.0 | Since 1.6.0 |
+> | --- | --- | --- |
+> | `Counter("events")` | `events_total` | `events_total` |
+> | `Counter("events_total")` | `events_total` | `events_total` |
+>
+> For arbitrary names ending in suffix-like strings, the exposed output may change:
+>
+> | Example | Before 1.6.0 | Since 1.6.0 |
+> | --- | --- | --- |
+> | `Gauge("test3.created")` | `test3` | `test3_created` |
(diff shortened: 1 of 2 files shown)


ADITYA-CODE-SOURCE:
Hi @zeitlinger
 I opened this docs-only PR for #2095.
The required GitHub Actions checks are currently waiting for maintainer approval because this PR comes from a fork. When you have a chance, could you please approve the workflows to run and take a look?

Thanks!

zeitlinger:
Thanks for the contribution, but I realized that this is a bug and should be fixed.

PR #2093, opened 2026-05-04
fix: restore OM1/protobuf micrometer suffix compatibility

zeitlinger:
Fixes https://github.com/prometheus/client_java/issues/2095

## Summary

Restores OM1/protobuf compatibility for dotted gauge names after
`feat: move suffix handling to scrape time (#1955)`.

The bug was that non-OpenMetrics exposition changed visible output for gauge
names that merely ended in suffix-like dotted strings such as `.created` and
`.total`.

Examples:

- `Gauge("test3.created")` regressed from `test3` to `test3_created`
- `Gauge("test6.total")` regressed from `test6` to `test6_total`

This PR restores the legacy OM1/protobuf behavior while keeping OpenMetrics on
literal-name handling.

## What changed

- Fix OM1 text exposition for dotted gauge names ending in `.created` and
  `.total`
- Fix protobuf exposition for the same compatibility cases
- Add regression tests that cover the restored OM1/protobuf behavior and the
  preserved OpenMetrics behavior
- Add downstream Micrometer compatibility tasks and workflow so this class of
  regression is exercised automatically against Micrometer
- Pin the downstream Micrometer target to an immutable commit SHA so PR CI is
  deterministic

## Related Micrometer work

While debugging this, we also found a separate Micrometer adapter bug:
registration-time collision detection needed snapshot-derived
`MultiCollector` metadata.

That fix lives on the downstream Micrometer branch used by the compatibility
job. The Micrometer coverage added here is regression protection; the prom-side
bug fix in this PR is the OM1/protobuf suffix compatibility restoration.

## Related

- Issue: https://github.com/prometheus/client_java/issues/2095

## Test plan

- `mise run build`
- `mise run lint`
- `./mvnw test -pl prometheus-metrics-exposition-textformats,prometheus-metrics-exposition-formats -Dtest=ExpositionFormatsTest,ProtobufExpositionFormatsTest -Dcoverage.skip=true -Dcheckstyle.skip=true`
- `GRADLE_USER_HOME=/tmp/gradle-micro ./gradlew --no-daemon -I /tmp/micrometer-prom-local.init.gradle :micrometer-registry-prometheus:test --tests io.micrometer.prometheusmetrics.PrometheusMeterRegistryTest`
- `GRADLE_USER_HOME=/tmp/gradle-micro ./gradlew --no-daemon -I /tmp/micrometer-prom-local.init.gradle :micrometer-registry-prometheus:test`

PR #2100, opened 2026-05-05
fix: restore legacy suffix compatibility

### LINKED ISSUE #2095: bug: restore OM1/protobuf suffix compatibility after #1955

zeitlinger:
Fixes https://github.com/prometheus/client_java/issues/2095

## Summary

Restores OM1/protobuf compatibility for dotted gauge names after
`feat: move suffix handling to scrape time (#1955)`.

The bug was that non-OpenMetrics exposition changed visible output for gauge
names that merely ended in suffix-like dotted strings such as `.created` and
`.total`.

Examples:

- `Gauge("test3.created")` regressed from `test3` to `test3_created`
- `Gauge("test6.total")` regressed from `test6` to `test6_total`

This PR restores the legacy OM1/protobuf behavior while keeping OpenMetrics on
literal-name handling.

This is the extracted prom-side fix from #2093. The Micrometer workflow and
related downstream testing were split into a stacked follow-up PR so this can
merge independently.

## What changed

- Fix OM1 text exposition for dotted gauge names ending in `.created` and
  `.total`
- Fix protobuf exposition for the same compatibility cases
- Add regression tests that cover the restored OM1/protobuf behavior and the
  preserved OpenMetrics behavior
- Clean up protobuf family-name resolution so legacy gauge handling lives in
  one path instead of pre-rewriting metadata objects

## Follow-up stacked PR

- Micrometer workflow/task split: https://github.com/zeitlinger/client_java/pull/1

## Related

- Replaces: https://github.com/prometheus/client_java/pull/2093
- Issue: https://github.com/prometheus/client_java/issues/2095

## Testing

- `mise run build`
- `mise run lint`
- `./mvnw test -pl prometheus-metrics-exposition-textformats,prometheus-metrics-exposition-formats -Dtest=ExpositionFormatsTest,ProtobufExpositionFormatsTest,DuplicateNamesProtobufTest -Dcoverage.skip=true -Dcheckstyle.skip=true`

### DIFF
--- a/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/internal/PrometheusProtobufWriterImpl.java
+++ b/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/internal/PrometheusProtobufWriterImpl.java
@@ -21,6 +21,7 @@ import io.prometheus.metrics.model.snapshots.MetricMetadata;
 import io.prometheus.metrics.model.snapshots.MetricSnapshot;
 import io.prometheus.metrics.model.snapshots.MetricSnapshots;
 import io.prometheus.metrics.model.snapshots.NativeHistogramBuckets;
+import io.prometheus.metrics.model.snapshots.PrometheusNaming;
 import io.prometheus.metrics.model.snapshots.Quantiles;
 import io.prometheus.metrics.model.snapshots.SnapshotEscaper;
 import io.prometheus.metrics.model.snapshots.StateSetSnapshot;
@@ -82,7 +83,7 @@ public class PrometheusProtobufWriterImpl implements ExpositionFormatWriter {
         builder.addMetric(convert(data, scheme));
       }
       setMetadataUnlessEmpty(
-          builder, snapshot.getMetadata(), null, Metrics.MetricType.GAUGE, scheme);
+          builder, snapshot.getMetadata(), null, Metrics.MetricType.GAUGE, scheme, true);
     } else if (snapshot instanceof HistogramSnapshot) {
       HistogramSnapshot histogram = (HistogramSnapshot) snapshot;
       for (HistogramSnapshot.HistogramDataPointSnapshot data : histogram.getDataPoints()) {
@@ -290,23 +291,51 @@ public class PrometheusProtobufWriterImpl implements ExpositionFormatWriter {
       @Nullable String nameSuffix,
       Metrics.MetricType type,
       EscapingScheme scheme) {
+    setMetadataUnlessEmpty(builder, metadata, nameSuffix, type, scheme, false);
+  }
+
+  private void setMetadataUnlessEmpty(
+      Metrics.MetricFamily.Builder builder,
+      MetricMetadata metadata,
+      @Nullable String nameSuffix,
+      Metrics.MetricType type,
+      EscapingScheme scheme,
+      boolean normalizeLegacyGaugeName) {
     if (builder.getMetricCount() == 0) {
       return;
     }
-    if (nameSuffix == null) {
-      builder.setName(SnapshotEscaper.getMetadataName(metadata, scheme));
-    } else {
-      String expositionBaseName = SnapshotEscaper.getExpositionBaseMetadataName(metadata, scheme);
-      if (expositionBaseName.endsWith(nameSuffix)) {
-        builder.setName(expositionBaseName);
-      } else {
-        builder.setName(SnapshotEscaper.getMetadataName(metadata, scheme) + nameSuffix);
-      }
-    }
+    builder.setName(
+        resolveMetricFamilyName(metadata, nameSuffix, scheme, normalizeLegacyGaugeName));
     if (metadata.getHelp() != null) {
       builder.setHelp(metadata.getHelp());
     }
     builder.setType(type);
+  }
+
+  private String resolveMetricFamilyName(
+      MetricMetadata metadata,
+      @Nullable String nameSuffix,
+      EscapingScheme scheme,
+      boolean normalizeLegacyGaugeName) {
+    if (normalizeLegacyGaugeName) {
+      String originalName = metadata.getOriginalName();
+      if (originalName.endsWith(".created")) {
+        return PrometheusNaming.escapeName(
+            originalName.substring(0, originalName.length() - ".created".length()), scheme);
+      }
+      if (originalName.endsWith(".total")) {
+        return PrometheusNaming.escapeName(
+            originalName.substring(0, originalName.length() - ".total".length()), scheme);
+      }
+    }
+    if (nameSuffix == null) {
+      return SnapshotEscaper.getMetadataName(metadata, scheme);
+    }
+    String expositionBaseName = SnapshotEscaper.getExpositionBaseMetadataName(metadata, scheme);
+    if (expositionBaseName.endsWith(nameSuffix)) {
+      return expositionBaseName;
+    }
+    return SnapshotEscaper.getMetadataName(metadata, scheme) + nameSuffix;
   }
 
   private long getNativeCount(HistogramSnapshot.HistogramDataPointSnapshot data) {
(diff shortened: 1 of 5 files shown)


prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/PrometheusTextFormatWriter.java
```diff
@@ -475,6 +476,19 @@ private static String resolveBaseName(String fullName, String suffix) {
...
+  private static String resolveLegacyGaugeName(MetricMetadata metadata, EscapingScheme scheme) {
+    String originalName = metadata.getOriginalName();
+    if (originalName.endsWith(".created")) {
+      return PrometheusNaming.escapeName(
+          originalName.substring(0, originalName.length() - ".created".length()), scheme);
+    }
+    if (originalName.endsWith(".total")) {
+      return PrometheusNaming.escapeName(
+          originalName.substring(0, originalName.length() - ".total".length()), scheme);
+    }
```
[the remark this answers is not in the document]
zeitlinger: Good catch. I moved the legacy suffix detection onto the raw pre-escape metric name and routed the text writer through the shared helper, so DOTS_ESCAPING and VALUE_ENCODING_ESCAPING keep the same legacy gauge normalization without duplicating the logic.

prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/internal/PrometheusProtobufWriterImpl.java
```diff
@@ -290,25 +291,53 @@ private void setMetadataUnlessEmpty(
...
+    if (normalizeLegacyGaugeName) {
+      String originalName = metadata.getOriginalName();
+      if (originalName.endsWith(".created")) {
+        return PrometheusNaming.escapeName(
+            originalName.substring(0, originalName.length() - ".created".length()), scheme);
+      }
+      if (originalName.endsWith(".total")) {
+        return PrometheusNaming.escapeName(
+            originalName.substring(0, originalName.length() - ".total".length()), scheme);
+      }
```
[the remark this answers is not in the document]
zeitlinger: Fixed this the same way on the protobuf path: the family-name normalization now uses the raw original metric name before escaping, via the shared helper, and I added coverage for DOTS_ESCAPING and VALUE_ENCODING_ESCAPING so both paths stay aligned.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/SnapshotEscaper.java
```diff
@@ -113,6 +113,26 @@ public static String getOriginalMetadataName(MetricMetadata metadata, EscapingSc
...
+      MetricMetadata metadata, String rawOriginalName, EscapingScheme scheme) {
+    String legacyGaugeBaseName = stripLegacyGaugeSuffix(rawOriginalName);
+    if (legacyGaugeBaseName != null) {
+      return PrometheusNaming.escapeName(legacyGaugeBaseName, scheme);
+    }
+    return getMetadataName(metadata, scheme);
+  }
+
+  @Nullable
+  private static String stripLegacyGaugeSuffix(String rawOriginalName) {
```
jaydeluca, line 126: trying to scrutinize changes against pre 1.60.0 behavior:

In v1.5.0, `writeGauge` called `getMetadataName(metadata, scheme)` which
returned `prometheusName` (dots → underscores) unchanged. No scrape-time suffix
stripping existed. This branch strips `.total` and `.created` unconditionally at
scrape time for all gauges regardless of origin.

```java
  // Documents the behavior gap in getLegacyGaugeName(): suffix stripping is applied
  // unconditionally to every GaugeSnapshot whose originalName ends in ".total" or ".created",
  // with no way to distinguish a genuine semantic name from a legacy bridged counter/timestamp.
  //
  // A gauge named "memory.total" is a legitimate semantic name.
  // Pre-1.6.0 this gauge emitted as "memory_total" (UNDERSCORE_ESCAPING) because no stripping
  // occurred at scrape time. The current implementation strips ".total" unconditionally, producing
  // "memory" — silently breaking any dashboard or alert keyed on "memory_total".
  @Test
  void testGaugeWithSemanticTotalSuffix_suffixIsStrippedUnconditionally() throws IOException {
    GaugeSnapshot memoryTotal =
        GaugeSnapshot.builder()
            .name("memory.total")
            .dataPoint(GaugeDataPointSnapshot.builder().value(1024).build())
            .build();

    // UNDERSCORE_ESCAPING: current code strips ".total" → "memory".
    // Pre-1.6.0 compatible output would be "memory_total".
    assertPrometheusText(
        """
        # TYPE memory gauge
        memory 1024.0
        """,
        memoryTotal);

    // DOTS_ESCAPING: current code strips ".total" from rawOriginalName → "memory".
    // Pre-1.6.0 compatible output would be "memory_dot_total".
    assertPrometheusText(
        """
        # TYPE memory gauge
        memory 1024.0
        """,
        memoryTotal,
        EscapingScheme.DOTS_ESCAPING);

    // ALLOW_UTF8: current code strips ".total" → "memory".
    // A Prometheus server with UTF-8 support would expect "memory.total" (the full UTF-8 name).
    ByteArrayOutputStream out = new ByteArrayOutputStream();
    getPrometheusWriter(PrometheusTextFormatWriter.builder().setIncludeCreatedTimestamps(false))
        .write(out, MetricSnapshots.of(memoryTotal), EscapingScheme.ALLOW_UTF8);
    assertThat(out)
        .hasToString(
            """
            # TYPE memory gauge
            memory 1024.0
            """);
  }
```

jaydeluca: i dont think this is actually valid

jaydeluca (approved):

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/SnapshotEscaper.java
zeitlinger: Thanks — I think the important comparison is the **rendered output**, because that is what regressed and what this PR is restoring.

For the concrete example `Gauge("test6.total")`:

| Format | v1.5.1 | After `#1955` | This PR | Effect |
|---|---|---|---|---|
| OM1 text | `test6` | `test6_total` | `test6` | restores v1.5.1 output |
| Protobuf family name | `test6` | `test6_total` | `test6` | restores v1.5.1 output |
| OpenMetrics | n/a / not the compatibility target here | literal suffixed behavior | same | unchanged by this PR |

And separately, the underlying name flow is still:

| Stage | v1.5.1 | After `#1955` | This PR |
|---|---|---|---|
| User-provided name | `test6.total` | `test6.total` | `test6.total` |
| Stored metadata | `test6.total` | `test6.total` | `test6.total` |
| Non-OM rendered output | `test6` | `test6_total` | `test6` |

So I agree with the narrow historical point that `v1.5.1` did not do this stripping inside `writeGauge()` itself. But the user-visible compatibility question here is the emitted OM1/protobuf name, and on that axis this PR is intentionally restoring the `v1.5.1` behavior that `#1955` regressed.

Issue #1411, opened 2025-06-04
Exporter Pushgateway runtime issue in versions 1.3.4 and above


rafauu:
Version 1.3.3 works correctly
I believe it is caused by https://github.com/prometheus/client_java/pull/1190
I am using default format in pushgateway which I think is protobuf.

Dependencies in pom.xml:
```
<dependency>
  <groupId>io.prometheus</groupId>
  <artifactId>prometheus-metrics-core</artifactId>
  <version>1.3.4</version>
</dependency>
<dependency>
  <groupId>io.prometheus</groupId>
  <artifactId>prometheus-metrics-exporter-pushgateway</artifactId>
  <version>1.3.4</version>
</dependency>
```

Runtime issue from shaded jar:
```
Exception in thread (...) java.lang.ExceptionInInitializerError
Caused by: java.lang.RuntimeException: class io.prometheus.metrics.expositionformats.PrometheusProtobufWriter is not available
at io.prometheus.metrics.exporter.pushgateway.PushGateway.<init>(PushGateway.java:99)
at io.prometheus.metrics.exporter.pushgateway.PushGateway.<init>(PushGateway.java:77)
at io.prometheus.metrics.exporter.pushgateway.PushGateway$Builder.build(PushGateway.java:456)
(...)
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.RuntimeException: class io.prometheus.metrics.expositionformats.PrometheusProtobufWriter is not available
```

So I guess it complains about
https://github.com/prometheus/client_java/blob/main/prometheus-metrics-exporter-pushgateway/src/main/java/io/prometheus/metrics/exporter/pushgateway/PushGateway.java#L9
or
https://github.com/prometheus/client_java/blob/main/prometheus-metrics-exporter-pushgateway/src/main/java/io/prometheus/metrics/exporter/pushgateway/PushGateway.java#L114

rafauu:
I can see it was addressed before https://github.com/prometheus/jmx_exporter/issues/1133

rafauu:
@zeitlinger I was playing with this maven shade filtering. Interestingly enough when you change (in your approach) 
`<include>io/prometheus/metrics/expositionformats/internal/PrometheusProtobufWriterImpl</include>`
into
`<include>**</include>`
it starts working. I'm not sure why tbh but as long as it works I'm good.

[...]

PR #2106, opened 2026-05-08
docs: document PushGateway shading workaround

ADITYA-CODE-SOURCE:
Documents #1411.

Adds documentation for the PushGateway runtime issue seen with shaded jars when  removes protobuf classes loaded via reflection.

Changes:
- add a troubleshooting note to the PushGateway docs
- clarify the shading and dependency requirements for protobuf exposition

### DIFF
--- a/docs/content/exporters/formats.md
+++ b/docs/content/exporters/formats.md
@@ -66,7 +66,7 @@ You can exclude the shaded protobuf classes including the
 `prometheus-metrics-exposition-formats-no-protobuf` module and excluding the
 `prometheus-metrics-exposition-formats` module in your build file.
 
-If you are using the PushGateway in a shaded jar with `minimizeJar=true`, do not use this setup.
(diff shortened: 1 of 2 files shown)


docs/content/exporters/formats.md
```diff
@@ -66,6 +66,12 @@ You can exclude the shaded protobuf classes including the
 `prometheus-metrics-exposition-formats-no-protobuf` module and excluding the
 `prometheus-metrics-exposition-formats` module in your build file.
 
+If you are using the PushGateway in a shaded jar with `minimizeJar=true`, do not use this setup.
```
zeitlinger: Nit: "do not use this setup" is a bit indirect. Suggest making it explicit:

```suggestion
If you are using the PushGateway in a shaded jar with `minimizeJar=true`, do not exclude the protobuf classes.
```

zeitlinger (approved):
LGTM, thanks!

PR #2114, opened 2026-05-13
feat: add typed metric family descriptors

zeitlinger:
Adds typed metric family descriptors and typed metadata support for the model snapshots.

This is the typed-descriptor branch for downstreams that want to provide registration-time metadata explicitly. The #1800 Collector/MultiCollector registration metadata hooks are already optional via default methods, so unmodified downstreams should not need this PR just to keep working.

This PR now also deprecates the fragmented registration metadata API (`getPrometheusName()`, `getMetricType()`, `getLabelNames()`, and `getMetadata()` plus the `MultiCollector` variants) in favor of `getMetricFamilyDescriptor()` / `getMetricFamilyDescriptors()`. The deprecated methods remain bridged by default implementations for compatibility.

Related validation:

- #2121 validates unmodified Micrometer independently of #2114, against `main` + #2124.
- #2123 validates a Micrometer branch that explicitly uses `MetricFamilyDescriptor` to implement the existing registration metadata hooks without invoking scrape/sample callbacks during registration.


zeitlinger:
Opened draft validation PR #2121 to run the combined compatibility stack: opt-in typed descriptors, #2094 suffix sanitization, and the Micrometer compatibility workflow. This is intended as a proof/test PR so #2114 can stay focused.

zeitlinger:
Opened #2122 as the smaller alternative implementation for #2114: same typed descriptor API, but legacy snapshot builders remain on plain metadata so the new descriptor behavior is opt-in. #2121 remains the combined validation PR with #2094 + Micrometer compatibility.

zeitlinger:
> Opened #2122 as the smaller alternative implementation for #2114: same typed descriptor API, but legacy snapshot builders remain on plain metadata so the new descriptor behavior is opt-in. #2121 remains the combined validation PR with #2094 + Micrometer compatibility.

closed that

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/CounterSnapshotTest.java
```diff
@@ -93,7 +94,18 @@ void testEmptyCounter() {
   @Test
   void testTotalSuffixPresent() {
     CounterSnapshot snapshot = CounterSnapshot.builder().name("test_total").build();
-    assertThat(snapshot.getMetadata().getPrometheusName()).isEqualTo("test_total");
+    assertThat(snapshot.getMetadata().getPrometheusName()).isEqualTo("test");
```
jaydeluca, line 99: isn't this the wrong behavior?

prometheus-metrics-instrumentation-caffeine/src/main/java/io/prometheus/metrics/instrumentation/caffeine/CacheMetricsCollector.java
```diff
@@ -313,7 +315,12 @@ public MetricSnapshots collect() {
         .build();
   }
 
+  /**
+   * @deprecated Use {@code getMetricFamilyDescriptors()} instead.
```
jaydeluca:
```suggestion
   * @deprecated Use {@link getMetricFamilyDescriptors()} instead.
```

prometheus-metrics-instrumentation-guava/src/main/java/io/prometheus/metrics/instrumentation/guava/CacheMetricsCollector.java
```diff
@@ -216,7 +216,12 @@ public MetricSnapshots collect() {
     return metricSnapshotsBuilder.build();
   }
 
+  /**
+   * @deprecated Use {@code getMetricFamilyDescriptors()} instead.
```
jaydeluca:
```suggestion
   * @deprecated Use {@link getMetricFamilyDescriptors()} instead.
```

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricFamilyDescriptor.java
```diff
@@ -0,0 +1,275 @@
...
+    @Override
+    protected SummaryBuilder self() {
+      return this;
+    }
+  }
+
+  public static final class InfoBuilder extends Builder<InfoBuilder> {
+
+    @Override
+    public InfoBuilder unit(@Nullable Unit unit) {
```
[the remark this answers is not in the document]
zeitlinger: Pushed a follow-up. I kept the new descriptor API strict here: `unit(null)` now fails consistently with an explicit null-argument error, and `InfoBuilder` still rejects any non-null unit.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricFamilyDescriptor.java
```diff
@@ -0,0 +1,275 @@
...
+    protected InfoBuilder self() {
+      return this;
+    }
+  }
+
+  public static final class StateSetBuilder extends Builder<StateSetBuilder> {
+
+    @Override
+    public StateSetBuilder unit(@Nullable Unit unit) {
+      throw new IllegalArgumentException("State set metric cannot have a unit.");
```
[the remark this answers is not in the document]
zeitlinger: Pushed a follow-up here too: `unit(null)` now fails explicitly as an invalid argument, while non-null units are still rejected for state-set metrics.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricFamilyDescriptor.java
```diff
@@ -0,0 +1,275 @@
...
+      return self();
+    }
+
+    public T labelNames(String... labelNames) {
+      Collections.addAll(this.labelNames, labelNames);
+      return self();
+    }
+
+    public T labelNames(Collection<String> labelNames) {
+      this.labelNames.addAll(labelNames);
```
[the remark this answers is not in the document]
zeitlinger: Good catch. I pushed a follow-up that normalizes descriptor label names in the builder so registration-time validation matches exposition-time escaping behavior.

prometheus-metrics-model/src/test/java/io/prometheus/metrics/model/snapshots/CounterSnapshotTest.java
zeitlinger: I added a clarifying comment here. The derived metadata name is intentionally the internal base name without the `_total` suffix; the exposition base/original names still retain the literal counter name used on the wire.

prometheus-metrics-instrumentation-caffeine/src/main/java/io/prometheus/metrics/instrumentation/caffeine/CacheMetricsCollector.java
zeitlinger: Applied, thanks.

prometheus-metrics-instrumentation-guava/src/main/java/io/prometheus/metrics/instrumentation/guava/CacheMetricsCollector.java
zeitlinger: Applied, thanks.

jaydeluca (approved):

PR #2121, opened 2026-05-18
test: validate Micrometer vanilla compatibility

zeitlinger:
Draft validation PR for the unmodified Micrometer compatibility story.

This intentionally does **not** depend on #2114. Vanilla Micrometer does not use
the typed descriptor API and does not implement the #1800 registration metadata
hooks, so this PR validates the patch-compatible path independently of typed
descriptors.

This validates upstream `micrometer-metrics/micrometer@main` against:

- current `main`, which already includes #2100 and #2124 (reserved suffix
  stripping in `PrometheusNaming.sanitizeMetricName()`).
- Micrometer compatibility test tooling/workflow from zeitlinger/client_java#1.

Local validation:

- `mise run lint`
- `MICROMETER_DIR=/tmp/micrometer-compat-vanilla-2124 mise run micrometer:test`

PR #2122, opened 2026-05-18
feat: add opt-in typed metric family descriptors

zeitlinger:
Alternative to #2114 that keeps typed metric family descriptors opt-in.

Difference from #2114:

- Keeps the legacy snapshot builders using plain `MetricMetadata` construction.
- Does not derive counter/info metadata automatically in `CounterSnapshot.Builder` or `InfoSnapshot.Builder`.
- Restores the builder fields to private and avoids forcing legacy/deprecated consumers into the typed descriptor path.

The new typed descriptor API remains available for collectors that explicitly opt in to the new behavior.

Related validation:

- #2121 validates vanilla Micrometer compatibility against this branch plus #2094.
- #2123 validates a Micrometer branch that explicitly opts in via `MetricFamilyDescriptor`.

Local validation:

- `mise run build`

### DIFF
--- a/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/CounterSnapshot.java
+++ b/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/CounterSnapshot.java
@@ -180,14 +180,6 @@ public class CounterSnapshot extends MetricSnapshot {
     }
 
     @Override
-    protected MetricMetadata buildMetadata() {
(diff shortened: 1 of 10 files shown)


zeitlinger:
Closing this alternate branch. After re-checking #1800, the Collector/MultiCollector registration metadata hooks are already optional via default methods. #2114 is the right typed-descriptor branch; the validation PRs are now restacked on #2114 instead of this alternate.

PR #2157, opened 2026-05-28
chore: add Zizmor

zeitlinger:


.gitattributes
```diff
@@ -0,0 +1,3 @@
+CHANGELOG.md linguist-generated
```
zeitlinger, line 1: flint now automatically skips generated files

jaydeluca (approved):

PR #2161, opened 2026-05-29
perf: Refactored sorting to use optimized sort algorithms

dhoard:
Refactored code to use optimized sort algorithms.

### DIFF
--- a/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/Labels.java
+++ b/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/Labels.java
@@ -177,31 +177,16 @@ public final class Labels implements Comparable<Labels>, Iterable<Label> {
     }
   }
 
+  /**
+   * Sorts all three parallel arrays in place using introspective quicksort.
+   *
+   * <p>Algorithm: 3-way quicksort with insertion sort for tiny partitions and heapsort fallback at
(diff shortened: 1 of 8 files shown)


prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/Labels.java
```diff
@@ -441,4 +426,177 @@ public Labels build() {
...
     }
   }
+
+  /**
+   * In-place introsort for label arrays, keyed by {@code prometheusNames}.
+   *
+   * <p>Uses 3-way quicksort partitioning for large ranges, insertion sort for tiny ranges, and a
+   * heapsort fallback at the recursion-depth limit to guarantee O(n log n) worst-case complexity.
+   */
+  private static final class StringArraySorter {
```
zeitlinger, line 436: looks pretty similar to last sorter - can we use generics or would this break perf?

zeitlinger (commented):
Great - only wonder if this could be re-usable. Either generics or generating code code.

prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/Labels.java
dhoard: Most likely. I chose to keep the code isolated like the Bubble sort implementation.

zeitlinger (approved):

PR #2167, opened 2026-06-01
test: validate JMX Exporter compatibility

zeitlinger:
Draft validation PR for the unmodified JMX Exporter compatibility story.

This intentionally does **not** depend on #2114. Vanilla JMX Exporter does not
use the typed descriptor API, so this PR validates the patch-compatible path
independently of typed descriptors.

This validates upstream `prometheus/jmx_exporter@main` against current
`client_java` by installing local `io.prometheus` artifacts and running the JMX
Exporter collector/common/javaagent/standalone Maven tests against them.

Local validation:

- `mise run jmx-exporter:test`
- `mise run lint:fix`
- `mise run lint`
- `actionlint .github/workflows/jmx-exporter-compatibility.yml`


dhoard:
@zeitlinger are you wanting to run JMX Exporter smoke test configuration or a quick test configuration?

Smoke test configuration (all integration tests, 6 Java distributions, 2 Prometheus distributions)

- https://github.com/prometheus/jmx_exporter/blob/main/run-smoke-test.sh
- more Java version coverage, longer execution time

Quick test configuration (all integration tests, but 1 Java distribution and 1 Prometheus distribution)

- https://github.com/prometheus/jmx_exporter/blob/main/run-quick-test.sh
- less Java version coverage, less execution time

dhoard (approved):
All in all looks good based on my understand of mise. We need to determine which level we want to run tests. smoke test configuration or quick test configuration

.github/workflows/jmx-exporter-compatibility.yml
```diff
@@ -0,0 +1,39 @@
...
+          sha256: cfac593469d028d7ae5fe36e37bd7c59118b5238e92d8a876209578464f24a84
+      - name: Cache local Maven repository
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+      - name: Run JMX Exporter compatibility tests
+        env:
+          JMX_EXPORTER_REPOSITORY: ${{ inputs.repository || 'prometheus/jmx_exporter' }}
+          JMX_EXPORTER_REF: ${{ inputs.ref || 'main' }}
```
jaydeluca: for both micrometer and JMX, could we use renovate and a tag /sha instead of `main`? just thinking of a case where either of these projects were to be compromised, we wouldn't want to pull from main and run them

jaydeluca (approved):
one comment but it is non-blocking, we can do a followup

zeitlinger:
> All in all looks good based on my understand of mise. We need to determine which level we want to run tests. smoke test configuration or quick test configuration

sorry I missed the comment before merging - what tests should we run in your opinion?

zeitlinger:
> > All in all looks good based on my understand of mise. We need to determine which level we want to run tests. smoke test configuration or quick test configuration
> 
> sorry I missed the comment before merging - what tests should we run in your opinion?

I think I'll start with quick

dhoard:
> > > All in all looks good based on my understand of mise. We need to determine which level we want to run tests. smoke test configuration or quick test configuration
> > 
> > 
> > sorry I missed the comment before merging - what tests should we run in your opinion?
> 
> I think I'll start with quick

quick configuration. This should be enough to sanity-check things.

zeitlinger:
> quick configuration. This should be enough to sanity-check things.

https://github.com/prometheus/client_java/pull/2178

PR #2168, opened 2026-06-01
feat: Add StableApi marker and API diff check

zeitlinger:
## Summary

- add `@StableApi` as the opt-in marker for published Java API
- seed the guessed stable API surface from docs plus Micrometer/JMX usage
- add `mise run api-diff` using japicmp against the configured baseline
- add an API diff workflow that fails on incompatible published API changes

## Notes

This is the bootstrap PR for the annotation-based API surface. Since `1.5.1` does not contain
`@StableApi`, the first diff is noisy and mostly shows the seeded API surface as new. After a
release contains the annotations, future diffs should be normal compatibility diffs.

The workflow does not post PR comments or upload artifacts. If the check fails, run this locally:

```bash
mise run api-diff
```

Reports are written to `**/target/japicmp/*`.

Intentional incompatible changes can be accepted by adding the PR label
`breaking-api-change-accepted`.

## Validation

- `mise run api-diff`
- `mise run build`
- `mise run lint`


.github/workflows/api-diff.yml
```diff
@@ -0,0 +1,79 @@
...
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
+  workflow_dispatch:
+    inputs:
+      baseline_version:
+        description: Version to compare the PR artifacts against
+        required: false
+        default: "1.5.1"
```
jaydeluca, line 17: how should we keep this up to date? should we create an issue to add a post-release step automation?

docs/stable-api-rationale.md
```diff
@@ -0,0 +1,325 @@
+# Stable API rationale
```
jaydeluca: this was good to have as review context, but we can probably remove at some point

jaydeluca (approved):

docs/stable-api-rationale.md
zeitlinger: removed

.github/workflows/api-diff.yml
zeitlinger: created https://github.com/prometheus/client_java/issues/2170

PR #2185, opened 2026-06-04
ci: verify release secrets before deploy

zeitlinger:
## Summary
- Replace `Debug gpg key` step with explicit checks for Sonatype token, GPG key import, and GPG passphrase match.
- Each failure points at the matching `RELEASING.md` section.
- Add `If the Sonatype Central Token is Invalid` section to `RELEASING.md`.

Context: recent release attempt failed with `403` on Maven Central upload because the user token was stale. The token mismatch was only visible deep in Maven output.

Note: Also fixes zizmor - I'm still investigating why flint didn't catch this before

## Test plan
- [ ] Re-run release workflow with current secrets and confirm verify step passes (or surfaces the actual broken secret).


.github/workflows/release.yml
```diff
@@ -18,14 +18,52 @@ jobs:
...
+          fail() {
+            echo "::error::$1"
+            echo "::error::Fix: see '$2' in RELEASING.md"
+            exit 1
+          }
+
+          # Sonatype token
+          response=$(curl -s -u "${MAVEN_USERNAME}:${MAVEN_CENTRAL_TOKEN}" \
+            "https://central.sonatype.com/api/v1/publisher/status?id=test")
+          echo "Sonatype response: ${response}"
```
jaydeluca, line 37: i wonder if this could be risky in terms of logging anything sensitive

zeitlinger: only says is token is valid or not

jaydeluca (approved):

PR #2191, opened 2026-06-04
feat: Add custom labels to exemplars

### LINKED ISSUE #1994: Is it possible to add a custom metadata identifier when using Exemplars?
Hello. We are currently upgrading from Prometheus old client 0.x to 1.x.

In the previous version, when using ExemplarSampler and Exemplar, the structure allowed us to include separate custom metadata, so in our system we used one additional identifier for management purposes.

However, while reviewing the 1.x version during the upgrade, it appears that the SpanContext and Exemplar classes only provide trace_id and span_id as identifiers, and do not seem to allow adding separate custom metadata. It is possible that I may have missed an extension point.

Is there any way to include a separate custom metadata identifier when using Exemplars?

jaydeluca:
Resolves https://github.com/prometheus/client_java/issues/1994


zeitlinger (changes requested):
One review item: the new exemplar label suppliers are documented as applying to automatically-sampled exemplars, but the current implementation also injects them into custom exemplars via updateCustomExemplar(). That changes the semantics of observeWithExemplar/incWithExemplar, which no longer remain fully caller-controlled when a global supplier is installed. Please avoid merging the extra supplier labels on the custom-exemplar path (while keeping trace/span enrichment if desired).

zeitlinger (approved):

PR #2194, opened 2026-06-05
fix: keep custom exemplar labels caller-controlled

zeitlinger:
## Summary
- keep supplier-provided exemplar labels on the automatic sampling path only
- preserve explicit custom exemplar labels as caller-controlled
- update tests to cover the intended custom-exemplar behavior

## Root cause
PR #2191 reused `doSampleExemplar()` from the custom exemplar path. After that method started merging per-metric and global label suppliers, `incWithExemplar(...)` and `observeWithExemplar(...)` also picked up supplier labels, even though the new API/docs describe that behavior for automatically-sampled exemplars only.

## Impact
Custom exemplars remain explicitly controlled by the caller, while automatically-sampled exemplars still receive the configured extra labels.

## Validation
- `mise run build`
- `./mvnw -q -pl prometheus-metrics-core -am -Dtest=CounterTest -Dsurefire.failIfNoSpecifiedTests=false test -Dcoverage.skip=true -Dcheckstyle.skip=true`
- `mise run lint:fix`


jaydeluca (approved):
nice thanks

PR #2202, opened 2026-06-08
feat: add MetricMetadata.Builder, deprecate wide constructors

zeitlinger:
## Summary

- Adds `MetricMetadata.builder()` with `name`, `help`, `unit`, `counterSuffix` fields
- Builder appends unit to the base name when absent, and appends `_total` to `expositionBaseName` when `counterSuffix=true`
- Deprecates the 4-arg and 5-arg constructors; internal callers (`MetricMetadataSupport`, `MetricMetadata.escape`) suppress the warning
- Updates `docs/apidiffs/current_vs_latest/prometheus-metrics-model.txt`

## Motivation

The OTel exporter ([opentelemetry/opentelemetry-java#8346](https://github.com/open-telemetry/opentelemetry-java/pull/8346)) needs to express per-strategy counter intent without pre-computing `expositionBaseName` manually. The builder encapsulates that logic and provides a cleaner public API for any downstream adapter that constructs `MetricMetadata` directly.

## Test plan

- [ ] `MetricMetadataTest` — 8 new builder tests covering: no unit, unit absent/present, counter suffix, counter + unit, UTF-8 name, non-counter, name-required validation
- [ ] Existing 4-arg/5-arg constructor tests annotated with `@SuppressWarnings("deprecation")`

### DIFF
--- a/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java
+++ b/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java
@@ -74,6 +74,82 @@ public final class MetricMetadata {
   }
 
   /**
+   * Creates a builder for {@link MetricMetadata}.
+   *
+   * <p>Use the builder instead of the multi-arg constructors for cleaner, more readable code:
+   *
+   * <pre>{@code
+   * MetricMetadata.builder()
+   *     .name("http_requests")
+   *     .help("Total HTTP requests")
+   *     .unit(Unit.BYTES)
+   *     .counterSuffix(true)
+   *     .build();
+   * }</pre>
+   */
+  @StableApi
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /** Builder for {@link MetricMetadata}. */
+  @StableApi
+  public static final class Builder {
+    @Nullable private String name;
+    @Nullable private String help;
+    @Nullable private Unit unit;
+    private boolean counterSuffix;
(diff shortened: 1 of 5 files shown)


prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricMetadata.java
```diff
@@ -73,6 +73,82 @@ public MetricMetadata(String name, @Nullable String help, @Nullable Unit unit) {
...
+   * }</pre>
+   */
+  @StableApi
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /** Builder for {@link MetricMetadata}. */
+  @StableApi
+  public static final class Builder {
```
jaydeluca, line 98: will this solve the use cases in this repo that we now have the suppressions on? i'm not sure it allows us to actually fully deprecate the constructors as-is.

For example, for MetricWithFixedMetadata i think we'd need:

```java
    this.metadata = MetricMetadata.builder()
      .name(name)
      .expositionBaseName(expositionBaseName)
      .originalName(originalName)
      .help(builder.help)
      .unit(builder.unit)
      .build();
```

It might be worth trying to migrate any of the ones where we suppress the deprecation warning, just to see if it's possible

zeitlinger: good idea - done

jaydeluca (approved):

PR #2214, opened 2026-06-11
fix: preserve OM2 info metric names

zeitlinger:
## Summary
- preserve OpenMetrics 2 info metric names instead of appending `_info`
- keep the existing name unchanged when it already includes `_info`
- update OM2 writer tests to cover both cases

## Testing
- `./mvnw test -pl prometheus-metrics-exposition-textformats -Dtest=OpenMetrics2TextFormatWriterTest -Dcoverage.skip=true -Dcheckstyle.skip=true`
- `mise run build`
- `mise run lint:fix`


zeitlinger:
contradicts spec

PR #2252, opened 2026-06-24
perf: optimize classic-only histogram observe path

### LINKED ISSUE #1915: Classic-only histogram: consider synchronized block instead of multi-LongAdder for observe() hot path
## Context

While benchmarking the [Prometheus shim PoC](https://github.com/prometheus/client_java/compare/main...zeitlinger:client_java:feat/metric-backend-spi) (bridging Prometheus client API to the OTel SDK), I found that classic-only histograms are **30% faster through the OTel SDK** than through native Prometheus.

## Benchmark numbers (JMH, single thread)

| Path | `observe()` latency |
|------|-------------------|
| Native Prometheus (classic-only) | 10.5 ns |
| OTel SDK (explicit bucket histogram) | 7.3 ns |

## Root cause

Native Prometheus `doObserve()` uses **3 separate CAS-based atomics** per call:

1. `classicBuckets[i].add(1)` — `LongAdder`
2. `sum.add(value)` — `DoubleAdder`
3. `count.increment()` — `LongAdder`

Plus a `buffer.append()` CAS attempt and volatile reads for reset/scale-down state.

The OTel SDK uses a **single `synchronized` block** with plain `+=`/`++` arithmetic:

```java
synchronized (lock) {
    this.sum += value;
    this.count++;
    this.counts[bucketIndex]++;
    // min/max tracking
}
```

In uncontended (single-thread) benchmarks, HotSpot elides the uncontended lock and optimizes the plain arithmetic freely, beating the multi-CAS approach.

## Suggestion

For **classic-only** histograms (where `nativeInitialSchema == CLASSIC_HISTOGRAM`), consider an alternative `doObserve()` implementation that uses a synchronized block with plain fields instead of multiple `LongAdder`/`DoubleAdder` instances. The buffer mechanism (needed for native histogram scale-down) could also be bypassed in classic-only mode.

This wouldn't affect native or hybrid histograms, which still need the current design.

## Multi-threaded consideration

The `LongAdder` approach was chosen for multi-threaded scalability (striped cells reduce contention). A synchronized block would serialize threads. However:

- Most real-world `observe()` calls happen on different label-value combinations (different data points), so contention on a single data point is rare
- Even under contention, the critical section is very short (~5 ns of arithmetic), so lock hold time is minimal
- A benchmark with 4 threads would clarify the actual tradeoff

Not a high priority — 10.5 ns is already excellent. But worth considering if classic histogram performance matters.

arnabnandy7:
## Summary

Optimizes `Histogram.classicOnly()` observations by adding a dedicated classic-only fast path that uses synchronized plain fields instead of the native/hybrid `LongAdder`/`DoubleAdder` plus buffer path.

Fixes #1915

## Changes

- Added classic-only storage for bucket counts, total count, and sum in `Histogram.DataPoint`.
- Routed `observe()` and `observeWithExemplar()` through the classic-only fast path when `nativeInitialSchema == CLASSIC_HISTOGRAM`.
- Kept native-only and hybrid histograms on the existing buffer/aduder implementation.
- Added multithreaded coverage for classic-only observe/collect behavior.

## Testing

- `.\mvnw compile -pl prometheus-metrics-core -Dcoverage.skip=true -Dcheckstyle.skip=true`
- `git diff --check`


zeitlinger:
the bench is not comparing to main - I'll fix that

arnabnandy7:
> the bench is not comparing to main - I'll fix that

thanks

zeitlinger (changes requested):
The shared-observer benchmark regression is blocking for this implementation: serializing every classic-only observation on the `DataPoint` monitor removes the contention-avoiding behavior we need on this path.

I opened #2329 as an **experimental alternate candidate** so we can compare an epoch-striped/double-buffered per-thread accumulator against `main` on the same benchmark runner. Its local four-thread JMH smoke result was 40,944.515 ± 302.666 ops/s, but that is not a controlled cross-branch result and should not be treated as proof. The full benchmark workflow for #2329 is queued at https://github.com/prometheus/client_java/actions/runs/29990012023.

The alternate is not production-ready yet. Before it could be considered for merge, it still needs:

- Java Memory Model review and preferably dedicated jcstress coverage for epoch handoff races;
- a decision on bounding/reclaiming per-thread cells under high thread churn;
- scrape latency and stalled-observer spin-wait measurements;
- evaluation of the epoch-flip/cell-scan cost in `getCount()` and `getSum()`;
- the full CI benchmark comparison and broader supported-JDK/architecture coverage.

The benchmark can be rerun by removing and re-adding the `benchmark` label on the alternate PR.

I recommend keeping #2252 blocked while we collect the alternate benchmark and decide whether the striped design can be hardened, rather than merging the global-monitor regression.

zeitlinger:
Follow-up: the alternate draft's current head is `6fe299560f57865c3ca33d399163a0639f84b47e`; the benchmark label has been toggled after the API-diff metadata commit. The current benchmark run is https://github.com/prometheus/client_java/actions/runs/29990237603 (queued/running). The blocking review and remaining-hardening caveats above are unchanged.

zeitlinger:
Closing this PR as superseded by [#2329](https://github.com/prometheus/client_java/pull/2329).

#2329 uses the striped accumulator design rather than the global-lock approach, addressing the contention regression identified here. It also includes the concurrency hardening and JCStress coverage; CI is green, and the [benchmark report](https://github.com/prometheus/client_java/pull/2329#issuecomment-5056712236) shows no counter regression and improved classic-histogram throughput.

Thank you for the groundwork.

PR #2255, opened 2026-06-26
ci: auto-label expected OTel api diffs

zeitlinger:
## Summary
- extend the OTel api-diff regeneration workflow to detect the expected shaded OTel incompatibility from generated japicmp reports
- upload a marker artifact only when the incompatible changes are limited to the expected shaded constructor removal
- add `breaking-api-change-accepted` to the open Renovate PR for `renovate/otel.instrumentation.version` after the workflow pushes regenerated apidiffs

## Why
The current workflow only regenerates `docs/apidiffs`. When an OTel bump changes the shaded package version, the PR can still fail the separate incompatible API gate until the acceptance label is added manually.

## Validation
- `mise run lint:fix`
- `actionlint .github/workflows/regenerate-api-diff-otel.yml`

### DIFF
--- a/.github/workflows/regenerate-api-diff-otel.yml
+++ b/.github/workflows/regenerate-api-diff-otel.yml
@@ -48,11 +48,41 @@ jobs:
             exit 1
           fi
           git diff --binary -- docs/apidiffs > /tmp/apidiffs.patch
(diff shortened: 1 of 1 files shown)


zeitlinger:
close in favor of https://github.com/prometheus/client_java/pull/2257

PR #2274, opened 2026-07-03
fix: correct native histogram bucket merge on scale-down

### LINKED ISSUE #2270: Bug in resizing buckets of a native histogram
https://github.com/prometheus/client_java/blob/2a2c73d7d23bfa291b10df85056027398e8a868d/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java#L617
It should be `int index = (keys[i] > 0 ? keys[i] + 1 : keys[i]) / 2`. (See https://github.com/prometheus/client_golang/blob/832cec72badf6ff6ad3bd5fed4338e4e27fd646b/prometheus/histogram.go#L1119, noting that Java handles negative integer division in the same way as Go.)

Because of this bug, when `doubleBucketWidth()` happens, counters of some buckets are merged into wrong new buckets (more specifically, the counter of a bucket with a negative even index is merged into a new bucket that is to the right of the correct one). This causes `histogram_count(rate(my_native_histogram[1m]))` (RPS calculation) to glitch with a weird spike, as Prometheus may treat such random change in counters as a _reset_. On the other hand, `histogram_quantile(..., my_native_histogram)` probably is not affected significantly (i.e. painfully visibly), as (some of) counters are merely shifted to adjacent buckets, and adjacent buckets have quite close boundaries, e.g. x1.044 for schema=4, I guess.

arnabnandy7:
## Summary

Fix native histogram bucket remapping when the bucket width is doubled during schema scale-down.

The previous merge formula used:

```
(keys[i] + 1) / 2
```

That works for positive bucket indexes, but not for negative even indexes. Java integer division truncates toward zero, so negative even buckets were merged into the bucket immediately to the right of the correct target.

This matches the behaviour described in #2270 and aligns the merge formula with client_golang:

```
(keys[i] > 0 ? keys[i] + 1 : keys[i]) / 2
```

## Why

When `doubleBucketWidth()` runs, counts from some negative native histogram buckets could be shifted into neighbouring buckets. That can make Prometheus interpret the reshuffled bucket counters as resets, causing spikes in queries such as:

```
histogram_count(rate(my_native_histogram[1m]))
```

## Tests

Added a focused regression test for negative, zero, and positive bucket index remapping during `doubleBucketWidth(Map)`.

Verified locally:

```
./mvnw test -pl prometheus-metrics-core -am \
  -Dtest=HistogramTest#testDoubleBucketWidthMergesNegativeIndexesCorrectly \
  -Dcoverage.skip=true \
  -Dcheckstyle.skip=true \
  -Dsurefire.failIfNoSpecifiedTests=false
```

Also verified:

```bash
./mvnw install -DskipTests -Dcoverage.skip=true
```

Fixes #2270

### DIFF
--- a/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java
+++ b/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java
@@ -614,7 +614,7 @@ public class Histogram extends StatefulMetric<DistributionDataPoint, Histogram.D
       }
       buckets.clear();
       for (i = 0; i < keys.length; i++) {
-        int index = (keys[i] + 1) / 2;
+        int index = (keys[i] > 0 ? keys[i] + 1 : keys[i]) / 2;
         LongAdder count = buckets.computeIfAbsent(index, k -> new LongAdder());
         count.add(values[i]);
       }
--- a/prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/HistogramTest.java
+++ b/prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/HistogramTest.java
@@ -28,7 +28,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


zeitlinger (approved):
thanks for the contribution!

PR #2316, opened 2026-07-19
fix: prevent Summary quantiles from collapsing to the minimum observation

### LINKED ISSUE #2292: Summary quantiles collapse to the minimum observation when 2·epsilon ≥ 1−quantile
## Version

`io.prometheus:prometheus-metrics-core:1.8.0` (the affected code in `CKMSQuantiles` is
unchanged since at least 1.3.x and is still present on `main`).

## Reproducer

```java
import io.prometheus.metrics.core.metrics.Summary;
import io.prometheus.metrics.model.snapshots.Quantile;
import io.prometheus.metrics.model.snapshots.SummarySnapshot;

import java.util.ArrayList;
import java.util.Collections;
import java.util.List;
import java.util.Random;

public class SummaryQuantileBugRepro {

    Summary summary =
        Summary.builder()
            .name("request_latency")
            .help("reproducer")
            .quantile(0.9, 0.05)
            .quantile(0.99, 0.005)
            .register(new PrometheusRegistry());

    // observe the values 1..100_000 shuffled, so the true p90 is 90_000 and p99 is 99_000
    List<Integer> values = new ArrayList<>();
    for (int i = 1; i <= 100_000; i++) {
      values.add(i);
    }
    Collections.shuffle(values, new Random(42));
    for (int v : values) {
      summary.observe(v);
    }

    SummarySnapshot.SummaryDataPointSnapshot dataPoint = summary.collect().getDataPoints().get(0);
    for (Quantile q : dataPoint.getQuantiles()) {
      System.out.printf(
          "quantile %.2f: reported = %10.1f, expected ~ %10.1f%n",
          q.getQuantile(), q.getValue(), q.getQuantile() * 100_000);
    }
  }
}
```

Output with 1.8.0:

```
quantile 0.90: reported =        1.0, expected ~    90000.0
quantile 0.99: reported =        1.0, expected ~    99000.0
```

Both quantiles report the **minimum** of all 100,000 observations. The result does not
depend on the input distribution or the shuffle seed. A single targeted quantile
`quantile(0.99, 0.005)` reproduces it too.

## How we found it

In our application we had

```java
.quantile(0.9, 0.05)
.quantile(0.99, 0.005)
```

on a latency Summary. It reported `p90 == p99 == 0.6ms` while a
classic `Histogram` fed by the same observations (cross-checked against an in-process
exact-percentile counter) showed p90 = 2ms, p99 = 44ms. The reported value was below the
true median. Adding a `.quantile(0.5, 0.05)` softens the failure but doesn't fix it: get(0.99)
then returns a value from around the 85th percentile.

## Observations

`CKMSQuantiles`:

1. **`compress()` destroys the sketch.** The error function from the CKMS paper
   (definition 5) allows a sample below a target `(q, ε)` at rank `r` to have width
   `2ε(n−r)/(1−q)`. When `2ε ≥ 1−q` — which holds with equality for both `(0.9, 0.05)` and
   `(0.99, 0.005)` — this is `≥ n−r`: one sample may span every rank from `r` to `n`.
   `compress()` then merges away everything between the middle of the distribution and the
   maximum; the sample list collapses to just a few samples no matter how many values are inserted.

2. **`get()` stops at the first wide sample.** The scan stops at the first sample with
   `r + g + delta > desiredRank + f(desiredRank)/2` and returns the previous sample's value.
   That rule is only correct when `g + delta` is small for all samples up to the target
   rank. Below a targeted quantile, deltas may legally be of order `n` — and freshly
   inserted samples get `delta = f(r) − 1` while `get()` flushes the insert buffer right
   before scanning, so maximally-wide fresh samples are always present at query time. The
   scan therefore stops far before the target rank.

## Tightening error/epsilon avoids the issue

The tight-epsilon (0.01 for p90, 0.001 for p99) Summary agrees with the histogram-derived p99,
while the identically-fed boundary-epsilon (0.05 for p90, 0.005 for p99) Summaries report `p90 == p99` from
the very same measurements. So the failure is fully determined by the quantile
configuration, not by the data.

## Side note

I have a fix with reproducer tests ready and happy to open a PR:  
it bounds merges in compress() so a sample can never span across the accuracy window of a target quantile, 
and changes get() to select the sample whose rank interval is centered closest to the desired rank.
With the reproducer config, get(0.99) goes from returning the minimum observation to a rank-accurate value.
The full existing test suite passes.

manduinca:
Fixes #2292

When a targeted quantile has `2*epsilon >= 1 - quantile`, all reported quantiles could collapse to the minimum observation. The query scan stopped too early: a freshly inserted low-rank sample can carry a `delta` on the order of `n`, so the running `rank + g + delta` exceeded the target far sooner than it should and returned a value near the minimum.

The fix has two parts, matching the diagnosis in the issue. `compress()` now keeps a merged sample from spanning a quantile's target rank, and the query uses the minimum of the error function over a sample's rank interval `[lo, hi]` rather than at a single point. To support that I split the single-argument `f(r)` into `f(lo, hi)` (the plain error function is `f(r, r)`), which is numerically identical for the existing call sites.

Added a regression test that reproduces the collapse and asserts distinct quantiles. `CKMSQuantilesTest` passes (18 tests).

### DIFF
--- a/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/CKMSQuantiles.java
+++ b/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/CKMSQuantiles.java
@@ -147,43 +147,71 @@ final class CKMSQuantiles {
       return samples.getLast().value;
     }
 
-    int r = 0; // sum of g's left of the current sample
     int desiredRank = (int) Math.ceil(q * n);
-    int upperBound = desiredRank + f(desiredRank) / 2;
 
-    ListIterator<Sample> iterator = samples.listIterator();
-    while (iterator.hasNext()) {
-      Sample sample = iterator.next();
-      if (r + sample.g + sample.delta > upperBound) {
-        iterator.previous(); // roll back the item.next() above
-        if (iterator.hasPrevious()) {
-          Sample result = iterator.previous();
-          return result.value;
-        } else {
-          return sample.value;
-        }
+    // Return the sample whose possible-rank interval [rMin, rMax] is centered closest to the
+    // desired rank. rMin is the sum of the g's up to and including the sample, rMax is rMin+delta.
+    //
+    // The previous implementation stopped at the first sample whose maximum rank
+    // (r + g + delta) exceeded desiredRank + f(desiredRank)/2 and returned the preceding sample.
+    // That rule only works while g + delta stays small for all samples up to the target rank.
+    // For a targeted quantile with 2*epsilon >= 1-quantile the error function allows delta of
+    // order n at low ranks, so a freshly inserted low-rank sample (delta = f(r) - 1) makes the
+    // scan stop far too early and return a value near the minimum observation. See issue #2292.
+    Sample result = samples.getFirst();
+    double smallestDistance = Double.POSITIVE_INFINITY;
+    int r = 0; // sum of g's left of the current sample
+    for (Sample sample : samples) {
+      int rMin = r + sample.g;
+      double center = rMin + sample.delta / 2.0;
+      double distance = Math.abs(center - desiredRank);
+      if (distance < smallestDistance) {
+        smallestDistance = distance;
+        result = sample;
       }
       r += sample.g;
     }
-    return samples.getLast().value;
+    return result.value;
   }
 
   /** Error function, as in definition 5 of the paper. */
   int f(int r) {
+    return f(r, r);
+  }
+
+  /**
+   * Minimum of the error function {@link #f(int)} over the rank interval {@code [lo, hi]}.
+   *
+   * <p>Each targeted quantile contributes a V-shaped error curve whose global minimum {@code
+   * 2*epsilon*n} is at rank {@code quantile*n}, so the minimum over the interval is reached at an
+   * interior target (when the interval straddles it) or otherwise at the endpoint nearest the
+   * target. {@link #compress()} uses this so that a merged sample can never span a target
+   * quantile's rank with a width coarser than that quantile's accuracy window (see issue #2292).
+   * With {@code lo == hi} this reduces to the plain error function at a single rank.
+   */
+  int f(int lo, int hi) {
     int minResult = Integer.MAX_VALUE;
     for (Quantile q : quantiles) {
       if (q.quantile == 0 || q.quantile == 1) {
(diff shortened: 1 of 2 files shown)


zeitlinger (changes requested):
Thanks for the focused reproducer. I found a correctness issue in the new rank-selection heuristic, plus a couple of changes that would make the fix easier to review:

1. `get()` currently chooses the sample whose possible-rank interval has the nearest center. That does not preserve the advertised rank bound. With values `1..10_000` shuffled using `new Random(2)`, a single `(q=0.99, epsilon=0.005)` configuration returns `9784`; the allowed range under the test's `2 * epsilon` bound is `[9800, 10000]`. The sample centered nearest to rank 9900 is therefore not a sufficient selection criterion. Please rework the selection against the CKMS error bounds and add this smaller deterministic regression case.

2. Please rename `f`. It is now overloaded for both a point rank and an interval, but the name gives no indication of either the mathematical meaning or the interval semantics. Descriptive names such as `errorBoundAtRank` and `minErrorBoundInRange` would make the compression logic much easier to audit. Please update the associated Javadocs and test references too.

3. The new tests only use `n=100_000`, and their broad `2 * epsilon` assertions do not verify the comment's claim that the quantiles remain distinct or exercise the `2 * epsilon > 1 - quantile` case. Please add direct rank-bound coverage for the failing smaller case and a strict-above-boundary case.

The two new boundary tests pass, but the additional deterministic case above fails on the current head.

olegkovalenko:
Heads-up: I've opened [#2396](https://github.com/prometheus/client_java/pull/2396) for the same issue (#2292). It shares the two-part diagnosis from this PR — the compress() merge bound and the early-stopping scan in get() — but bounds sample widths differently (window-anchored cap, applied at both merge and insert time) and keeps f() unchanged.

The deterministic case from the review here (values 1..10,000, Random(2), single (0.99, 0.005)) is included [#2396](https://github.com/prometheus/client_java/pull/2396) as a regression test and passes (returns 9940, within [9800, 10000]), along with the strict-above-boundary coverage the review asked for. Details and the comparison harness are linked in the PR description. Happy to have either PR land — flagging it mainly so the two efforts don't diverge silently.

zeitlinger:
closed in favor of https://github.com/prometheus/client_java/pull/2396

PR #2329, opened 2026-07-23
perf: experiment with striped classic histogram accumulator

zeitlinger:
## Summary

This is an **experimental alternate implementation** for benchmarking against `main` and the synchronized candidate in #2252. It is not being presented as production-ready yet.

For classic-only histograms, this replaces the existing `Buffer`/`LongAdder` recording path with an epoch-striped, double-buffered per-thread accumulator. Each recording thread owns a cell, observations update plain thread-confined bucket/count/sum fields, and collection advances a global epoch before draining the inactive buffers. This avoids a global `synchronized (DataPoint)` monitor on the shared four-thread observe path while preserving coherent bucket/count/sum snapshots. `getCount()` and `getSum()` use the same snapshot mechanism.

Native-only and hybrid histograms keep their existing implementation. The public API is unchanged.

## Local evidence

- `mise run lint`
- `./mvnw -pl prometheus-metrics-core -Dtest=HistogramTest test` — 35 tests passed
- focused concurrent collect/lifecycle tests repeated 5 times
- `./mvnw -pl prometheus-metrics-core -am verify` — reactor passed; core ran 157 tests
- `git diff --check`
- representative local JMH smoke run (JDK 25.0.3, 4 threads, 1 fork, 3x1s warmup, 5x1s measurement):
  - `HistogramBenchmark.prometheusClassic`: **40,944.515 ± 302.666 ops/s**

The local JMH number is a smoke result, not a controlled cross-branch conclusion. The `benchmark` workflow is being requested so this branch and `main` are measured on the same runner.

## Added concurrency coverage

- collection while eight shared observers record concurrently
- monotonic and coherent bucket/count/sum snapshots
- exact final `getCount()` / `getSum()` values
- retention of observations from short-lived recording threads
- `clear()` lifecycle behavior for old and newly-created data points

## Remaining hardening before merge consideration

- Review/validate the epoch handoff under the Java Memory Model, ideally with dedicated jcstress coverage rather than relying only on repeated JUnit stress tests.
- Decide how to bound or reclaim retained per-thread cells for workloads with high thread churn; cells currently live for the data point lifetime so completed-thread observations cannot be lost.
- Measure scrape latency and the spin-wait behavior when a recording thread is delayed after entering an epoch.
- Measure the cost of `getCount()` / `getSum()`, which currently perform an epoch flip and cell scan to return exact accumulated values.
- Run the full CI benchmark matrix and test additional supported JDKs/architectures before treating this as mergeable.

Related: #2252


zeitlinger:
Evaluation (local, JDK 25.0.4, x86_64, 1 fork / 2 warmups / 5 measurements; indicative rather than production benchmark):

- Shared-observer JMH: #2329 44.4K ops/s vs #2252 1.63K ops/s; per-thread #2329 44.9K vs #2252 64.7K.
- A temporary churn probe created 1,000/5,000 retained cells from one-shot threads; `getCount()` grew from about 1 ms to 5 ms, and collection was about 13.8 ms at 5,000 cells.

Recommendation: use #2329 as the basis, not #2252, but do not merge yet. Please add jcstress/JMM coverage, bound or reclaim retained cells, and define bounded scrape progress/latency when a writer is delayed. Existing focused concurrency tests pass repeatedly, but they do not close those risks.

zeitlinger:
Hardening update: pushed `03c3da30` (includes the additive merge of current `main`).

- Snapshot waits are bounded to a 1 ms total budget; a delayed writer's buffer is skipped safely and collected on a later epoch flip, so scrapes cannot spin indefinitely.
- Empty cells are safely deregistered/reclaimed after both buffers are collected and re-register when a thread-local is reused.
- Added JMM/concurrency coverage for stalled writers, eventual delayed-buffer collection, cell reclamation/reuse, and 8 concurrent writers (80,000 observations with coherent bucket/count/sum results).
- Added JMH probes for 1,000 short-lived recorder threads and the resulting epoch flip/cell scan. Local JDK 25.0.4, x86_64, 1 fork, 1×100 ms warmup, 2×100 ms measurement: `collect` 1,644 ops/s; `getCount` 1,941 ops/s. Thread setup is outside the timed method; these are indicative scrape/scan numbers, not a cross-branch comparison.

Focused accumulator/Histogram tests, repeated concurrency tests, benchmark compilation, and formatter/lint pass locally. Full CI benchmark matrix and cross-JDK/architecture validation remain appropriate before merge.

zeitlinger:
Follow-up hardening pushed in `ebe3d8e8`:

- Removed the shared (and per-cell) spin/wait budget entirely. A snapshot now drains a cell immediately when its inactive buffer is not being written, and skips only an actively-writing same-parity cell. This makes total handoff latency proportional to the current cell/bucket scan, independent of any writer stall; delayed cells are intentionally stale until a later snapshot.
- Added a multi-cell regression (one stalled cell plus four healthy cells) proving healthy cells are drained, plus active-writer pause/resume coverage.
- Added explicit comments documenting stale `collect()`, `getCount()`, and `getSum()` semantics.

Local focused tests: 5 accumulator tests and 35 Histogram tests pass, with the accumulator suite repeated five times. JMH smoke (JDK 25.0.4, x86_64, 1 fork, 1×100 ms warmup, 2×100 ms measurement, 1,000 short-lived recorders; setup excluded): collect 1,218 ops/s and getCount 1,600 ops/s. These are indicative scan-cost measurements only.

The current `native-tests` CI failure is unrelated infrastructure: Maven Wrapper could not download Maven 3.9.16 from repo.maven.apache.org. No review threads are open.

zeitlinger:
JCStress evidence added in `37792360`:

- Added a dedicated `prometheus-metrics-jcstress` module using JCStress 0.16, with tests for epoch publication, stalled-writer parity and eventual collection, cell reclamation/re-registration, and concurrent observation count/sum coherence.
- Built the shaded harness and ran the four accumulator tests on JDK 25.0.4 with 2 CPUs, sanity mode, 1 normal + 1 stress fork, 5 iterations × 20 ms: **112/112 planned results passed; 0 forbidden/error results**.
- Existing focused `ClassicOnlyAccumulatorTest`/`HistogramTest`, full Maven package, and Flint formatter/lint pass.

The stalled-writer test uses reflection only in the JCStress test module to model a pause at the internal epoch boundary; production code has no test hooks. Cross-JDK/architecture stress coverage remains for CI/maintainer validation.

zeitlinger:
## Release review: defer from 1.9.0

Decision from the pre-release review: keep this PR open for further hardening, but do not include it
in the upcoming 1.9.0 release.

Reviewed head: `c95bbdc32cb050bd0c5eb7695529bd0ea8b5644c`.

The recent reclamation, nonblocking snapshot, and JCStress work materially improves this candidate.
The remaining concern below is based on code inspection; it has **not been experimentally
reproduced** in this review.

### Potential permanent observation loss during cell reclamation

In [ClassicOnlyAccumulator.snapshot()](https://github.com/prometheus/client_java/blob/c95bbdc32cb050bd0c5eb7695529bd0ea8b5644c/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/ClassicOnlyAccumulator.java#L102-L108),
the idle/empty checks and cell removal are not atomic with
[writer entry](https://github.com/prometheus/client_java/blob/c95bbdc32cb050bd0c5eb7695529bd0ea8b5644c/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/ClassicOnlyAccumulator.java#L46-L72).
The following interleaving appears possible:

1. After draining the inactive buffer, the collector observes `NOT_WRITING` and both buffers empty,
   then pauses before `registered.compareAndSet(true, false)`.
2. The owning writer starts another observation in the current epoch, publishes `writingEpoch`,
   and passes both the registration and epoch checks.
3. The collector successfully unregisters the cell and removes it from `cells`.
4. The writer updates its buffer and returns normally.

Subsequent snapshots cannot discover that cell. If its owning thread never observes again, the
observation is permanently lost rather than merely delayed.

Before reconsidering this for release:
- Add a deterministic regression for this interleaving.
- Make reclamation and writer entry use a protocol that prevents removal after a writer has been
  admitted.
- Add actual concurrent reclamation/re-registration JCStress coverage. The current
  `ClassicOnlyAccumulatorReRegistrationTest` performs observe/reclaim/reuse sequentially in one
  actor, while its second actor is empty, so it does not exercise this handoff.
- Revalidate count, sum, and bucket coherence, eventual visibility, reclamation, and nonblocking
  snapshots after the fix.

### Release packaging

The PR also adds the JCStress harness to the consumer BOM and default reactor. Please confirm
whether publishing this test harness as a consumer-facing artifact is intentional; otherwise keep
it out of the consumer BOM and configure publication appropriately.

A [fresh benchmark run](https://github.com/prometheus/client_java/actions/runs/34959957707) has been
triggered. Benchmark results can establish performance value but cannot settle the reclamation
correctness concern above.

zeitlinger:
Keeping this experiment out of the 1.9.0 release. For a future investigation, the benchmark/design should cover classic, native, and hybrid histograms with parity across:

- concurrent observe, collect, and reset/churn workloads;
- per-thread and shared accumulator states;
- native dynamic sparse-bucket growth, schema downscaling, and reset reconciliation;
- bucket lookup and allocation profiling.

Classic histograms use fixed arrays and are comparatively straightforward. Native histograms have dynamic sparse buckets and schema transitions, while hybrid mode must keep classic/native state consistent. The goal is an equally well-proven optimization, not identical throughput across modes. The reclamation correctness concern should be resolved before making performance claims.

PR #2332, opened 2026-07-23
fix: hide HTTP scrape error details

### LINKED ISSUE #2283: Security Report: Full Stack Trace Disclosure in HTTP Error Responses
# Security Report: Full Stack Trace Disclosure in HTTP Error Responses

## 1. Vulnerability Summary

| Field                  | Value                                                                                         |
|------------------------|-----------------------------------------------------------------------------------------------|
| **Product**            | prometheus/client_java                                                                        |
| **Affected Version**   | 1.8.0 (main branch); all prior 1.x releases using `prometheus-metrics-exporter-httpserver` are likely affected |
| **Component**          | `prometheus-metrics-exporter-httpserver`                                                       |
| **Vulnerable File**    | `prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpExchangeAdapter.java` |
| **Vulnerable Lines**   | 103-141                                                                                       |
| **CWE**                | CWE-209: Generation of Error Message Containing Sensitive Information                         |
| **CVSS 3.1 Score**     | **5.3 (Medium)**                                                                              |
| **CVSS 3.1 Vector**    | `AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N`                                                       |
| **Commit (HEAD)**      | `a017f809`                                                                                    |
| **Discoverer**         | Security Research                                                                             |

---

## 2. Vulnerability Description

The `HttpExchangeAdapter` class in the `prometheus-metrics-exporter-httpserver` module exposes full Java stack traces to remote, unauthenticated HTTP clients when any exception occurs during a Prometheus metrics scrape. The method `sendErrorResponseWithStackTrace()` serializes the complete exception stack trace -- including exception type, message, chained causes, class names, method names, file names, and line numbers -- into the HTTP 500 response body and sends it directly to the requesting client.

The `/metrics` endpoint served by the built-in `HTTPServer` is accessible without authentication by default. No configuration option exists to suppress stack trace output in error responses. Any exception raised during the scrape cycle (configuration errors, metric collection failures, runtime exceptions from custom collectors) triggers the disclosure.

This is a classic information disclosure vulnerability. Stack traces are considered sensitive information because they reveal internal implementation details that assist attackers in reconnaissance and targeted exploitation.

---

## 3. Affected Component Architecture

The vulnerability sits at the intersection of three components:

```
PrometheusScrapeHandler.handleRequest()
        |
        |  catches IOException / RuntimeException
        v
PrometheusHttpExchange.handleException()  (interface)
        |
        |  implemented by
        v
HttpExchangeAdapter.sendErrorResponseWithStackTrace()  <-- VULNERABILITY
        |
        v
HTTP 500 Response with full stack trace sent to client
```

**Key observation:** The servlet-based adapters (`prometheus-metrics-exporter-servlet-jakarta` and `prometheus-metrics-exporter-servlet-javax`) are **not** affected. Their `handleException()` implementations re-throw the exception to the servlet container rather than serializing stack traces into the response. This vulnerability is specific to the standalone `HTTPServer` exporter module.

---

### 4.1 Primary Vulnerable Method

**File:** `prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpExchangeAdapter.java`
**Lines:** 103-141

```java
private void sendErrorResponseWithStackTrace(Exception requestHandlerException) {
    if (!responseSent) {
      responseSent = true;
      try {
        StringWriter stringWriter = new StringWriter();
        PrintWriter printWriter = new PrintWriter(stringWriter);
        printWriter.write("An Exception occurred while scraping metrics: ");
        requestHandlerException.printStackTrace(new PrintWriter(printWriter));
        byte[] stackTrace = stringWriter.toString().getBytes(StandardCharsets.UTF_8);
        httpExchange.getResponseHeaders().set("Content-Type", "text/plain; charset=utf-8");
        httpExchange.sendResponseHeaders(500, stackTrace.length);
        httpExchange.getResponseBody().write(stackTrace);
      } catch (IOException errorWriterException) {
        // ...
      }
    }
}
```

The critical issue is the call to `requestHandlerException.printStackTrace(new PrintWriter(printWriter))` on line 110, which dumps the entire stack trace into a `StringWriter` whose contents are then written verbatim to the HTTP response body on line 114.

### 4.2 Call Sites

The vulnerable method is invoked from the two `handleException()` overloads (lines 93-101):

```java
@Override
public void handleException(IOException e) throws IOException {
    sendErrorResponseWithStackTrace(e);
}

@Override
public void handleException(RuntimeException e) {
    sendErrorResponseWithStackTrace(e);
}
```

### 4.3 Trigger Point

`PrometheusScrapeHandler.handleRequest()` (lines 99-102) routes all caught exceptions through the exchange's `handleException()`:

```java
public void handleRequest(PrometheusHttpExchange exchange) throws IOException {
    try {
      // ... scrape metrics
    } catch (IOException e) {
      exchange.handleException(e);
    } catch (RuntimeException e) {
      exchange.handleException(e);
    } finally {
      exchange.close();
    }
}
```

---

## 5. CVSS 3.1 Detailed Breakdown

| Metric                     | Value          | Justification                                                                                         |
|----------------------------|----------------|-------------------------------------------------------------------------------------------------------|
| **Attack Vector (AV)**     | Network (N)    | The `/metrics` HTTP endpoint is network-accessible. The `HTTPServer` binds to a network socket (default: all interfaces). |
| **Attack Complexity (AC)** | Low (L)        | No special conditions required. The attacker only needs network access to the metrics port. An exception must occur during scraping, which can be triggered via environment variable misconfiguration or may happen naturally. |
| **Privileges Required (PR)** | None (N)    | The `HTTPServer` does not require authentication by default. The `Authenticator` is an optional builder parameter that is `null` unless explicitly configured. |
| **User Interaction (UI)**  | None (N)       | No user interaction is needed. The attacker sends an HTTP request directly to the endpoint.            |
| **Scope (S)**              | Unchanged (U)  | The vulnerability affects only the confidentiality of the vulnerable component; it does not cross trust boundaries to impact other components. |
| **Confidentiality (C)**    | Low (L)        | The disclosed information (stack traces, class names, library versions, configuration values, file paths) provides valuable reconnaissance data but does not directly expose credentials, secrets, or user data. |
| **Integrity (I)**          | None (N)       | No data modification occurs.                                                                          |
| **Availability (A)**       | None (N)       | No denial-of-service impact.                                                                          |

**CVSS 3.1 Vector String:** `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N`
**CVSS 3.1 Base Score:** **5.3 (Medium)**

---

## 6. Information Disclosed in Stack Traces

A stack trace exposed via this vulnerability can reveal the following categories of sensitive information:

| Category                          | Example                                                                                              |
|-----------------------------------|------------------------------------------------------------------------------------------------------|
| **Library version**               | Package naming conventions and class structures identify the exact `client_java` version             |
| **Internal class names**          | `io.prometheus.metrics.config.Util`, `io.prometheus.metrics.config.HistogramProperties`, etc.        |
| **Internal method names**         | `parseDoubleList`, `collect`, `scrape`                                                               |
| **Source file names and line numbers** | `Util.java:90`, `HistogramProperties.java:45`, `Histogram.java:198`                             |
| **Configuration values**          | Exception messages from `PrometheusPropertiesException` embed the invalid configuration value        |
| **JVM internals**                 | Thread stack frames expose JVM version, internal classes, and runtime behavior                        |
| **Application structure**         | Custom collector class names, application framework stack frames                                     |
| **File system paths**             | Absolute paths may appear in exception messages (e.g., properties file locations)                    |
| **Package structure**             | Full package hierarchy of the application and its dependencies                                       |

---

### 7.1 Prerequisites

- A Java application using `prometheus-metrics-exporter-httpserver` with the built-in `HTTPServer`
- Network access to the metrics HTTP port

### 7.2 Reproduction Steps

**Step 1: Configure an invalid property to guarantee an exception during scrape.**

```bash
export IO_PROMETHEUS_METRICS_HISTOGRAM_CLASSIC_UPPER_BOUNDS=invalid
```

This environment variable sets an invalid value for the histogram upper bounds configuration. When a Histogram metric is collected, `HistogramProperties` attempts to parse this value, throwing a `PrometheusPropertiesException`.

**Step 2: Start the application with a Prometheus HTTP server.**

Any Java application using the `HTTPServer` builder pattern:

```java
import io.prometheus.metrics.core.metrics.Histogram;

public class VulnerableApp {
        Histogram histogram = Histogram.builder()
            .name("example_histogram")
            .help("An example histogram")
            .register();

            .port(9400)
            .buildAndStart();

        System.out.println("Server started on port " + server.getPort());
        Thread.currentThread().join();
    }
}
```

**Step 3: Request the metrics endpoint.**

```bash
curl -s http://localhost:9400/metrics
```

**Step 4: Observe the full stack trace in the response.**

Expected HTTP 500 response body:

```
An Exception occurred while scraping metrics: io.prometheus.metrics.config.PrometheusPropertiesException:
  io.prometheus.metrics.histogram.classic_upper_bounds=invalid: Expecting comma separated list of double values
    at io.prometheus.metrics.config.Util.parseDoubleList(Util.java:90)
    at io.prometheus.metrics.config.HistogramProperties.<init>(HistogramProperties.java:45)
    at io.prometheus.metrics.core.metrics.Histogram.collect(Histogram.java:198)
    at io.prometheus.metrics.model.registry.PrometheusRegistry.scrape(PrometheusRegistry.java:67)
    at io.prometheus.metrics.exporter.common.PrometheusScrapeHandler.scrape(PrometheusScrapeHandler.java:143)
    at io.prometheus.metrics.exporter.common.PrometheusScrapeHandler.handleRequest(PrometheusScrapeHandler.java:63)
    at io.prometheus.metrics.exporter.httpserver.MetricsHandler.handle(MetricsHandler.java:25)
    at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:77)
    at sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:82)
    at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:80)
    at sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:692)
    at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:77)
    ...
```

### 7.3 Attack Scenario Without Misconfiguration

Even without attacker-controlled misconfiguration, stack traces are disclosed when:

1. A custom `Collector` registered in the `PrometheusRegistry` throws a `RuntimeException` during `collect()` (e.g., `NullPointerException`, `IllegalStateException`).
2. A transient I/O error occurs during metric serialization.
3. An internal consistency error is triggered by concurrent metric updates.

In production environments, these conditions are not uncommon, making the vulnerability exploitable through passive observation as well as active triggering.

---

### 8.1 Direct Impact

- **Information Disclosure:** Every error during metrics scraping sends implementation details to the network client. In environments where the metrics port is exposed (common in Kubernetes, cloud deployments, and monitoring infrastructure), any network-adjacent attacker can collect this information.

### 8.2 Reconnaissance Value

The disclosed stack traces provide an attacker with:

1. **Dependency mapping:** Exact library names and versions enable identification of known CVEs in dependencies.
2. **Code path analysis:** Method names and line numbers reveal execution flow, helping identify injection points or logic flaws.
3. **Configuration inference:** Exception messages from `PrometheusPropertiesException` and similar classes expose configuration parameters and their values.
4. **Technology fingerprinting:** JVM version, application server type, and framework stack frames identify the technology stack.

### 8.3 Deployment Context

The `HTTPServer` component is commonly used in:

- Standalone Java applications without a servlet container
- Java agents for JVM instrumentation
- Microservices exposing metrics alongside application traffic

In these deployments, the metrics port is frequently accessible to broader network segments than intended, amplifying the exposure risk.

---

### Affected Module

| Module | Affected | Reason |
|--------|----------|--------|
| `prometheus-metrics-exporter-httpserver` | **Yes** | `HttpExchangeAdapter.sendErrorResponseWithStackTrace()` writes full stack traces to the HTTP response body |

### Unaffected Modules

| Module | Affected | Reason |
|--------|----------|--------|
| `prometheus-metrics-exporter-servlet-jakarta` | No | `handleException()` re-throws the exception to the servlet container (`throw e;`) |
| `prometheus-metrics-exporter-servlet-javax` | No | `handleException()` re-throws the exception to the servlet container (`throw e;`) |

---

### 10.1 Immediate Fix

Replace the stack trace serialization in `sendErrorResponseWithStackTrace()` with a generic error message. Log the full stack trace server-side for debugging purposes.

```java
private void sendErrorResponseWithStackTrace(Exception requestHandlerException) {
    if (!responseSent) {
      responseSent = true;
      try {
        // Log the full stack trace server-side only
        Logger.getLogger(this.getClass().getName())
            .log(Level.SEVERE,
                "An Exception occurred while scraping metrics.",
                requestHandlerException);

        // Send a generic error message to the client
        byte[] message = "An internal error occurred while scraping metrics.\n"
            .getBytes(StandardCharsets.UTF_8);
        httpExchange.getResponseHeaders().set("Content-Type", "text/plain; charset=utf-8");
        httpExchange.sendResponseHeaders(500, message.length);
        httpExchange.getResponseBody().write(message);
      } catch (IOException errorWriterException) {
        // ... existing error handling
      }
    }
}
```

### 10.2 Configurable Error Verbosity

Introduce a configuration property (e.g., `io.prometheus.metrics.exporter.httpserver.debug`) that controls whether detailed error information is included in HTTP responses. Default to `false` (generic messages only).

### 10.3 Defense in Depth

- Document that the metrics HTTP port should be protected by network-level access controls (firewall rules, network policies).
- Encourage use of the `Authenticator` builder option when the metrics port is exposed to untrusted networks.
- Consider implementing the same error-handling pattern used by the servlet adapters, which delegate exception handling to the container rather than serializing stack traces.

---

## 11. References

- **CWE-209:** [Generation of Error Message Containing Sensitive Information](https://cwe.mitre.org/data/definitions/209.html)
- **OWASP:** [Improper Error Handling](https://owasp.org/www-community/Improper_Error_Handling)
- **OWASP Testing Guide:** [Testing for Stack Traces (OTG-ERR-002)](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Tests/08-Testing_for_Error_Handling/02-Testing_for_Stack_Traces)
- **Source Repository:** [prometheus/client_java](https://github.com/prometheus/client_java)
- **Vulnerable File:** [`HttpExchangeAdapter.java`](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpExchangeAdapter.java)

zeitlinger:
## Summary

- return a generic HTTP 500 body instead of exposing exception details
- include a safe hint explaining how to enable better diagnostics
- avoid adding server-side logging by default
- let users explicitly opt into either detailed HTTP responses or server-side error reporting
- verify the secure default does not expose the exception type or message

This is the focused replacement for the #2283 portion of #2297.

Fixes #2283

## Implementation plan

1. Make `HttpErrorHandlingPolicy` builder-based and pass the built policy through the `HTTPServer` builder to the exchange adapter. Response verbosity and error reporting remain orthogonal choices.
2. Make the default policy return a generic HTTP 500 response with a short hint to configure server-side error reporting for diagnostic details. The default must not add a new log entry for the scrape exception.
3. Let callers configure either axis independently:
   - attach a caller-supplied error reporter while keeping the generic response, so applications and Java agents can route diagnostics to an appropriate sink; and
   - enable an explicitly unsafe debug response mode that includes exception details in the HTTP body. Avoid “legacy” naming; the API and docs must make the disclosure risk clear.
4. Preserve the existing logging behavior for failures where the error response itself cannot be sent or response headers were already committed, because no useful client response remains possible in those paths.
5. Add focused tests for the secure default, diagnostic hint, independent verbosity/reporter configuration, reporter invocation and isolation, reporter failure handling, and the explicitly unsafe debug-response mode. Add user documentation for the available policies and their security tradeoffs, plus the repository’s release-please changelog entry linking to that documentation.

## Alternatives considered

- **Unconditional server-side logging:** rejected because it replaces the response disclosure with a new operational regression: one stack trace per failed scrape, with possible log-ingestion cost and application-logging side effects for unshaded integrations.
- **Keep the detailed response as the default:** rejected because it does not remediate #2283 unless every user discovers and enables the secure mode.
- **Generic response with no diagnostic guidance:** rejected because it leaves operators with a silent, unexplained HTTP 500 and no discoverable path to better diagnostics.
- **Hard-code rate limiting or deduplication in the adapter:** deferred in favor of a reporter abstraction. Correct suppression requires bounded state, concurrency handling, distinct-failure classification, and suppressed-count reporting; callers or a later reusable reporter can implement that policy without coupling it to HTTP response handling.
- **Ambient debug/verbosity configuration:** rejected in favor of an explicit builder option on `HttpErrorHandlingPolicy`, so re-enabling unsafe debug responses is deliberate and carries a clear security warning.
- **Network controls or authentication documentation alone:** rejected as the primary fix. They remain useful defense in depth but should not substitute for a safe default response.

## Validation

- `mise run lint:fix`
- `mise run build`
- `./mvnw test -pl prometheus-metrics-exporter-httpserver -Dcoverage.skip=true -Dcheckstyle.skip=true`

## Release note

Release Please will use this override for the generated changelog and GitHub release notes after a squash merge:

BEGIN_COMMIT_OVERRIDE
fix(httpserver): make scrape error responses secure and configurable

Scrape failures now return a generic HTTP 500 response by default. Applications can configure a server-side error reporter or explicitly enable an unsafe debug response containing exception details. See the [HTTPServer scrape error handling documentation](https://github.com/prometheus/client_java/blob/main/docs/content/exporters/httpserver.md#scrape-error-handling).
END_COMMIT_OVERRIDE


prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpExchangeAdapter.java
```diff
@@ -92,52 +94,42 @@ public HttpResponse getResponse() {
...
-        printWriter.write("An Exception occurred while scraping metrics: ");
-        requestHandlerException.printStackTrace(new PrintWriter(printWriter));
-        byte[] stackTrace = stringWriter.toString().getBytes(StandardCharsets.UTF_8);
         httpExchange.getResponseHeaders().set("Content-Type", "text/plain; charset=utf-8");
-        httpExchange.sendResponseHeaders(500, stackTrace.length);
-        httpExchange.getResponseBody().write(stackTrace);
+        httpExchange.sendResponseHeaders(500, ERROR_RESPONSE.length);
+        httpExchange.getResponseBody().write(ERROR_RESPONSE);
       } catch (IOException errorWriterException) {
-        // We want to avoid logging so that we don't mess with application logs when the HTTPServer
```
jaydeluca, line 116: this says it was a deliberate choice not to log here to avoid issues with the java agent. do you know the context behind that and whetehr this might cause issues?

zeitlinger: I'm recreating the PR - see description

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpExchangeAdapter.java
```diff
@@ -92,52 +97,53 @@ public HttpResponse getResponse() {
...
     } else {
       // If the exception occurs after response headers have been sent, it's too late to respond
       // with HTTP 500.
-      Logger.getLogger(this.getClass().getName())
-          .log(
-              Level.SEVERE,
-              "The Prometheus metrics HTTPServer caught an Exception while trying to send "
-                  + "the metrics response.",
-              requestHandlerException);
+      logger.log(
```
jaydeluca: This branch never calls `errorHandlingPolicy.report(...)` — it logs at `SEVERE` unconditionally
instead. Two things fall out of that:

**The policy leaks.** An application that wired `errorReporter` into its telemetry pipeline
silently misses this entire class of failure. It gets no callback, and the exception lands in
`java.util.logging` — exactly the destination the policy exists to avoid.

**The Java-agent scenario this PR is built for is unfixed here.** A collector that throws
deterministically during `writer.write(...)` reaches `handleException` with `responseSent ==
true` on every single scrape. That's a `SEVERE` + full stack trace into the host application's
logs every 15s, with no way to turn it off — which is the original complaint, on a different
code path.

Suggested shape (needs a package-private `boolean hasErrorReporter()` on the policy):

```java
} else {
  // If the exception occurs after response headers have been sent, it's too late to respond
  // with HTTP 500.
  if (errorHandlingPolicy.hasErrorReporter()) {
    reportException(requestHandlerException);
  } else {
    logger.log(
        Level.SEVERE,
        "The Prometheus metrics HTTPServer caught an Exception while trying to send "
            + "the metrics response.",
        requestHandlerException);
  }
}
```

Always-report-and-log-only-when-unconfigured works too — the requirement is just that a
configured reporter sees it.

There's no test covering this branch today, which is how it slipped through. A test that calls
`sendHeadersAndGetBody(200, 0)` and then `handleException(...)` would pin it.

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/MetricsHandler.java
```diff
@@ -13,25 +13,40 @@
...
-    prometheusScrapeHandler = new PrometheusScrapeHandler(config);
+    this(new PrometheusScrapeHandler(config), HttpErrorHandlingPolicy.builder().build());
   }
 
   public MetricsHandler(PrometheusProperties config, PrometheusRegistry registry) {
-    prometheusScrapeHandler = new PrometheusScrapeHandler(config, registry);
+    this(new PrometheusScrapeHandler(config, registry), HttpErrorHandlingPolicy.builder().build());
+  }
+
+  MetricsHandler(
```
jaydeluca: This overload is package-private, so only `HTTPServer.Builder` can reach it — but
`MetricsHandler` is `@StableApi` with four public constructors, and its whole reason to exist as
a public `HttpHandler` is registering it on a user-owned `com.sun.net.httpserver.HttpServer`
without going through `HTTPServer`.

For those users this PR is a strict regression rather than a configurable tradeoff: the old
behavior (stack trace to the client) is gone *and* the new behavior (reporter) is unreachable.
They get a bare 500 telling them to "Configure an HTTP error reporter for details" with no API
to do so.

Same applies to `HttpExchangeAdapter`'s package-private constructor
(`HttpExchangeAdapter.java:30`), which is the documented seam for wiring
`PrometheusScrapeHandler` into any `HttpExchange`-based server.

Both are purely additive and would show up cleanly in the api-diff:

```java
public MetricsHandler(
    PrometheusProperties config,
    PrometheusRegistry registry,
    HttpErrorHandlingPolicy errorHandlingPolicy) { ... }

public HttpExchangeAdapter(HttpExchange httpExchange, HttpErrorHandlingPolicy policy) { ... }
```

docs/content/exporters/httpserver.md
```diff
@@ -26,6 +26,41 @@ or [inetAddress()](</client_java/api/io/prometheus/metrics/exporter/httpserver/H
 The default handler can be changed
 with [defaultHandler()](</client_java/api/io/prometheus/metrics/exporter/httpserver/HTTPServer.Builder.html#defaultHandler(com.sun.net.httpserver.HttpHandler)>).
 
+## Scrape error handling
+
+By default, scrape failures return a generic HTTP 500 response. Exception details are not
+included in the response or logged, because the server may run inside an application or a
+Java agent with its own diagnostic pipeline.
```
jaydeluca: "or logged" isn't accurate — the default policy does log, at `SEVERE`, with a full throwable, in
two paths: `HttpExchangeAdapter:120-128` (couldn't write the 500) and `:133-137` (headers already
sent). If we don't take the change in my other comment, the second one is routine rather than
exceptional, which makes this sentence actively misleading for the agent use case it's written
for.

```suggestion
By default, scrape failures return a generic HTTP 500 response. Exception details are not
included in the response, and are not logged when the error response can be delivered, because
the server may run inside an application or a Java agent with its own diagnostic pipeline.
```

(If the response itself can't be delivered, logging is the only remaining signal — worth saying
so explicitly rather than leaving it as a surprise.)

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpErrorHandlingPolicy.java
```diff
@@ -0,0 +1,103 @@
...
+    private Builder() {}
+
+    /**
+     * Pass scrape exceptions to {@code errorReporter}.
+     *
+     * <p>The reporter runs synchronously on the HTTP request thread. It should return promptly and
+     * must be safe to call concurrently. Runtime exceptions thrown by the reporter are isolated
+     * from HTTP response handling.
+     */
+    public Builder errorReporter(Consumer<Throwable> errorReporter) {
```
jaydeluca: Only `Exception` is ever delivered here — `handleException` has exactly two overloads
(`IOException`, `RuntimeException`) and `PrometheusScrapeHandler:99-103` catches exactly those
two. A bare `Throwable` can't reach this consumer.

`Consumer<? super Exception>` would be both more honest about that contract and strictly more
permissive: it accepts `Consumer<Exception>`, `Consumer<Throwable>`, and `Consumer<Object>`,
where today someone holding a `Consumer<Exception>` needs a cast or a wrapper lambda.

Raising it now rather than as a nit because `@StableApi` on line 17 freezes this signature at
merge. Existing-style callers keep compiling (contravariance covers
`Consumer<Throwable> r = ...; policy.errorReporter(r)`, and erasure is identical), so it's
cheap now and impossible later.

Related, same "frozen at merge" reasoning: a named `@FunctionalInterface ScrapeErrorReporter`
instead of `Consumer` would leave room to add a second parameter later (the failing request, a
scrape context). A `Consumer` can never grow one. Not saying we need that but a thought.

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpErrorHandlingPolicy.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>The default policy built by {@link #builder()} does not expose exception details and does not
+ * report the exception. Configure the builder to route diagnostic details to an
+ * application-appropriate sink.
+ */
+@StableApi
+public final class HttpErrorHandlingPolicy {
+
+  private static final byte[] GENERIC_RESPONSE =
+      ("An internal error occurred while scraping metrics. "
+              + "Configure an HTTP error reporter for details.\n")
```
jaydeluca, line 26: The out-of-the-box experience for a plain `HTTPServer.builder().buildAndStart()` in a standalone
app is now: scrape fails → bare 500 → nothing anywhere. This message tells the operator to
configure a reporter, which requires a code change and a redeploy to act on.

The default itself is defensible — agent safety is the harder constraint and it should win. But
the *opt-in ergonomics* mean everyone writes the same five-line lambda. Worth shipping it:

```java
/** Reports scrape exceptions to {@code java.util.logging} at {@code SEVERE}. */
public static Consumer<Throwable> julReporter() { ... }
```

so the override is `.errorReporter(HttpErrorHandlingPolicy.julReporter())`. Makes "we
deliberately don't do this for you" a one-liner to reverse instead of a research task.

prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/HttpExchangeAdapter.java
```diff
@@ -92,52 +97,53 @@ public HttpResponse getResponse() {
...
   public void handleException(RuntimeException e) {
-    sendErrorResponseWithStackTrace(e);
+    sendErrorResponse(e);
   }
 
-  private void sendErrorResponseWithStackTrace(Exception requestHandlerException) {
+  private void sendErrorResponse(Exception requestHandlerException) {
     if (!responseSent) {
       responseSent = true;
+      reportException(requestHandlerException);
```
jaydeluca: The reporter runs before the 500 is written. The javadoc says it "should return promptly," but a
reporter that blocks — synchronous telemetry export, saturated appender — delays the client's
500, and one that hangs prevents it entirely. Moving this call below the `try/catch` guarantees
the response goes out first and costs nothing.

If you do move it, the "Original Exception that caused the Prometheus scrape error" log at
`:125-128` becomes redundant whenever a reporter is configured — it has already seen the
exception by then. Logging only `errorWriterException` there would keep that path quiet for
agent users.

zeitlinger:
Addressed all outstanding review feedback:

- Made the policy-taking `MetricsHandler` and `HttpExchangeAdapter` constructors public without changing existing defaults.
- Changed reporters to `Consumer<? super Exception>` and added explicit synchronous `HttpErrorHandlingPolicy.julReporter()` at JUL `SEVERE`.
- Run configured reporters after the response is written; configured reporters also receive post-response failures without unconditional JUL logging.
- Preserved transport-failure logging and documented the behavior and opt-in JUL reporter.
- Added constructor, ordering, post-response, JUL-level, and policy behavior tests; regenerated the HTTPServer API diff.

Validated with `mise run lint:fix`, `mise run build`, `mise run api-diff`, and the HTTPServer module test suite (26 tests).

jaydeluca (approved):

Issue #1816, opened 2026-01-26
Code quality improvements


zeitlinger:
# Code Quality Improvement Plan

This document tracks code quality improvements for the Prometheus Java Client library. Work through these items incrementally across sessions.

### 1. Add Missing Test Coverage for Exporter Modules
- [x] `prometheus-metrics-exporter-common` - base module, no tests
- [x] `prometheus-metrics-exporter-servlet-jakarta` - no tests
- [x] `prometheus-metrics-exporter-servlet-javax` - no tests
- [x] `prometheus-metrics-exporter-opentelemetry-otel-agent-resources` - no tests

### 2. Eliminate Dropwizard Module Duplication
- [x] Create shared base class or use generics for `prometheus-metrics-instrumentation-dropwizard` and `prometheus-metrics-instrumentation-dropwizard5` (~297 lines each, nearly identical)

### 3. Address Technical Debt (TODOs)
- [ ] `prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java:965` - "reset interval isn't tested yet"
- [ ] `prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Summary.java:205` - "Exemplars (are hard-coded as empty)"
- [ ] `prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/SlidingWindow.java:18` - "synchronized implementation, room for optimization"
- [ ] `prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/PrometheusPropertiesLoader.java:105` - "Add environment variables like EXEMPLARS_ENABLED"
- [ ] `prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java:44` - "filter configuration for OpenTelemetry exporter"
- [ ] `prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/ExporterOpenTelemetryProperties.java:7` - "JavaDoc missing"

### 4. Improve Exception Handling
Replace broad `catch (Exception e)` with specific exception types:
- [ ] `prometheus-metrics-instrumentation-dropwizard5/src/main/java/.../DropwizardExports.java:237`
- [ ] `prometheus-metrics-instrumentation-caffeine/src/main/java/.../CacheMetricsCollector.java:229`
[...]

PR #2344, opened 2026-07-25
feat: support metric name filtering in OpenTelemetry exporter

subhramit:
Closes the last pending high-priority item in https://github.com/prometheus/client_java/issues/1816 (as all others have already been addressed).

Currently, `PrometheusMetricProducer` always exported all metrics in the registry unconditionally. The code already had a TODO sketching how filtering could be added:
https://github.com/prometheus/client_java/blob/9432fdc1933611d34d944d4640ebca3bf91f8ee4/prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java#L47-L54

this PR implements that by reusing `ExporterFilterProperties` and `MetricNameFilter`, as already done by `PrometheusScrapeHandler`.

As a result, the OpenTelemetry exporter now honors the shared `io.prometheus.exporter.filter.*` configuration, making its behavior consistent with the HTTP/Servlet exporters.

### DIFF
--- a/prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java
+++ b/prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java
@@ -35,34 +35,29 @@ class PrometheusMetricProducer implements CollectionRegistration {
   private final boolean preserveNames;
   @Nullable private final Predicate<String> nameFilter;
 
-  /**
-   * Creates a producer with no metric name filtering, i.e. all metrics in {@code registry} are
-   * exported.
-   */
-  public PrometheusMetricProducer(
-      PrometheusRegistry registry,
-      InstrumentationScopeInfo instrumentationScopeInfo,
-      Resource resource,
-      boolean preserveNames) {
-    this(
-        registry,
-        instrumentationScopeInfo,
-        resource,
-        preserveNames,
-        ExporterFilterProperties.builder().build());
-  }
-
-  public PrometheusMetricProducer(
+  private PrometheusMetricProducer(
       PrometheusRegistry registry,
       InstrumentationScopeInfo instrumentationScopeInfo,
       Resource resource,
       boolean preserveNames,
-      ExporterFilterProperties filterProperties) {
+      @Nullable Predicate<String> nameFilter) {
     this.registry = registry;
     this.instrumentationScopeInfo = instrumentationScopeInfo;
     this.resource = resource;
     this.preserveNames = preserveNames;
-    this.nameFilter = makeNameFilter(filterProperties);
+    this.nameFilter = nameFilter;
+  }
+
+  /**
+   * Creates a builder for a producer with no metric name filtering by default, i.e. all metrics in
+   * {@code registry} are exported unless filter properties are configured on the builder.
+   */
+  static Builder builder(
+      PrometheusRegistry registry,
+      InstrumentationScopeInfo instrumentationScopeInfo,
+      Resource resource,
+      boolean preserveNames) {
+    return new Builder(registry, instrumentationScopeInfo, resource, preserveNames);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


subhramit:
cc @zeitlinger :)

prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java
```diff
@@ -30,29 +33,70 @@ class PrometheusMetricProducer implements CollectionRegistration {
...
       boolean preserveNames) {
+    this(
+        registry,
+        instrumentationScopeInfo,
+        resource,
+        preserveNames,
+        ExporterFilterProperties.builder().build());
+  }
+
+  public PrometheusMetricProducer(
```
zeitlinger: We avoid adding new public constructors because constructors are difficult to evolve compatibly. Please expose this through a builder instead and keep the constructor non-public. We should also document this API-design rule in `AGENTS.md` or `CONTRIBUTING.md` so future changes follow it consistently.

zeitlinger (changes requested):
The filtering implementation is consistent with `PrometheusScrapeHandler`: it builds the same `MetricNameFilter` from all four `ExporterFilterProperties` fields and uses `registry.scrape(predicate)`. I ran the exporter module test reactor locally and it passed (including all 35 exporter-module tests). The two new tests cover exact-name exclusion and allowed-prefix filtering; please also add coverage through the user-facing configuration path (`OtelAutoConfig`) rather than only constructing `PrometheusMetricProducer` directly, so the wiring itself is protected against regression.

prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java
subhramit: Would you wish for this documentation change to be a part of this PR's scope, or a separate one (perhaps, as a follow-up)?

prometheus-metrics-exporter-opentelemetry/src/test/java/io/prometheus/metrics/exporter/opentelemetry/ExportTest.java
```diff
@@ -381,6 +382,46 @@ void preserveNamesWithoutUnit() {
...
+            Resource.create(Attributes.builder().put("staticRes", "value").build()),
+            false,
+            ExporterFilterProperties.builder().allowedPrefixes("http_").build()));
+
+    Counter.builder().name("http_requests").register(filteredRegistry).inc();
+    Counter.builder().name("jvm_threads").register(filteredRegistry).inc();
+
+    List<MetricData> metrics = new ArrayList<>(reader.collectAllMetrics());
+    assertThat(metrics).hasSize(1);
+    OpenTelemetryAssertions.assertThat(metrics.getFirst()).hasName("http_requests");
```
subhramit: `getFirst` isnt available in java 17, hence the CI test failure.
Will fix this in the next commit as well, when addressing your comment on the api design..

prometheus-metrics-exporter-opentelemetry/src/main/java/io/prometheus/metrics/exporter/opentelemetry/PrometheusMetricProducer.java
subhramit: Exposed now through builder.

subhramit:
> please also add coverage through the user-facing configuration path (`OtelAutoConfig`) rather than only constructing `PrometheusMetricProducer` directly, so the wiring itself is protected against regression.

This is also added now.

zeitlinger (commented):
Could you add a short note to `docs/content/otel/otlp.md` that the OpenTelemetry exporter now honors the shared `io.prometheus.exporter.filter.*` metric-name filter properties, with a link to the existing “Exporter Filter Properties” section in `docs/content/config/config.md`? The property keys are already documented there, so a separate page is not needed.

subhramit:
> Could you add a short note to `docs/content/otel/otlp.md` that the OpenTelemetry exporter now honors the shared `io.prometheus.exporter.filter.*` metric-name filter properties, with a link to the existing “Exporter Filter Properties” section in `docs/content/config/config.md`? The property keys are already documented there, so a separate page is not needed.

Added.

zeitlinger (approved):
Thanks for addressing the builder API, end-to-end wiring test, and documentation. The filtering behavior now matches the existing scrape-handler path, and CI is green.

zeitlinger:
Thanks for the contribution!

PR #2415, opened 2026-08-24
fix(exposition): export internal package for OSGi resolution

### LINKED ISSUE #2395: expositionformats.internal is imported by exposition-textformats but exported by nothing, breaking OSGi resolution
## Summary

In 1.8.0, `prometheus-metrics-exposition-textformats` imports `io.prometheus.metrics.expositionformats.internal`, but no artifact exports it. The classes exist in `prometheus-metrics-exposition-formats`, which keeps them unexported, so the two bundles cannot be wired together and any OSGi container fails to resolve `io.prometheus.metrics.expositionformats`.

## Evidence

From the published 1.8.0 manifests:

```
prometheus-metrics-exposition-textformats-1.8.0.jar
  Export-Package: io.prometheus.metrics.expositionformats
  Import-Package: io.prometheus.metrics.expositionformats.internal      <- nothing provides this

prometheus-metrics-exposition-formats-1.8.0.jar
  Export-Package: io.prometheus.metrics.expositionformats.generated
                  io.prometheus.metrics.expositionformats.generated.com_google_protobuf_4_35_0
  Import-Package: io.prometheus.metrics.expositionformats
```

The implementation classes are in the second jar but not exported from it:

```
$ unzip -l prometheus-metrics-exposition-formats-1.8.0.jar | grep expositionformats/internal
  io/prometheus/metrics/expositionformats/internal/PrometheusProtobufWriterImpl.class
  io/prometheus/metrics/expositionformats/internal/ProtobufUtil.class
```

`prometheus-metrics-exposition-textformats-1.8.0.jar` contains none of them, so the import is not satisfiable from within its own bundle either.

## How I hit it

Micrometer has an OSGi test that is still pinned to the old `simpleclient` because of an earlier failure (micrometer-metrics/micrometer#5063). Switching it to the current client, the bnd resolution now stops here:

```
micrometer-osgi-test-tests cannot be resolved
  ⇒ Import-Package io.micrometer.prometheusmetrics could not be provided
    ⇒ Import-Package io.prometheus.metrics.expositionformats;version=[1.8.0,2.0.0) could not be provided
      ⇒ Import-Package io.prometheus.metrics.expositionformats.internal could not be provided
```

The blocker reported back in 2024 (the shaded protobuf module not being a bundle) is gone, this is what is left.

## Possible fixes

Either export `io.prometheus.metrics.expositionformats.internal` from `prometheus-metrics-exposition-formats`, or move the two classes so the package does not have to cross a bundle boundary. The first is a one line manifest change but does publish an internal package; the second is cleaner if you would rather not.

I saw #989 was closed as not planned, though that was about `prometheus-metrics-shaded-protobuf`, which no longer exists in 1.8.0. This one is a metadata inconsistency between two shipped artifacts rather than a request for OSGi support, so I thought it was worth reporting separately. Happy to send a PR for whichever direction you prefer.

## Environment

- `io.prometheus:*` 1.8.0
- bnd resolution via Micrometer's `micrometer-osgi-test` (Felix framework)

arimu1:
Fixes #2395

## Summary

- Export `io.prometheus.metrics.expositionformats.internal` from both `prometheus-metrics-exposition-formats` and `prometheus-metrics-exposition-formats-no-protobuf` OSGi bundles
- `prometheus-metrics-exposition-textformats` imports that package because `PrometheusProtobufWriter` loads `PrometheusProtobufWriterImpl` via `Class.forName`, but bnd kept the implementation package private in the formats bundles, so OSGi resolution failed

## Approach

Used bnd `_exportcontents` to export the otherwise-private internal package while preserving existing `generated*` exports. This matches the metadata wiring fix suggested in the issue (option 1) without moving implementation classes across bundle boundaries.

## Prove-it (before fix on HEAD)

```
prometheus-metrics-exposition-textformats
  Import-Package: ... io.prometheus.metrics.expositionformats.internal ...

prometheus-metrics-exposition-formats
  Export-Package: io.prometheus.metrics.expositionformats.generated*, ...
  (internal not exported)
```

## After fix

Both formats bundles now export `io.prometheus.metrics.expositionformats.internal`, satisfying the textformats import.

## Test plan

- [x] `mvn -pl prometheus-metrics-exposition-formats,prometheus-metrics-exposition-formats-shaded,prometheus-metrics-exposition-textformats -am test` (Java 25)
- [x] Inspected `META-INF/MANIFEST.MF` in built JARs to confirm internal package is exported and generated exports remain

### DIFF
--- a/prometheus-metrics-exposition-formats-shaded/pom.xml
+++ b/prometheus-metrics-exposition-formats-shaded/pom.xml
@@ -62,6 +62,21 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
+            <Export-Package>
+              io.prometheus.metrics.expositionformats.generated*;version="${project.version}"
+            </Export-Package>
+            <_exportcontents>
+              io.prometheus.metrics.expositionformats.internal;version="${project.version}"
+            </_exportcontents>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-resources-plugin</artifactId>
         <executions>
--- a/prometheus-metrics-exposition-formats-shaded/pom.xml
+++ b/prometheus-metrics-exposition-formats-shaded/pom.xml
@@ -66,15 +66,26 @@
         <artifactId>maven-bundle-plugin</artifactId>
         <configuration>
           <instructions>
-            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
             <Export-Package>
-              io.prometheus.metrics.expositionformats.generated*;version="${project.version}"
+              io.prometheus.metrics.expositionformats.generated*
             </Export-Package>
+            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
             <_exportcontents>
-              io.prometheus.metrics.expositionformats.internal;version="${project.version}"
+              io.prometheus.metrics.expositionformats.internal
             </_exportcontents>
+            <Import-Package>
+              !com.google.protobuf,*
+            </Import-Package>
           </instructions>
         </configuration>
+        <executions>
+          <execution>
+            <!-- Write MANIFEST.MF in process-classes for OSGi header tests. -->
+            <goals>
+              <goal>manifest</goal>
+            </goals>
+          </execution>
+        </executions>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/prometheus-metrics-exposition-formats/pom.xml
+++ b/prometheus-metrics-exposition-formats/pom.xml
@@ -62,6 +62,21 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
+            <Export-Package>
+              io.prometheus.metrics.expositionformats.generated*;version="${project.version}"
+            </Export-Package>
+            <_exportcontents>
+              io.prometheus.metrics.expositionformats.internal;version="${project.version}"
+            </_exportcontents>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>build-helper-maven-plugin</artifactId>
         <executions>
--- a/prometheus-metrics-exposition-formats/pom.xml
+++ b/prometheus-metrics-exposition-formats/pom.xml
@@ -66,15 +66,23 @@
         <artifactId>maven-bundle-plugin</artifactId>
         <configuration>
           <instructions>
-            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
             <Export-Package>
-              io.prometheus.metrics.expositionformats.generated*;version="${project.version}"
+              io.prometheus.metrics.expositionformats.generated*
             </Export-Package>
+            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
             <_exportcontents>
-              io.prometheus.metrics.expositionformats.internal;version="${project.version}"
+              io.prometheus.metrics.expositionformats.internal
             </_exportcontents>
           </instructions>
         </configuration>
+        <executions>
+          <execution>
+            <!-- Write MANIFEST.MF in process-classes for OSGi header tests. -->
+            <goals>
+              <goal>manifest</goal>
+            </goals>
+          </execution>
+        </executions>
       </plugin>
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
--- a/prometheus-metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/OsgiBundleManifestTest.java
+++ b/prometheus-metrics-exposition-formats/src/test/java/io/prometheus/metrics/expositionformats/OsgiBundleManifestTest.java
@@ -18,8 +18,7 @@ import org.junit.jupiter.api.Test;
 
 class OsgiBundleManifestTest {
 
-  private static final String INTERNAL_PACKAGE =
-      "io.prometheus.metrics.expositionformats.internal";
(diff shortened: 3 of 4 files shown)


prometheus-metrics-exposition-formats-shaded/pom.xml
```diff
@@ -61,6 +61,21 @@
...
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
+            <Export-Package>
+              io.prometheus.metrics.expositionformats.generated*;version="${project.version}"
+            </Export-Package>
+            <_exportcontents>
+              io.prometheus.metrics.expositionformats.internal;version="${project.version}"
+            </_exportcontents>
```
jaydeluca, line 75: `${project.version}` is `1.8.1-SNAPSHOT` on snapshot builds, which isn't valid OSGi version
syntax, so bnd rewrites it to `1.8.1.SNAPSHOT`. That sorts *below* a released `1.8.1`. Built both
branches to confirm the delta:


```text
main:   io.prometheus.metrics.expositionformats.generated;version="1.8.1"
branch: io.prometheus.metrics.expositionformats.generated;version="1.8.1.SNAPSHOT"
```

bnd already defaults exported-package versions to the cleaned project version, so on a release
build the attribute produces exactly what you'd get without it — it only changes behavior on
snapshots, and only for the worse. Dropping it restores `1.8.1` (verified locally).

```suggestion
            <Export-Package>
              io.prometheus.metrics.expositionformats.generated*
            </Export-Package>
            <_exportcontents>
              io.prometheus.metrics.expositionformats.internal
            </_exportcontents>
```

Same applies to `prometheus-metrics-exposition-formats-shaded/pom.xml`

prometheus-metrics-exposition-formats-shaded/pom.xml
```diff
@@ -61,6 +61,21 @@
...
+          </instructions>
```
jaydeluca, line 79: While verifying this fix I hit a second manifest problem in this bundle that's pre-existing, but
it matters for whether #2395 is actually resolved.

`maven-bundle-plugin` runs before `maven-shade-plugin`, so bnd computes the manifest against
*unrelocated* classes. On both `main` and this branch the shaded bundle declares:

```text
Import-Package: com.google.protobuf;version="[4.36,5)"   ← mandatory
```

…while the jar contains **zero** `com/google/protobuf/` classes. I decompiled
`internal/ProtobufUtil.class` out of the built shaded jar and it references
`io/prometheus/metrics/shaded/com_google_protobuf_4_36_0/*`, as you'd expect. So the shaded
artifact demands a real `protobuf-java` bundle be installed, which rather defeats the point of
shading.

**Why this blocks the original report**: `micrometer-registry-prometheus` depends on
`io.prometheus:prometheus-metrics-exposition-formats` — the shaded artifact — at runtime scope. So
the bnd resolve in #2395 will get past `internal` after this fix and then stop at
`com.google.protobuf` unless `protobuf-java` happens to be in their resolution repository.

Second-order effect from this PR specifically: the newly exported `internal` package inherits
`uses:="com.google.protobuf,..."`, propagating the bogus constraint to every consumer that wires
to it.

Adding `!com.google.protobuf` removes both the phantom import and the bad `uses` clause. Verified
against a local build:

```suggestion
            <Import-Package>!com.google.protobuf,*</Import-Package>
          </instructions>
```

Resulting manifest:

```text
Export-Package: ...internal;version="1.8.1";uses:="io.prometheus.metrics.config,..."
                                              ↑ no com.google.protobuf
Import-Package: io.prometheus.metrics.config;version="[1.8,2)",...
                ↑ com.google.protobuf gone
```

with all that being said, are there some tests we can include to verify that all of this actually works and fixes the issue? I have limited experience with OSGi, so I defer to you and @anjeongkyun to hopefully weigh in

prometheus-metrics-exposition-formats/pom.xml
```diff
@@ -61,6 +61,21 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <!-- Required for OSGi: textformats loads protobuf impl via Class.forName -->
```
jaydeluca: This comment is the right diagnosis, and it points at the half of the problem that this PR doesn't fix.

`createProtobufWriter()` is explicitly written to tolerate the impl being absent — it catches, returns `null`, and `isAvailable()` reports false (`prometheus-metrics-exposition-textformats/.../PrometheusProtobufWriter.java:26-37`). But bnd emits the import as **mandatory**:

```text
Import-Package: ... io.prometheus.metrics.expositionformats.internal   ← no resolution:=optional
```

So a deployment that installs only `prometheus-metrics-exposition-textformats` — which is the
entire reason that artifact ships separately from the protobuf-carrying ones — still fails to
resolve after this change. Exporting `internal` fixes the case where the formats bundle *is*
present; it doesn't fix the case where it deliberately isn't.

One line in `prometheus-metrics-exposition-textformats/pom.xml`, in a `maven-bundle-plugin`
`<instructions>` block mirroring the ones you've added here:

```xml
<Import-Package>io.prometheus.metrics.expositionformats.internal;resolution:=optional,*</Import-Package>
```

I built this and confirmed the manifest:

```text
Import-Package: io.prometheus.metrics.expositionformats.internal;resolution:=optional,
                io.prometheus.metrics.config;version="[1.8,2)",...
```

jaydeluca: also, the comment sits above `<Export-Package>` but explains `<_exportcontents>`

jaydeluca (commented):
I'm not super familiar with OSGI, so I would like to see some tests or other evidence that this solves the issue if possible

arimu1:
Added `OsgiBundleManifestTest` (runs in both exposition-formats bundles) that reads the bnd-generated `MANIFEST.MF` and asserts:

- formats `Export-Package` includes `io.prometheus.metrics.expositionformats.internal`
- textformats `Import-Package` includes the same package

Local Temurin 25: `OsgiBundleManifestTest` 2/2 in both formats modules; exposition tests 107 (textformats) + 68 (formats) + 68 (shaded).

zeitlinger (approved):
Reviewed latest head: CI is green, no merge conflicts, and no blocking review comments remain.

zeitlinger:
Thanks for the contribution!

zeitlinger:
@arimu1 please fix the format error

arimu1:
Fixed the lint failure: `google-java-format` 1.36.1 wanted `INTERNAL_PACKAGE` on one line in `OsgiBundleManifestTest`. `flint`/GJF dry-run is clean locally.

jaydeluca (approved):

PR #2441, opened 2026-09-01
perf: skip snapshot rebuild in mergeDuplicates when names are unique

david-mollitor-db:
## What

`TextFormatUtil.mergeDuplicates` currently only short-circuits when there is a single snapshot. For every scrape with two or more metric families it unconditionally builds a `LinkedHashMap`, an `ArrayList` per group, a `MetricSnapshots.Builder` (with its own list + set) and a freshly sorted `MetricSnapshots` — even when there is nothing to merge, which is the common case.

Since `MetricSnapshots` is always sorted by prometheus name (see its constructor), any duplicate names are **adjacent**. This adds a single allocation-free pass to detect duplicates; when there are none, the input is returned unchanged.

```java
boolean hasDuplicates = false;
for (int i = 1; i < metricSnapshots.size(); i++) {
  if (metricSnapshots.get(i).getMetadata().getPrometheusName()
      .equals(metricSnapshots.get(i - 1).getMetadata().getPrometheusName())) {
    hasDuplicates = true;
    break;
  }
}
if (!hasDuplicates) {
  return metricSnapshots;
}
```

The merge path for actual duplicates is unchanged.

## Why

It removes the map / per-group list / builder / re-sort allocations on every scrape that has no duplicate metric names. Measured on a JMH benchmark of a histogram-heavy scrape (collect + serialize to the Prometheus text format), allocation on the serialize path dropped by ~880 B/op, scaling with the number of metric families.

## Correctness

Output is byte-identical. The existing exposition-format tests pass, including `DuplicateNamesExpositionTest`, which exercises the merge path with real duplicate names.


prometheus-metrics-exposition-textformats/src/main/java/io/prometheus/metrics/expositionformats/TextFormatUtil.java
```diff
@@ -41,6 +41,24 @@ public static MetricSnapshots mergeDuplicates(MetricSnapshots metricSnapshots) {
       return metricSnapshots;
     }
 
+    // MetricSnapshots is sorted by prometheus name, so any duplicates are adjacent. Detect them in
```
jaydeluca, line 44: overall this change and associated benchmarks look good. I'd like to see a few extra tests added though.

All four of the existing mergeDuplicates tests either exercise the merge path or pass any empty snapshots list. An inverted condition or `i=0` start would not be caught by the existing tests. Can we add cases for:

* unique names → `assertThat(TextFormatUtil.mergeDuplicates(s)).isSameAs(s)`
* a mixed case where the duplicate pair is not at index 0/1 (e.g. `a`, `m`, `z`, `z`) → still merged, which pins the loop range.

Also, the sorted invariant is cross-module and nothing pins the sort key, and the fast path is only correct while `MetricSnapshots` sorts by `getPrometheusName()`. 

`MetricSnapshotsTest.testSort` asserts order via `getMetadata().getName()` using names where `name`, `prometheusName` and `expositionBaseName` all coincide so it would keep passing if the sort key moved to `name` or `expositionBaseName`. If that happens, `mergeDuplicates` silently stops merging and the writers emit repeated `# HELP`/`# TYPE` blocks for one family.

we could add some tests around this:

* a test in `MetricSnapshotsTest` that pins ordering by `getPrometheusName()` with names where the three differ (e.g. `my.metric` vs `my_metric_b`, or a `_total`-suffixed counter);
* a comment on the sort in `MetricSnapshots` naming `TextFormatUtil.mergeDuplicates` as a
     dependant, so the next person to touch the comparator sees it.

zeitlinger: Thanks Jay — all four requests are now addressed:

* `testMergeDuplicates_uniqueNames_returnsSameInstance` covers the unique-name fast path.
* `testMergeDuplicates_duplicateNotAtStart_merges` covers a duplicate pair late in the list.
* `testSortIsByPrometheusName` pins ordering with `foo.bar`/`fooZbar`, whose Prometheus names sort differently from their original names.
* The `MetricSnapshots` comparator comment now names `TextFormatUtil.mergeDuplicates` as a dependent.

The current Build and Test on Java 17 failures are the unrelated `HistogramTest.testObserveMultithreaded` in-flight-observation timeout. I plan to update and retest this PR once #2471 lands, rather than treating those failures as evidence against this optimization. Please re-review when convenient; I’ll leave this thread open for now.

zeitlinger: I have now retested this branch with the late-observation fix from #2471 (commit d8a34a1). The previously failing HistogramTest timeout passes, and the focused Buffer/Histogram/Summary tests plus the full build (with Docker integration tests skipped locally) pass. I pushed the additive fix to the PR branch; CI is rerunning.

zeitlinger (approved):

PR #2442, opened 2026-09-01
perf: avoid allocating a mapping lambda on every labelValues() call

david-mollitor-db:
## What

`StatefulMetric.labelValues(...)` — the path every `counter.labelValues(...).inc()` /
`histogram.labelValues(...).observe(...)` goes through — called
`data.computeIfAbsent(key, l -> newDataPoint())` on every invocation. The mapping function
captures `this` (via `metadata`/`labelNames`), so it is **not** a cached singleton: a new lambda
instance is allocated on every call, including the overwhelmingly common case where the data point
already exists (the lambda argument is constructed before `computeIfAbsent` runs, even though it is
only invoked on a miss).

This adds a `data.get(key)` fast path that returns the existing data point without constructing the
lambda:

```java
List<String> key = Arrays.asList(labelValues);
T dataPoint = data.get(key);
if (dataPoint != null) {
  return dataPoint;
}
// miss: validate the raw array outside the CHM lock, then create at-most-once
for (int i = 0; i < labelValues.length; i++) {
  if (labelValues[i] == null) {
    throw new IllegalArgumentException(
        "null label value for metric " + metadata.getName() + " and label " + labelNames[i]);
  }
}
return data.computeIfAbsent(key, l -> newDataPoint());
```

Two secondary points folded in on the miss path:
- the null-value validation now runs **before** `computeIfAbsent`, i.e. outside the
  `ConcurrentHashMap` bin lock, and operates on the raw `String[]` (no `List.get` indirection);
- creation stays inside `computeIfAbsent` on purpose — `newDataPoint()` has a side effect (a native
  histogram may schedule a reset task via `Scheduler.schedule`), so at-most-once creation must be
  preserved; `putIfAbsent` with a pre-built value would leak the loser's scheduled task on a race.

## Why

`labelValues(...)` is the hottest path in the library (every metric update). In a JMH benchmark of a
histogram-heavy workload, the `get()` fast path cut record-path allocation by ~18% — exactly the
16-byte captured lambda per observation.

## Correctness

Behavior is unchanged (a null label value still throws on first use, since a null-containing key is
never inserted and so always reaches the miss branch). Verified by the core tests, including
`StatefulMetricTest`.

This pull request and its description were written by Isaac.


zeitlinger:
Pending the benchmark-only PR [#2468](https://github.com/prometheus/client_java/pull/2468)
before making the release/merge decision on this optimization.

That PR adds repeated existing-label lookup/increment benchmarks with cached-data-point baselines,
single-threaded and contended variants, and B/op reporting. The existing counter benchmarks cache
their data points and therefore do not directly measure the claimed lookup allocation savings.

Once #2468 is merged, update this branch and retrigger the benchmark label so both base and head
contain the same lookup benchmark code. Then evaluate throughput and allocation for this change.
The benchmark PR itself initially has head-only results for the new methods and cannot establish
this optimization's benefit.

zeitlinger:
I ran a matched local JMH comparison using the identical #2468 lookup/cached harness on both candidate base and head, with #2471 applied to both. Configuration: JDK 25.0.3, 3 forks, 3x10s warmup, 5x10s measurement, GC profiler. Results (base -> head): lookup single-thread 62.85M -> 88.12M ops/s (+40.2%), 64 -> 48 B/op (-25%); lookup 4-thread 225.36M -> 295.40M ops/s (+31.1%), 64 -> 48 B/op (-25%). Cached controls: 520.25M -> 534.07M (+2.7%) single-thread and 1.504B -> 1.513B (+0.6%) 4-thread, both ~0 B/op. This supports #2442 as a worthwhile release optimization. Full raw JSON is available locally during review.

zeitlinger (approved):

zeitlinger:
thanks for the contribution

PR #2468, opened 2026-09-15
test: focus PR benchmarks on client_java and measure label lookups

zeitlinger:
## Summary

Benchmark-only follow-up for the release review; no production metric changes.

- Restrict both base and head PR benchmark runs to client_java counter/histogram methods.
  OpenTelemetry, Codahale, and legacy simpleclient comparisons remain in full/local/nightly runs.
  Keep all client_java exposition benchmarks, including OpenMetrics.
- Add repeated existing-label lookup + increment benchmarks and cached-data-point baselines,
  with one-thread and four-thread variants. One invocation is one metric update.
- Report GC profiler allocation in B/op separately from throughput, including descriptive
  base/head allocation deltas only when configurations match.
- Add selection/report regression tests and run benchmark tooling tests in lint CI.
- Document operation units and the new benchmark workflow.

## Related PR

[Label lookup optimization #2442](https://github.com/prometheus/client_java/pull/2442) is pending this
benchmark infrastructure. After this lands, update that branch and rerun the benchmarks so base and
head contain identical lookup benchmark code. The new methods in this PR itself have head-only
results and do not establish the optimization's benefit.

## Validation

- `mise run lint:fix` — passed; formatter changes retained.
- `mise run test` — passed.
- All three benchmark tooling test scripts — 24 tests passed.
- `mise run build -- -DskipITs=true` — passed.
- Plain `mise run build` compiled the benchmark module, but Docker-backed integration tests ran
  despite `-DskipTests` and failed because no Docker environment was available.
- JMH listing with each workflow pattern confirmed client_java-only selection and retained
  OpenMetrics/Prometheus exposition cases.
- All four new benchmarks completed with GC profiling: 1 fork, 2 x 1s warmups, 3 x 1s measurements,
  JDK 25.0.3, `-Xms128m -Xmx256m`. This was a smoke test on a shared development host, not a
  controlled base/head performance comparison. Repeated lookup reported about 64 B/update;
  cached increments were near zero.
- Generated the Markdown report from real smoke-run JSON and checked allocation output.
- `git diff --check` — passed.


zeitlinger:
The histogram timeout was investigated separately from this benchmark work.
[#2471](https://github.com/prometheus/client_java/pull/2471) reproduces and fixes a Buffer
generation-handoff race capable of producing the same timeout.

The forced interleaving fails on the old implementation after five seconds, then passes with the
fix. Both regression cases plus the original multithreaded histogram test passed 450 local
executions across 1, 2, and 4 configured processors; full unit tests and core coverage checks also
passed. This proves the race, though not the exact interleaving of this historical CI failure.

This PR remains benchmark-only. Once the correctness fix lands, update/retest this branch rather
than increasing the timeout or weakening the test.

zeitlinger:
I rebased this benchmark PR additively with the focused buffer handoff fix from #2471 so its full CI can validate on the same head. Relevant core regression tests pass locally, lint and `mise run build -- -DskipITs=true` pass. No benchmark scope changes.

jaydeluca (approved):

PR #2471, opened 2026-09-15
fix: keep late observations out of subsequent collection buffers

zeitlinger:
## Problem and proof

A late appender can increment its stripe during collection A, then read
`activeGeneration` only after A has completed and collection B has started.

B already includes that observation in `expectedCount`. If the appender puts it into B's
buffer, B waits for an observation that will only be replayed after B's wait ends.
This deterministically causes the five-second collection timeout.

The old late-appender test paused after reading the generation, missing this window.
Moving the existing injected pause hook before that read, without the fix, reproduced:

```text
BufferTest.lateAppenderCountedByNextGenerationMustNotBeBufferedAgain
Time elapsed: 5.037 s
ExecutionException: IllegalStateException:
Timed out while waiting for in-flight observations.
```

The production hook remains a no-op; the test forces a scheduling window that real threads can hit.

This has the same failure signature as the histogram failures observed in
[#2468](https://github.com/prometheus/client_java/pull/2468) and
[#2441](https://github.com/prometheus/client_java/pull/2441).
The reproduction proves this race exists, not which interleaving occurred in any particular CI run.

## Fix

- Record each stripe's absolute observation count at generation activation, under `appendLock`.
- Under that same lock, reject buffering an observation whose ticket is already included in the
  current generation's starting count. It completes through the existing direct observation path.
- Reuse a single cutoff array per Buffer: one additional long per stripe, plus array overhead,
  with no new per-collection allocation or additional fast-path lock.
- Keep cutoffs independent of the reset-adjusted aggregate observation count.
- Do not change the five-second timeout, buffer capacity, or public API.

## Regression coverage

Strengthen the late-appender test to force the previously untested window and capture worker
exceptions through futures, with bounded waits and executor cleanup.

Cover both normal collection and a preceding reset. Each case also verifies that a genuinely new
observation buffers normally, stays out of the current snapshot, replays exactly once, and appears
in the following collection.

## Validation

- Before fix: deterministic timeout reproduced as above.
- After fix: `BufferTest`, `HistogramTest`, and `SummaryTest` — 50 tests passed.
- Temporary local JUnit launcher: both regression cases and the original
  `HistogramTest.testObserveMultithreaded`, repeated 50 times with each of
  `-XX:ActiveProcessorCount=1`, `2`, and `4` — **450 successful test executions** on JDK 25.
- `mise run lint:fix` — passed; formatter changes retained.
- `mise run build -- -DskipITs=true` — passed. Docker-backed integration tests skipped because
  Docker is unavailable locally.
- `mise run test` — full unit-test suite passed.
- `./mvnw verify -pl prometheus-metrics-core -am` — passed, including coverage checks.
- `git diff --check` — passed.

Kept separate from the benchmark-only PR.

## Current scope after update onto main

The original production fix is already present on `main` through [#2441](https://github.com/prometheus/client_java/pull/2441) and [#2468](https://github.com/prometheus/client_java/pull/2468). This PR now contains the remaining merge-update and review hardening: explicit coverage for both generation handoff windows, fail-closed absolute cutoffs, clearer test seams, and non-masking executor cleanup assertions. The historical race reproduction and validation context above are retained.


prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Buffer.java
```diff
@@ -179,8 +191,10 @@ <T extends DataPointSnapshot> T run(
       try {
         activeGeneration = generation;
         long total = 0;
-        for (AtomicLong counter : stripedObservationCounts) {
-          total += counter.getAndAdd(BUFFER_ACTIVE_BIT);
+        for (int i = 0; i < stripedObservationCounts.length; i++) {
+          long count = stripedObservationCounts[i].getAndAdd(BUFFER_ACTIVE_BIT);
+          generationStartCounts[i] = count;
```
jaydeluca: Consider masking the active bit here, so the cutoff is unconditionally the absolute count:

```suggestion
          generationStartCounts[i] = count & ~BUFFER_ACTIVE_BIT;
```

The comparison at `:112` masks the appender's ticket (`count & ~BUFFER_ACTIVE_BIT`) but compares it
against an unmasked cutoff. That's correct today — activation and deactivation each add
`BUFFER_ACTIVE_BIT` exactly once per `run()` and `runLock` serializes runs, so the bit is provably
clear at activation — but the guard's correctness then rests on a *global* bit-balance argument
rather than on anything local to these two lines.

It also fails in the unsafe direction if that argument ever breaks. `BUFFER_ACTIVE_BIT` is
`1L << 63` = `Long.MIN_VALUE`, so a leaked bit makes the cutoff **negative**, `masked <= negative`
is always false, and the guard would silently accept every late observation — reintroducing
exactly this bug. Masking makes it fail closed (route to `observeDirect`) instead.

`total += count` below should stay unmasked, so this is a one-line change with no behaviour
difference today.

jaydeluca (commented):
Comments are all non-blocking.

prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Buffer.java
```diff
@@ -76,33 +80,41 @@ private static final class Generation {
...
     // The active bit is the exact handoff decision. An observation either increments its stripe
     // before the collector's getAndAdd(BUFFER_ACTIVE_BIT) and takes the direct path, or sees the
-    // active bit and is buffered in the current generation.
+    // active bit and may be buffered. The stripe ticket below also checks that it was not counted
+    // by a later collection that started before this thread read activeGeneration.
     if ((count & BUFFER_ACTIVE_BIT) == 0) {
       return false;
     }
+    // Allow tests to pause between allocating an observation ticket and reading the generation.
+    beforeAppendLock.run();
```
jaydeluca: Now that the hook fires above the `activeGeneration` read, the name `beforeAppendLock` describes
the wrong seam — the new comment on `:100` is accurate but contradicts it. `beforeGenerationRead`
would match. Worth renaming (field at `:65`, constructor param at `:75`) since this field is the
only documentation of where the seam actually is.

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/BufferTest.java
```diff
@@ -180,83 +183,123 @@ void interruptedAppenderLeavesBoundedBufferWait() throws InterruptedException {
...
-              stalled.countDown();
-              try {
-                release.await();
-              } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
+              if (pauseFirstAppender.compareAndSet(true, false)) {
+                observationCounted.countDown();
+                awaitLatch(readGeneration);
               }
             });
```
jaydeluca: Moving the single hook swapped which window is deterministically testable. The old test paused
after the generation read, which pinned `current != generation` at `Buffer.java:108-111`; nothing
pins it now.

I checked whether that branch still runs at all by replacing its body with a throwing assertion
and running the whole module: it is still reached, but only by
`HistogramTest.testObserveMultithreaded` hitting it by chance. So it's covered incidentally rather
than uncovered — which is the kind of coverage that quietly evaporates when an unrelated test
changes.

Two hook points, or one hook that takes the call site as an argument, would pin both windows
deterministically and let this test keep the old one's guarantee as well as the new one.

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/BufferTest.java
```diff
@@ -180,83 +183,123 @@ void interruptedAppenderLeavesBoundedBufferWait() throws InterruptedException {
...
+                          new CounterSnapshot.CounterDataPointSnapshot(
+                              completedObservations.get(), Labels.EMPTY, null, 0),
+                      ignored -> completedObservations.incrementAndGet())
+                  .getValue())
+          .isEqualTo(2);
+    } finally {
+      finishFirstSnapshot.countDown();
+      readGeneration.countDown();
+      executor.shutdownNow();
+      assertThat(executor.awaitTermination(10, TimeUnit.SECONDS)).isTrue();
```
jaydeluca: An assertion in `finally` will replace the original exception if the body fails *and* shutdown is
slow, so a real regression could surface as a misleading "executor did not terminate". Consider
moving the assertion after the body, or capturing it without failing the test:

```suggestion
      boolean terminated = executor.awaitTermination(10, TimeUnit.SECONDS);
      assertThat(terminated).as("executor terminated").isTrue();
```

(That keeps the check but reads more clearly about what failed; moving it out of `finally`
entirely would be better still.)

prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Buffer.java
zeitlinger: Applied. generationStartCounts now stores count & ~BUFFER_ACTIVE_BIT; total remains based on the unmasked count, so behavior is unchanged while the cutoff is fail-closed if the bit balance ever regresses.

prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Buffer.java
zeitlinger: Applied. Renamed the seam field and constructor parameter to beforeGenerationRead; the additional seam is named afterGenerationRead.

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/BufferTest.java
zeitlinger: Applied. Added the afterGenerationRead seam and lateAppenderAfterGenerationReadMustNotBeBufferedAgain, which deterministically pauses after reading generation A and verifies the stale-identity path. As a mutation proof, I temporarily changed the current != generation branch to throw; the test failed immediately with that exact assertion, then the source was restored and the test passed.

prometheus-metrics-core/src/test/java/io/prometheus/metrics/core/metrics/BufferTest.java
zeitlinger: Applied. The executor is still released and shut down in finally, but termination is now awaited and asserted only after the successful test body, so cleanup cannot mask the original failure. The assertion is labeled executor terminated.

jaydeluca (approved):

zeitlinger:
The reviewed hook plumbing caused a measurable JIT regression: `Buffer.append` grew from 321 bytes and inlined (hot) on the original fix to 330 bytes and failed to inline (hot) after the review changes.

I split the ticket/active-bit hot path from the generation/lock path. The resulting `Buffer.append` is 52 bytes and inlines hot; the cold helper is 287 bytes. All hooks and generation guards remain unchanged. This adds no locks or allocations.

A same-host paired JMH run (JDK 25.0.3, 3 forks, 3x3s warmup, 5x3s measurement, classic histogram methods) improved the reviewed head by +11.3% per-thread and +15.0% single-thread (34,074 vs 30,626 ops/s; 9,278 vs 8,063 ops/s). This is a smaller local sample than CI, so a fresh CI benchmark is still needed.

